PR #9, opened 2016-01-28
Expose Tuple constructors

bripkens:
Stream combinators, mappers and other constructs used to manipulate
streams can be extracted to standalone functions. It is desirable to
test these functions in isolation, i.e. without initializing a Reactor
environment.

Unfortunately the Tuple constructors are package private, making it hard
to create instances for testing purposes. This change makes them public
for the sole purposes of testability in projects that use the reactor
project.

### DIFF
--- a/src/main/java/reactor/fn/tuple/Tuple1.java
+++ b/src/main/java/reactor/fn/tuple/Tuple1.java
@@ -34,7 +34,7 @@ public class Tuple1<T1> extends Tuple {
 
 	public final T1 t1;
 
-	Tuple1(int size, T1 t1) {
+	public Tuple1(int size, T1 t1) {
 		super(size);
 		this.t1 = t1;
 	}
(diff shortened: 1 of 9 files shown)


ifesdjeen:
You ain't gonna believe it :D

ifesdjeen:
Related: https://github.com/reactor/reactor/pull/531

bripkens:
Oh, damn. I didn't see that one. Thanks! :D

ifesdjeen:
funny thing is that i struggled with same problem. so i guess we should document it better...

bripkens:
I can only imagine documenting it at the Tuple constructors. At least, this is where I looked. I never would have expected to find a factory function on `Tuple` for `Tuple1`, since `Tuple1 extends Tuple`.

ifesdjeen:
True

PR #31, opened 2016-02-29
Use ms for long parameter and Duration instead of long + TimeUnit

sdeleuze:


akarnokd:
Note that `Duration.ofXXX` allocates.

smaldini:
@sdeleuze I would keep the Timer TimeUnit scheduling methods in effect. And therefore some path off Duration altogether. You might wonder why it matters to allocate on a timed signal ? Actually it is for all those pipelines that use a timeout / delay security without using them at all, so they would have to pay an extra object cost to do nothing without a static path.

sdeleuze:
Ok, I will update the PR to keep the `TimeUnit` variants in `Timer`.

sdeleuze:
I have updated this PR in order to:
 - Keep using `long` internally in order to avoid allocation
 - Use milliseconds consistently everywhere where a duration is expressed with a long
 - Add `Duration` variants for high level types

Issue #21, opened 2016-02-15
Add RxJava1CompletableConverter


sdeleuze:
Since [RxJava 1.1.1](https://github.com/ReactiveX/RxJava/releases/tag/v1.1.1) has been released, it would be nice to add a `Completable` to `Mono<Void>` converter to Reactor Core.

PR #33, opened 2016-03-06
Implements a completable converter to  Mono with no values

joaoevangelista:
Related to #21

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -47,7 +47,7 @@ ext {
   // Libraries
   jacksonDatabindVersion = '2.5.1'
   jsonPathVersion = '2.0.0'
-  rxJavaVersion = '1.0.14'
+  rxJavaVersion = '1.1.1'
 
   // Testing
   spockVersion = '1.0-groovy-2.4'
--- a/gradle/doc.gradle
+++ b/gradle/doc.gradle
@@ -22,7 +22,7 @@ configure(rootProject) {
 		group = "Reactor Core Javadoc"
 
 		description = "Generates aggregated Javadoc API documentation."
-		title = "Reactor Core ${version} API"
+		title = "Reactor Core M2"
 
 		if (JavaVersion.current().isJava8Compatible()) {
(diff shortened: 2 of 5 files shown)


joaoevangelista:
Did I got the right idea around the issue guys?

smaldini:
Wow excellent work ! Let me do a quick pass with it as we were freezing for releasing M2 :) Thanks a bunch

smaldini:
You might want to keep checking the presence of the class here https://github.com/joaoevangelista/reactor-core/blob/add-completable-mono/src/main/java/reactor/core/converter/DependencyUtils.java#L53  as for Observable (actually there's a slight bug on Single that should be loaded using the getClassloader instead).

Only Rx 1.1.1 has Completable and we could defeat the loading if one of the old rxjava is used instead as there is a widespread range between 1.0.7 and 1.1.1 that have first got Single then Completable..

joaoevangelista:
I'm not so into those `0b10000`, I think is some kind of boolean bitwise operartor, so  I follower what we already has to try to load the `Completable` class

```java
            Flux.class.getClassLoader().loadClass("rx.Observable");
            detected = RXJAVA_1_OBSERVABLE;
            Class.forName("rx.Single");
            detected |= RXJAVA_1_SINGLE;
            Class.forName("rx.Completable");
            detected |= RXJAVA_1_COMPLETABLE;

```

Is like that we use it ? As I thought we try to load the `rx.Observable` if it does exists detected got populated, if Single exists then the version of `rx` is up enough to have this class and so on.

smaldini:
Nice, it's ready to merge ( I just added flux.as(Observable.class), mono.as(Completable.class) etc to celebrate your PR :) 
Can you just make sure you got the contributor agreement filled up here please ? -> https://support.springsource.com/spring_committer_signup

joaoevangelista:
Nicee! 😆
I already signed up the CLA, for Spring Boot once!

PR #35, opened 2016-03-07
Make Exceptions more consistent

sdeleuze:
After more thoughts, keeping the name `wrap()` and `wrapUpstream()` is more consistent with the rest of the API and more easy to understand to the user given the new behavior (this was also names used previously). Please notice that I am not 100% sure we should keep `throwIfFatal()` call as it is but that can be polished later.

The second and third commits are a proposal to reduce Exceptions API which can be confusing (I think `Exceptions.onNextDropped()` should be removed as well).

Globally I think we should avoid to mix public API for end users and internal stuff coming from RSC.

### DIFF
--- a/src/main/java/reactor/core/publisher/EmitterProcessor.java
+++ b/src/main/java/reactor/core/publisher/EmitterProcessor.java
@@ -386,7 +386,7 @@ public final class EmitterProcessor<T> extends FluxProcessor<T, T>
 	public void onError(Throwable t) {
 		super.onError(t);
 		if (autoCancel && done) {
-			Exceptions.onErrorDropped(t);
+			throw Exceptions.wrapUpstream(t);
 		}
 		reportError(t);
 		done = true;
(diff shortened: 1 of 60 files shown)


akarnokd:
I don't recommend swapping `onErrorDropped` with `propagate` because that crashes the sequence even further. The purpose of `onErrorDropped` is to give the opportunity to catch and deliver exceptions to a global error handler if said exception is no longer allowed to travel as onError (that would violate the onNext* (onError|onComplete)? protocol).

smaldini:
@akarnokd is right this is a specific behavior pending further review for error circuit building, a big missing part in here. I think thi Pr should be re-scheduled post M2 with more thinking into what to do with the error. RSC prefers to log where we throw because our Generator/Processor input are catching error and doing something with it. That's up to what we do know and not encouraged as in RS, but maybe that's what we have to do, that's one of the few items to review for RC1.

PR #43, opened 2016-03-11
adds FlatMapIterable aliases

### LINKED ISSUE #38: Add flatMapIterable alias
It seems to be a good regular candidate to be factored in Flux/Mono.

joaoevangelista:
This close #38 

@smaldini I did not wrote the documentation since I'm afraid I could do some miss interpretation, also the Mono implementation of flatMapIterables didn't felt right since my understanding of Iterable as the stream of objects coming into the Flux, and Mono only has 0 or 1 element.

### DIFF
--- a/src/main/java/reactor/core/publisher/Flux.java
+++ b/src/main/java/reactor/core/publisher/Flux.java
@@ -16,6 +16,11 @@
 
 package reactor.core.publisher;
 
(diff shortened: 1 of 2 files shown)


smaldini:
We have dedicated operator for it 👍 thanks to @akarnokd

Issue #46, opened 2016-03-23
repeatWhenEmpty() versions show different behaviour


twoseat:
We tried switching cf-java-client from using `.repeatWhenEmpty()` with a `maxRepeat` value, to using the version that just takes a `repeatFactory`. This leads to a number of test failures caused by 'insufficient' mocking in the tests - a test that successfully provided two responses to a repeated request [(example)](https://github.com/cloudfoundry/cf-java-client/blob/master/cloudfoundry-operations/src/test/java/org/cloudfoundry/operations/applications/DefaultApplicationsTest.java#L584-L594) before our change fails with a null pointer exception because it's being asked to supply a third response with the new approach.

For now we're using the two parameter version with `MAXINT-1` as a workaround [(example)](https://github.com/cloudfoundry/cf-java-client/blob/master/cloudfoundry-util/src/main/java/org/cloudfoundry/util/JobUtils.java#L47). Should the two versions of `.repeatWhenEmpty()` differ in this way?

akarnokd:
The operator is implemented differently for `MAX_VALUE` and non `MAX_VALUE`. I'm not sure what it supposed to do btw. I'd assume this repeats the source if it just completes without an `onNext`, but it is implemented via `repeatWhen()` and some other logic but I don't see the source being shared thus the operator could subscribe to the source twice.

akarnokd:
See #47 for a possible fix; not sure how it would affect your mocks (as I don't do mocks myself).

smaldini:
Pending verification after #47 merge

twoseat:
Tested and working in latest snapshot

PR #47, opened 2016-03-23
Reimplement Mono.repeatWhenEmpty()

akarnokd:
In response to #46.

The operator works by deferring a stateful sequence that sends a non-emptyness indicator to the `repeatWhen` where a `takeWhile` is evaluated with it and closes the repeat if the source did emit something.

I couldn't find tests for it so I'm not sure what the expected behavior was before (i.e, if repeat of 2 means 2 subscriptions total or 3 thus repeat 0 doesn't even subscribe?).


smaldini:
good assumptions, this was led by https://github.com/cloudfoundry/cf-java-client requirements let's go ahead with this as they reported internally a similar behavior.

twoseat:
Tested with latest snapshot and this fixes my issue, thanks!

PR #53, opened 2016-04-01
repeatWhenEmpty Iterations

nebhale:
Previously, a fix when into the `repeatWhenEmpty` implementations to deal with finite versus infinite repeat loops.  This fix broke two pieces of functionality.  First, it returned the wrong error when the max retries was exceeded.  Second it returned 0 for each iteration of the loop, rather than incrementing as designed.  This broke any functionality dependent on the iteration counter (like a backoff algorithm).

This change refactors the implementation to fix both of those problems and updates the tests to more accurately test that both bits of functionality are satisfied.

### DIFF
--- a/src/main/java/reactor/core/publisher/Mono.java
+++ b/src/main/java/reactor/core/publisher/Mono.java
@@ -22,7 +22,7 @@ import java.util.concurrent.Callable;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.Future;
 import java.util.concurrent.atomic.AtomicBoolean;
-import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicLong;
 import java.util.function.BiConsumer;
 import java.util.function.Consumer;
 import java.util.function.Function;
@@ -30,6 +30,7 @@ import java.util.function.LongConsumer;
 import java.util.function.Predicate;
 import java.util.function.Supplier;
 import java.util.logging.Level;
+import java.util.stream.LongStream;
 
 import org.reactivestreams.Publisher;
 import org.reactivestreams.Subscriber;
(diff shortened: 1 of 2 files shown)


src/main/java/reactor/core/publisher/Mono.java
```diff
@@ -1537,20 +1538,32 @@ public String getName() {
...
-			return Mono.defer(() -> {
-				AtomicBoolean nonEmpty = new AtomicBoolean();
-				AtomicInteger count = new AtomicInteger();
-				return Flux.from(this.doOnSuccess(e -> nonEmpty.lazySet(e != null)))
-						.repeatWhen(o -> repeatFactory.apply(o.takeWhile(e -> !nonEmpty.get() && count.getAndIncrement() < maxRepeat)))
-						.single();
-			});
-		}
 		return Mono.defer(() -> {
+			Flux<Long> iterations;
```
akarnokd: I think there is no need for the complicated sequence.

```java
	    if (maxRepeat != Integer.MAX_VALUE) {
	        return Mono.defer(() -> {
	            AtomicBoolean nonEmpty = new AtomicBoolean();
	            AtomicLong count = new AtomicLong();
	            return Flux.from(this.doOnSuccess(e -> nonEmpty.lazySet(e != null)))
	                    .repeatWhen(o -> repeatFactory.apply(
	                            o.map(v -> count.get())
	                            .takeWhile(e -> {
	                                long c = count.getAndIncrement();
	                                if (c < maxRepeat) {
	                                    return !nonEmpty.get();
	                                }
	                                throw new IllegalStateException("Exceeded maximum number of repeats");
	                            })
                            )
                        )
	                    .single();
	        });
	    }
	    return Mono.defer(() -> {
	        AtomicBoolean nonEmpty = new AtomicBoolean();
	        AtomicLong count = new AtomicLong();
	        return Flux.from(this.doOnSuccess(e -> nonEmpty.lazySet(e != null)))
	                .repeatWhen(o -> repeatFactory.apply(
	                        o.takeWhile(e -> !nonEmpty.get())
	                        .map(v -> count.getAndIncrement())
                        )
                    )
	                .single();
	    });
```

nebhale:
The problem that started this whole chain was that the finite and infinite versions had two disjoint code-paths and they operated differently.  An update to one path didn't automatically apply to both paths.  By creating a conditional exclusively for the finite/infinite difference and ensuring that all other code is common we avoid that problem now and in the future.

akarnokd:
Then have just one path and widen the repeat count:

```java
        long max = maxRepeat != Integer.MAX ? maxRepeat : Long.MAX_VALUE;
        return Mono.defer(() -> {
            AtomicBoolean nonEmpty = new AtomicBoolean();
            AtomicLong count = new AtomicLong();
            return Flux.from(this.doOnSuccess(e -> nonEmpty.lazySet(e != null)))
                    .repeatWhen(o -> repeatFactory.apply(
                            o.map(v -> count.get())
                            .takeWhile(e -> {
                                long c = count.getAndIncrement();
                                if (c < max) {
                                    return !nonEmpty.get();
                                }
                                throw new IllegalStateException("Exceeded maximum number of repeats");
                            })
                        )
                    )
                    .single();
        });
```

Zteve:
For what it's worth, I think separating the generation of the iteration from the repeat logic is rather nice, and the error when the repeats are exceeded is clearly the responsibility of the iteration generator. This separates the "on this condition do something" logic from the "generate this number of triggers" logic. It also makes the final solution easier to read (and easier to test by decomposition into parts).

smaldini:
We can improve separately but let's keep the signature for now.

PR #62, opened 2016-04-11
Avoid stack overflow from repeatedly applying xxxWith()

akarnokd:
This PR uses macro-fusion: rewrite to avoid StackOverflowError when one applies `mergeWith`, `concatWith`, `startWith` and `ambWith` repeatedly.

The `mergeWith` is ugly because without a dedicated fixed-length `merge` operator, one has to verify the proper input to the fromArray.


akarnokd:
Will post these inlined after sync with Rsc.

PR #66, opened 2016-04-17
Implements onError* methods with exception type as parameter

### LINKED ISSUE #58: Improve handling of specific Throwable types
Right now it is a bit cumbersome to handle an error of a specific type.

For example, if I need `Mono.otherwise` to handle only a specific error I just do:

```java
.otherwise( t -> {
	if(t instanceof SpecificException) {
		return ...
	}
	return Mono.error(t);
}
```

It would be nice if we had something similar to `Flux.doOnError(Class<E> exceptionType, Consumer<E> onError)`.

There are likely other places that sort of handling would be nice.

joaoevangelista:
Now a Mono#doOnError method is added accepting the type of exception it should handle, more than one can be chained togheter
so we can create diffent types of handlers by the type of exception, the new method use the current method that only accepts
a consumer.

Another otherwise method was created to use a supplier when the exception types match,
falling back to throwing the current error into a Mono#error

Flux can now resume with when a given exception type matches the error type, the error type can not be changed like Mono#otherwiseReplace
because of UpstreamException wraps fatals exceptions based on Reactive Streams list of exceptions allowed to bubble (as seen on documentation)

Fixes #58

### DIFF
--- a/src/main/java/reactor/core/publisher/Flux.java
+++ b/src/main/java/reactor/core/publisher/Flux.java
@@ -16,36 +16,23 @@
 
 package reactor.core.publisher;
 
-import org.reactivestreams.Processor;
-import org.reactivestreams.Publisher;
-import org.reactivestreams.Subscriber;
-import org.reactivestreams.Subscription;
-import reactor.core.flow.Cancellation;
-import reactor.core.flow.Fuseable;
-import reactor.core.queue.QueueSupplier;
-import reactor.core.scheduler.Scheduler;
-import reactor.core.scheduler.TimedScheduler;
-import reactor.core.scheduler.Timer;
-import reactor.core.state.Backpressurable;
-import reactor.core.state.Introspectable;
-import reactor.core.subscriber.LambdaSubscriber;
-import reactor.core.subscriber.SignalEmitter;
-import reactor.core.subscriber.SubscriberWithContext;
-import reactor.core.subscriber.Subscribers;
-import reactor.core.tuple.*;
-import reactor.core.util.Exceptions;
-import reactor.core.util.Logger;
-import reactor.core.util.PlatformDependent;
-import reactor.core.util.ReactiveStateUtils;
-
 import java.time.Duration;
 import java.util.*;
-import java.util.concurrent.Callable;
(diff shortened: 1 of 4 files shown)


src/main/java/reactor/core/publisher/Flux.java
```diff
@@ -968,6 +969,31 @@
...
+	 *
+	 * @param errorType exception type to match
+	 * @param source source sequence
+	 * @param fallback supplier called when the exceptions and errorType matches
+	 * @param <T> {@link Subscriber} type target
+	 * @return a {@link Flux} handling error if exceptions match otherwise an rejected Flux
+	 */
+	public final Flux<T> onErrorResumeWith(
+			Class<? extends Throwable> errorType,
+			Publisher<? extends T> source,
```
akarnokd: What is the purpose of `source` on an instance operator?

smaldini: source is a common typo that should be called other.

smaldini:
@joaoevangelista you are a hero ! We need to merge your other PR for Cont.MD but I need to do a mass apply style commit before M3. Let's see before if the question I raised in the issue is valid.

src/main/java/reactor/core/publisher/Flux.java
akarnokd: The method doesn't use `this` so either it should be static or the `source` parameter should be removed and have `new FluxResume<>(this, ...)`.

smaldini: Ha didn't see the rest of the change, indeed it should have this and remove source parameter !

joaoevangelista: Indeed, my bad, it supposed to be a static method, I changed it to be static since a instance method is provided too

joaoevangelista:
Sure! I think is better to do apply the style after the PR are merged, so we can avoid headache from git merge

smaldini:
Damn mistake sorry.

joaoevangelista:
Reopen it?

smaldini:
Can't reopen on this one apparenly new PR possible please ? I am so sorry -_-

PR #68, opened 2016-04-17
Implements onError* methods with exception type as parameter

joaoevangelista:
See #66

### DIFF
--- a/src/main/java/reactor/core/publisher/Flux.java
+++ b/src/main/java/reactor/core/publisher/Flux.java
@@ -3336,18 +3336,6 @@ public abstract class Flux<T> implements Publisher<T>, Introspectable, Backpress
 		return new FluxResume<>(this, fallback);
 	}
 
-	/**
-	 * Subscribe to a returned fallback publisher when any error occurs.
-	 * <p>
-	 * <img class="marble" src="https://raw.githubusercontent.com/reactor/projectreactor.io/master/src/main/static/assets/img/marble/onerrorresumewith.png" alt="">
-	 * <p>
-	 * @param fallback the {@link Function} mapping the error to a new {@link Publisher} sequence
-	 *
-	 * @return a new {@link Flux}
-	 */
-	public final Flux<T> onErrorResumeWith(Function<Throwable, ? extends Publisher<? extends T>> fallback) {
-		return new FluxResume<>(this, fallback);
-	}
 
 	/**
 	 * Subscribe to a fallback Publisher if the error exception matches the given type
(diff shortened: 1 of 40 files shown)


src/main/java/reactor/core/publisher/Flux.java
```diff
@@ -3311,6 +3337,29 @@ public void subscribe(Subscriber<? super T> s) {
...
+	 *
+	 * @param errorType exception type to match
+	 * @param fallback supplier called when the exceptions and errorType matches
+	 * @return a {@link Flux} handling error if exceptions match otherwise an rejected Flux
+	 */
+	public final Flux<T> onErrorResumeWith(
+			Class<? extends Throwable> errorType,
+			Supplier<Publisher<? extends T>> fallback) {
+
+		return new FluxResume<>(this, throwable -> {
```
akarnokd: You could just call `onErrorResumeWith(errorType, this, fallback);`

smaldini:
I don't think otherwiseReplace is a generic bet we can make as these operators are equally useful at providing a fallback. However something like otherwise(Class, Mono) in addition to Supplier would be handy.

joaoevangelista:
Yea, otherwiseReplace is not that generic, I made it as an alias for `mono.otherwise(ex, e -> Mono.error(new YourException(e));` I just don't know if the name is alright, and if the explicit version is better, considering the `otherwise(Class,Mono)` version, there is no problem for me to remove it, I made it for a validation

joaoevangelista:
What about replace `otherwiseReplace` with something like : 

```java
	public final Mono<T> otherwise(Class<? extends Throwable> ex,
			Function<Throwable, Mono<T>> fallback) {
		return otherwise(throwable -> {
			if (ex.isAssignableFrom(throwable.getClass())) {
				return otherwise(fallback);
			}
			else {
				return Mono.error(throwable);
			}
		});
	}
```

That way we (now) preserve the original exception and filter the exception type.

This will leave us with `otherwise(Class,Mono)` and `otherwise(Class,Function)`, the `otherwise(Class,Supplier)` will be removed, first because is kinda ambiguous to have Mono and Supplier Mono, and it will generate ClassCastException because of the overloaded method, at least for Groovy

smaldini:
The Function will be ambiguous with Mono (both takes 1 arg), but wdyt something like `otherwise(Class, Mono)`  and `mapError(Class, Function)` ?

joaoevangelista:
More meaningful indeed, you provide another Mono in case of error or map it to something else, I'll do it tonight!

smaldini:
Possible to get the stash treatment for this one ?

joaoevangelista:
Closed in favor of #70

PR #64, opened 2016-04-12
Contribution guide

joaoevangelista:
This is a contribution guide which is based on the Spring Framework's, it also include the code style configuration files for IDEA and Eclipse and a link to it in README.

Closes #42 

> I have signed and agree to the terms of the Spring Individual Contributor License Agreement.

PR #69, opened 2016-04-18
Contribution guide and Code Style

### LINKED ISSUE #42: Add code style configuration
A code style configuration file would be nice to have, so contributors don't mess up the already formatted code.

An eclipse format file could be enough since IntelliJ has a converter from that file.

joaoevangelista:
This is a contribution guide which is based on the Spring Framework's, it also include the code style configuration files for IDEA and Eclipse.

Replacement for #64 
Closes #42 

> I have signed and agree to the terms of the Spring Individual Contributor License Agreement.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,262 @@
+# Welcome to Reactor's Contributor guide!
+
+### Understand the basics
+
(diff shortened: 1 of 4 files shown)


smaldini:
Nicesome!

PR #70, opened 2016-04-19
Add exception mappers based on the exception type

### LINKED ISSUE #58: Improve handling of specific Throwable types

joaoevangelista:
On Mono it includes a new `otherwise`  method that executes the callback
 only when the exception type matches the error and supply a Mono fallback,
 a new `mapError `was added to allow map the error into a new Mono or a new rejected Mono preserving the original exception.

On Flux a `onErrorResumeWith` with exception type matchers was added both for contructing a new Flux and an operator, similar to `Mono#otherwise(Class, Mono)`

Fixes #58 

> I have signed and agree to the terms of the Spring Individual Contributor
License Agreement.

### DIFF
--- a/src/main/java/reactor/core/publisher/Flux.java
+++ b/src/main/java/reactor/core/publisher/Flux.java
@@ -51,6 +51,7 @@ import reactor.core.util.*;
  *
  * @author Sebastien Deleuze
  * @author Stephane Maldini
+ * @author Joao Pedro Evangelista
  *
  * @see Mono
  * @since 2.5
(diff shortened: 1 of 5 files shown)


src/main/java/reactor/core/publisher/Flux.java
```diff
@@ -970,6 +971,30 @@
...
+	 *
+	 * @param errorType exception type to match
+	 * @param source source sequence
+	 * @param fallback supplier called when the exceptions and errorType matches
+	 * @param <T> {@link Subscriber} type target
+	 * @return a {@link Flux} handling error if exceptions match otherwise an rejected Flux
+	 */
+	public static <T> Flux<T> onErrorResumeWith(
+			Class<? extends Throwable> errorType,
+			Publisher<? extends T> source,
```
akarnokd: I'd have the `source` as the first parameter.

joaoevangelista: It could be, but I thought on having the error type since it is what is different from the other method and it operates based on it

smaldini: Placing error in second arg opens Vararg support and multi-catch tho :)

joaoevangelista: That would be great, I think I could not finish until the release of M3 that is planned, go for it!

smaldini:
Already implemented now :) Thanks a lot for the inspiration and interest @joaoevangelista !!!

joaoevangelista:
Awesome, thanks for all support!

PR #78, opened 2016-04-29
DeferCompose operator proposition

akarnokd:
This PR adds two simple, in-sequence operators that let's one defer the composition itself which then allows per-subscriber state, similar to defer().


smaldini:
Could that superseed the current compose() ?

akarnokd:
The new `compose` behaves just like this. Did you update `Mono.compose`?

smaldini:
Yes , but would you prefer to have it as an operator or as(defer(this) is enough ?

akarnokd:
Yes.

smaldini:
I just re-open to steal the tests anyway :) will cleanup deferCompose right after

PR #99, opened 2016-06-07
Review, javadoc and code cleanup

akarnokd:
Fixing annotations, adding missing tags in javadoc, removing unnecessary casts and type arguments.


smaldini:
Awesome that's quite helpful!

PR #115, opened 2016-07-08
Rename methods with long durations with Millis suffix

sdeleuze:
The goal is to make the time unit more explicit + enable using
Reactor on Android Nougat.


JakeWharton:
:+1: truly sad we're not getting JSR310 in N 😢

PR #118, opened 2016-07-14
Remove @since from Javadoc

sdeleuze:
Since so much classes have changed since 2.0, I propose to remove the current `@since 2.5` Javadoc comments instead of updated them to `@since 3.0`. We will start dealing with `@since` from 3.0.x.


sdeleuze:
Merged as https://github.com/reactor/reactor-core/commit/d9126d07bf1b782b4f79607dfca3e281508dd0c3

PR #120, opened 2016-07-15
Tuple Hash Code

nebhale:
Previously, `Tuple` hash codes were broken because at the root of the hierarchy, `Object` identity was used to start the calculation.  This bug was due to the recent removal of the root `Tuple` type, leaving `Tuple2` calling `Object.hashCode()` instead of a properly structured `Tuple.hashCode()`.  This change fixes the bug by using the `t1` as the start of the hash code calculation.

### DIFF
--- a/src/main/java/reactor/util/function/Tuple2.java
+++ b/src/main/java/reactor/util/function/Tuple2.java
@@ -104,24 +104,19 @@ public class Tuple2<T1, T2> implements Iterable, Serializable {
 
 	@Override
 	public boolean equals(Object o) {
-		if (o == null) return false;
+		if (this == o) return true;
+		if (o == null || getClass() != o.getClass()) return false;
 
-		if (!(o instanceof Tuple2)) return false;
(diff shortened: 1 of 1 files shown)


src/main/java/reactor/util/function/Tuple2.java
```diff
@@ -120,8 +120,7 @@ public boolean equals(Object o) {
```
Zteve: You might want to replace `equals()` as well:
```
public boolean equals(Object o) {
        if (this == o) return true;
        
        if (o == null || getClass() != o.getClass()) return false;

        Tuple2<?, ?> tuple2 = (Tuple2<?, ?>) o;

        if (size() != tuple2.size()) return false;
        
        if (t1 != null ? !t1.equals(tuple2.t1) : tuple2.t1 != null) return false;
        
        return t2 != null ? t2.equals(tuple2.t2) : tuple2.t2 == null;

    }
```
as well.  (Original version doesn't test == to start with.)

nebhale: Good idea, let me investigate.

PR #125, opened 2016-07-28
Fixes #124. Introduce transform() operator and update javadoc for com…

### LINKED ISSUE #124: Add transform() operator to Flux/Mono to support transforming stream without deferring
Discussed in gitter:

**Daniel Feist @dfeist**
I see as() as an operator i'd primarily use when converting flux to something else, not for chaining, thats also typically how "as" is interpreted in API's. The fact that you can chain if you use a function that returns Flux is almost a convenient side-effect, but not the main purpose of that operator in my opinion.

Compose on the other hand is for chained composition, but it defers which isn't required/desired for simple Flux transformation.

**Stephane Maldini @smaldini**
no i see the point totally

dfeist:
…pose()

### DIFF
--- a/src/main/java/reactor/core/publisher/Flux.java
+++ b/src/main/java/reactor/core/publisher/Flux.java
@@ -5171,7 +5171,11 @@ public abstract class Flux<T> implements Publisher<T> {
 	}
 
 	/**
-	 * Transform this {@link Flux} in order to generate a target {@link Flux}.
+	 * Transform this {@link Flux} in order to generate a target {@link Flux}. Unlike {@link #compose(Function)}, the
+	 * provided function is executed as part of assembly.
+	 *
+	 * {@code Function<Flux, Flux> applySchedulers = flux -> flux.subscribeOn(Schedulers.io()).publishOn(Schedulers.computation());
+	 *        flux.transform(applySchedulers).map(v -> v * v).subscribe(Subscribers.unbounded())}
 	 *
 	 * @param transformer the {@link Function} to immediately map this {@link Flux} into a target {@link Flux}
 	 * instance.
--- a/src/main/java/reactor/core/publisher/Flux.java
+++ b/src/main/java/reactor/core/publisher/Flux.java
@@ -2053,21 +2053,6 @@ public abstract class Flux<T> implements Publisher<T> {
 	}
 
 	/**
-	 * Transforms this {@link Flux} in order to generate a target {@link Flux}.
-	 *
-	 * @param transformer the {@link Function} to immediately map this {@link Flux} into a target {@link Flux}
-	 * instance.
-	 * @param <V> the item type in the returned {@link Flux}
-	 *
-	 * @return a new {@link Flux}
-	 * @see #compose(Function) for deferred composition of {@link Flux} for each {@link Subscriber}
-	 * @see #as for a loose conversion to an arbitrary type
-	 */
-	public final <V> Flux<V> transform(Function<? super Flux<T>, ? extends Publisher<V>> transformer) {
-		return from(transformer.apply(this));
-	}
-
-	/**
 	 * Bind dynamic sequences given this input sequence like {@link #flatMap(Function)}, but preserve
 	 * ordering and concatenate emissions instead of merging (no interleave).
 	 * Errors will immediately short circuit current concat backlog.
@@ -2654,7 +2639,7 @@ public abstract class Flux<T> implements Publisher<T> {
 	 * @return a transforming {@link Flux} that emits tuples of time elapsed in milliseconds and matching data
 	 */
 	public final Flux<Tuple2<Long, T>> elapsed() {
-		return compose(f -> f.map(new Elapsed<>()));
+		return transform(f -> f.map(new Elapsed<>()));
 	}
 
 	/**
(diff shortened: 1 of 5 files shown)


smaldini:
Few remarks:
- We should stick to alphabetical ordering for declaration even close of compose this is
- Mono#transform should use Mono#from
- We should use that PR to lift the Mono restriction on compose given we use "from" for the returned Publisher. Eventually adapt the doc to precise that a single signal will be observed from the transformed sequence.

src/main/java/reactor/core/publisher/Flux.java
```diff
@@ -5173,6 +5171,21 @@ public String toString() {
 	}
 
 	/**
+	 * Transforms this {@link Flux} in order to generate a target {@link Flux}.
```
artembilan: Method JavaDocs must be in the imperative.
See the previous for `compose()`: `Defer the transformation...`. 
https://github.com/spring-projects/spring-framework/wiki/Spring-Framework-Code-Style

Although that is for Spring. Reactor is a different story :smile:

src/main/java/reactor/core/publisher/Flux.java
```diff
@@ -5173,6 +5171,21 @@ public String toString() {
 	}
 
 	/**
+	 * Transform this {@link Flux} in order to generate a target {@link Flux}.
```
akarnokd: "Unlike compose(), the provided function is executed during assembly time. Example use:

Function<Flux<Integer>, Flux<Integer>> applySchedulers = flux -> flux.subscribeOn(Schedulers.io()).publishOn(Schedulers.computation());

source.transform(applyScheduler).map(v -> v * v)...
"

src/test/java/reactor/core/publisher/FluxPublishTest.java
```diff
@@ -112,7 +112,7 @@ public void cancelComposes2() {
 	public void pairWise() {
 		TestSubscriber<Tuple2<Integer, Integer>> ts = TestSubscriber.create();
 
-		range(1, 9).compose(o -> zip(o, o.skip(1)))
+		range(1, 9).transform(o -> zip(o, o.skip(1)))
```
akarnokd: I'd keep the original compose version as well and have a separate test for this. In addition, I'd test if the function is executed the time the transform is applied an not when the subscription happens.

src/test/java/reactor/core/publisher/scenarios/FluxTests.java
```diff
@@ -852,7 +852,7 @@ public void testParallelAsyncStream2() throws InterruptedException {
 			final String source = "ASYNC_TEST " + i;
 
 			Flux.just(source)
-			    .as(operationStream -> operationStream.publishOn(asyncGroup)
+			    .transform(operationStream -> operationStream.publishOn(asyncGroup)
```
akarnokd: Again, keep the original test and do a version with transform.

src/test/java/reactor/core/publisher/tck/FluxAndProcessorTests.java
```diff
@@ -67,7 +67,7 @@
 		                                                                   .doOnNext(this::monitorThreadUse))
 		                                      .doOnNext(array -> cumulatedJoin.getAndIncrement())
 		                                      .subscribeWith(TopicProcessor.create("fluxion-raw-join", bufferSize))
-		                                      .as(Flux::from)
+		                                      .transform(Flux::from)
```
akarnokd: Duplicate test case instead.

dfeist: Unless i'm mistaken this line is redundant anyway and can be removed.

src/main/java/reactor/core/publisher/Mono.java
```diff
@@ -2517,6 +2515,25 @@ public String toString() {
...
+	 *
+	 * @param transformer the {@link Function} to immediately map this {@link Mono} into a target {@link Mono}
+	 * instance.
+	 * @param <V> the item type in the returned {@link Mono}
+	 *
+	 * @return a new {@link Mono}
+	 * @see #compose(Function) for deferred composition of {@link Mono} for each {@link Subscriber}
+	 * @see #as for a loose conversion to an arbitrary type
+	 */
+	public final <V> Mono<V> transform(Function<? super Mono<T>, ? extends Mono<V>> transformer) {
```
smaldini: Should be Publisher return ?

dfeist: No, why?  This isn't loose `as()` but rather designed for Mono chaining in the same way composed is used but without deferring the transformation.  What i didn't update here (but did in compose) is the restriction in function type. Should be:
`public final <V> Mono<V> transform(Function<? super Mono<T>, ? extends Publisher<V>> transformer)`

src/test/java/reactor/core/publisher/tck/FluxAndProcessorTests.java
akarnokd: It tests the `as` operator, they should have some test coverage so an unwanted change can be captured.

src/main/java/reactor/core/publisher/Mono.java
akarnokd: The problem is that Publisher can be of any length and now you have to apply an extra operator to restrict it to 0 or 1. Leave it Mono->Mono.

smaldini:
Nicesome

src/main/java/reactor/core/publisher/Flux.java
```diff
@@ -2641,7 +2639,7 @@
 	 * @return a transforming {@link Flux} that emits tuples of time elapsed in milliseconds and matching data
 	 */
 	public final Flux<Tuple2<Long, T>> elapsed() {
-		return compose(f -> f.map(new Elapsed<>()));
+		return transform(f -> f.map(new Elapsed<>()));
```
akarnokd: `Elapsed` is [stateful](https://github.com/reactor/reactor-core/blob/master/src/main/java/reactor/core/publisher/Elapsed.java#L29) so it needs to be deferred for each subscriber.

dfeist: Good catch.  Just opened another PR to revert this.

smaldini: I fixed it after the merge

PR #127, opened 2016-07-29
Fixes #124. Introduce transform() operator. Revert change to elapsed().

### LINKED ISSUE #124: Add transform() operator to Flux/Mono to support transforming stream without deferring

dfeist:


dfeist:
Looks like @smaldini already fixed this, so closing PR.

PR #128, opened 2016-07-30
Fixes #80. Provide Exception methods to support use of functions that…

### LINKED ISSUE #80: Consider introducing utility method for checked functions
When adopting reactor, but needing to wrap and use existing code you find yourself using try/catch blocks within `map` and other operators repeatedly which severely impacts the readability of code.  The introduction of a CheckedFunction could help with this significantly.

`Exceptions.checked(CheckedFunction) : Function`

dfeist:
… throw checked exceptions and implicitly propagate exceptions.

(Opening PR to share and get feedback on approach, will fill out javadoc and update method names as required based on feedback)

### DIFF
--- a/src/main/java/reactor/core/Exceptions.java
+++ b/src/main/java/reactor/core/Exceptions.java
@@ -17,6 +17,10 @@
 package reactor.core;
 
 import java.util.concurrent.atomic.AtomicReferenceFieldUpdater;
+import java.util.function.Consumer;
+import java.util.function.Function;
+import java.util.function.Predicate;
+import java.util.function.Supplier;
 
 import org.reactivestreams.Subscription;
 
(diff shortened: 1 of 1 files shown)


smaldini:
Round of approval / feedback : @akarnokd , @rstoyanchev , @sdeleuze

akarnokd:
How about having CheckedX on the operators directly? Avoids wrapping and unwrapping checked exceptions.

smaldini:
Wait, changing all the signatures ? naaaaaaaah

smaldini:
Plus not to forget that usually wrapping Exception stuff means you are mapping a "blocking" call which overtime shouldn't be that necessary or you would use flatMap more for that.

dfeist:
Agree with Stephane. I argued for this being in core Exceptions and not a
seperate module,  but putting it in Flux is going too far IMO.

This will be used quite commonly but still isn't best practice and usage
should reduce gradually over time I'd hope.

smaldini:
Closing for now, don't hesitate to port that as a standalone new addon :D

PR #129, opened 2016-07-31
Update README for 3.0.0.RC1

praseodym:


smaldini:
Good eye thanks a lot @praseodym !

PR #132, opened 2016-08-01
Fixes #86. Add mapOrFilter operator.

### LINKED ISSUE #86: Flux.map() and null-handling when wrapping existing blocking API's
The only way to do this currently is via the use of `flatMap` and `Mono.justOrEmpty()` which isn't as efficient and causes more allocations/stack.

One option is to introduce something in Flux e.g. `Flux.mapOrEmpty()` that behaves the same as `map` but drops message and does `request(1)` when null is returned, but I understand this clutters the primary API where nulls simply are not allowed.  So maybe an alternative is to implement something outside of Flux in something like `LegacyUtils` that can then somehow be pulled into and used in a Flux stream just in the cases where you need to deal with legacy API's that throw exceptions/return null.

This is somewhat related to: https://github.com/reactor/reactor-core/issues/80

dfeist:
@smaldini What do you think regarding this approach using boolean flag vs creating specific FluxSource implementations for mapOrFilter? Let me know and I can update/rework as required..

### DIFF
--- a/src/main/java/reactor/core/publisher/Flux.java
+++ b/src/main/java/reactor/core/publisher/Flux.java
@@ -138,10 +138,10 @@ public abstract class Flux<T> implements Publisher<T> {
             Publisher<? extends T> source = sources[0];
             if (source instanceof Fuseable) {
 	            return onAssembly(new FluxMapFuseable<>(source,
-			            v -> combinator.apply(new Object[]{v})));
+			            v -> combinator.apply(new Object[]{v}), false));
             }
 			return onAssembly(new FluxMap<>(source,
-					v -> combinator.apply(new Object[]{v})));
+					v -> combinator.apply(new Object[]{v}), false));
 		}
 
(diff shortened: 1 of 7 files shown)


smaldini:
Mhhh I'd go more for an individual operator.

dfeist:
Ok,  i'll update.  That was first approach I took but wasn't sure it made
sense to duplicate all those classes..

dfeist:
@smaldini Hang on,  I have seperate 'mapOrFilter operator already in PR. I was referring to if it made sense to duplicate FluxMap,  MonoMap etc.  Or if use of boolean is here is ok.  Take a look at PR and it'll be clear.  Please confirm.

akarnokd:
I don't like this addition to `map` either. You should go with a separate operator. In addition, you neglected the fusion parts and didn't provide unit tests that verify the behavior.

dfeist:
@akarnokd 👍 Will update.  Wanted to confirm approach before adding test..

PR #134, opened 2016-08-03
Test case demonstrating options for use of a custom operator

dfeist:


src/main/java/reactor/core/publisher/Flux.java
```diff
@@ -87,6 +87,8 @@
  */
 public abstract class Flux<T> implements Publisher<T> {
 
+	public interface Operator<I, O> extends Function<Subscriber<? super I>, Subscriber<? super O>> {}
```
akarnokd: I'd rather name the arguments explicitly:

```java
Operator<Upstream, Downstream> extends Function<Subscriber<? super Downstream>, Subscriber<? super Upstream>>
```

The function takes the downstream's Subscriber and transforms it into something that will be subscribed to the upstream.

smaldini:
If you could rework that PR btw that would be awesome (without the extra API but the tests are nice and you keep credit for the good stuff :))

PR #137, opened 2016-08-07
Fixes #136 Fix inconsistencies in some method name between Flux/Mono/…

### LINKED ISSUE #136: Fix inconsistencies in some method name between Flux/Mono/FluxSink/MonoSink
1. FluxSink/SynchronousSink/BlockingSink use `fail()` but Flux uses `doOnError`, Subscriber `onError` etc. Sinks should use `error()` IMO.
2. MonoSink uses `complete()/complete(T t)` (like Flux, FluxSink) but Mono uses `doOnSuccess()`.  IMO MonoSink shouldn't use `complete()` given Mono doesn't use concept of completion, but rather `success()`.
3. MonoSink uses `fail()`, same as other sinks.  This might be OK if it was consistent with Mono, but it's not as Mono uses `doOnError()` rather than the concept of failure.

Ideally sinks (at least least the flux ones) should share an interface.

dfeist:
…FluxSink/MonoSink

### DIFF
--- a/src/main/java/reactor/core/publisher/BlockingSink.java
+++ b/src/main/java/reactor/core/publisher/BlockingSink.java
@@ -209,14 +209,14 @@ public final class BlockingSink<E>
 
 	/**
 	 *
-	 * Try calling {@link Subscriber#onError(Throwable)} on the delegate {@link Subscriber}. {@link BlockingSink#fail(Throwable)}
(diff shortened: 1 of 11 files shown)


dfeist:
Closes as seems @smaldini couldn't sleep and decided to do this himself :-)

PR #206, opened 2016-10-11
Fixes #204 Self Suppression IllegalArgumentException when 'onErrorRes…

### LINKED ISSUE #204: Self Suppression IllegalArgumentException when 'onErrorResume' propagates existing exception
Sometime conditional logic is required to determine if an error signal should be handled or propagated.  The easiest and most flexible way to do this is via the use of something like this:

```
flux.onErrorResume(e -> {
   if(something){
      return just(someValue);
   else {
      throw propagate(e);
   });
```

In this above example the else clause propagates the existing exception.  This fails though due to FluxResume attempting to do `e.addSuppressed(e)` [here](https://github.com/reactor/reactor-core/blob/master/src/main/java/reactor/core/publisher/FluxResume.java#L92)

dfeist:
…ume' propagates existing exception


smaldini:
🏆

PR #208, opened 2016-10-11
Fixes #205 reactor.core.publisher.FluxPeekFuseable.PeekFuseableSubscr…

### LINKED ISSUE #205: doOnNext does not use Operators.onOperatorError
Exceptions are rather caught and handled in the previous operator, but depending on the fuseable mode when they are handled the data signal is null.

dfeist:
…iber#poll does not use Operators.onOperatorError


dfeist:
travis

PR #214, opened 2016-10-14
Fixes #213 Reduce logging level used in Hooks.onXX

### LINKED ISSUE #213: Reduce logging level used in Hooks.onXX
The use of WARN for things like `reactor.core.publisher.Hooks#onOperatorError` which is some situations (like mind) needs to be used consistently is too high IMO.

dfeist:


smaldini:
Thanks sir @dfeist as usual good pragmatic call :)

PR #243, opened 2016-10-28
Fix for the onNext/onComplete callback reversal in async-fused sequences

akarnokd:
Mono's `doOnSuccess` logic expects the onNext/onComplete calls from upstream to happen in order but in fused sources, `onComplete` callback may happen first because the `onNext` callback is done in `poll`. This changes the `peek` operator to not call `onComplete` upfront with async sources (sync sources never call `onComplete` anyway) but call the `onComplete` callback when the source is drained and done itself in poll.

### DIFF
--- a/src/main/java/reactor/core/publisher/FluxPeekFuseable.java
+++ b/src/main/java/reactor/core/publisher/FluxPeekFuseable.java
@@ -92,7 +92,7 @@ final class FluxPeekFuseable<T> extends FluxSource<T, T> implements Fuseable,
 
 		int sourceMode;
 
-		boolean done;
+		volatile boolean done;
 
 		public PeekFuseableSubscriber(Subscriber<? super T> actual, SignalPeek<T> parent) {
 			this.actual = actual;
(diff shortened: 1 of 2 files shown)


src/main/java/reactor/core/publisher/FluxPeekFuseable.java
```diff
@@ -258,7 +263,7 @@ public T poll() {
 					throw Exceptions.propagate(Operators.onOperatorError(s, e, v));
 				}
 			}
-			if (v == null && sourceMode == SYNC) {
+			if (v == null && (d || sourceMode == SYNC)) {
```
smaldini: I'd reverse the condition to only touch volatile if necessary if we go that route

akarnokd: Doesn't work, you have to read done then q.isEmpty. Reversed and you may see an empty queue but when you see done after that, the queue may have received more values from upstream. Remember that poll may run async to onNext/onComplete.

smaldini: but you read only at that moment, is that due to memory barrier ? In that case what about poll in TakeFuseable  for instance ?

akarnokd: This avoids a volatile read for sync use but doubles the code size, reduces inline chances, etc.
```java
		    if (sourceMode == SYNC) {
	            T v = s.poll();
	            if (v != null && parent.onNextCall() != null) {
	                try {
	                    parent.onNextCall().accept(v);
	                }
	                catch (Throwable e) {
	                    throw Exceptions.propagate(Operators.onOperatorError(s, e, v));
	                }
	            }
	            if (v == null) {
	                Runnable call = parent.onCompleteCall();
	                if (call != null) {
	                    call.run();
	                }
	                call = parent.onAfterTerminateCall();
	                if (call != null) {
	                    call.run();
	                }
	            }
	            return v;
		    }
            boolean d = done;
            T v = s.poll();
            if (v != null && parent.onNextCall() != null) {
                try {
                    parent.onNextCall().accept(v);
                }
                catch (Throwable e) {
                    throw Exceptions.propagate(Operators.onOperatorError(s, e, v));
                }
            }
            if (v == null && d) {
                Runnable call = parent.onCompleteCall();
                if (call != null) {
                    call.run();
                }
                call = parent.onAfterTerminateCall();
                if (call != null) {
                    call.run();
                }
            }
            return v;
```

smaldini: Your solution is good anyway, I have tried a similar effect anyway when I worked from the MonoSusbcriber directly (it had to double touch a volatile on poll for checking complete). But as discussed offline, this is not unique behavior to MonoSubscriber, PublishOn, UnicastProcessor and other who use onNext(null) might onComplete just after.

smaldini:
I'm going to squash merge and separately commit to reformat the space

smaldini:
Reference #230

PR #273, opened 2016-11-27
Fixes #272 - Exception in doOnError does not signal downstream error.

### LINKED ISSUE #272: Exception in doOnError does not signal downstream error.
My understanding is that an error in doOnError (or any other Peek operator) should cause an error to be signaled downstream.  This doesn't seem to be happening:

This fails:
```
    Mono.just(1)
        .doOnNext(i -> {
          throw new IllegalArgumentException();
        })
        .doOnError(e -> {
          throw new IllegalStateException(e);
        })
        .doOnError(e -> System.out.println("ERROR"))
        .subscribe(assertSubscriber);

    assertSubscriber
        .await()
        .assertError(IllegalStateException.class)
        .assertNotComplete();
```

dfeist:


dfeist:
These changes have been incorporated into https://github.com/reactor/reactor-core/pull/269 instead.

Issue #274, opened 2016-11-27
Inconsistent handling of RejectedExecutionException in publishOn during subscribe/block.


dfeist:
When an Executor/ExecutorService throws a `RejectedExecutionException` as part of the subscribe/request the way in which it is handled differs between `ExecutorScheduler` and `ExecutorServiceScheduler`

```
i. Mono.just(1).publishOn(Schedulers.fromExecutor(executor)).block();
ii. Mono.just(1).publishOn(Schedulers.fromExecutorService(executor)).block();

```

When using `ExecutorScheduler` the example above **hangs** whereas with the `ExecutorServiceScheduler` the block()/subscribe() method throws `RejectedExecutionException`.

I've seen this with both Flux and Mono with cold sources.

dfeist:
@simonbasle @smaldini FlatMap behaviour changed as part of this work.  If flapMap concurrency is higher than the number of available threads, and executor buffer is not unlimited then execution fails with a bubbled RejectedExecutionException, whereas prior to this change these errors were ignored and flapMap would run with maximum possible concurrency based on availability to submit work. 

(That said, there was an anomaly in previous behaviour where some events got lost)

dfeist:
Not sure what best/correct behaviour is but these are my observations:
**RxJava 2:** catch exception and call hook and log, drop event but continue with next.
**Reactor Pre #274:** drop rest of intial demand then continue, on each failure drop rest of batch (if any) and continue. No logging/error.
**Reactor Post #274:** Fail with bubbled exception, do not continue with next.

PR #276, opened 2016-11-28
Fixes #275 Poor handling of RejectedExecutionException during processing

### LINKED ISSUE #275: Poor handling of RejectedExecutionException during processing
If an Executor throws a `RejectedExecutionException` for some reason during processing of a data signal it is not handled resulting hung requests or loss of signals.  This appears to be an issue with both ExecutorScheduler and ExecutorServiceScheduler.

I understand that the expected behaviour is potentially questionable because passing a RejectedExecutionException downstream in a way violates the expectation that publishOn always signals on the scheduler.  *But,* I'd argue that a more pragmatic solution is needed here to avoid loss of signals, and `1.4 If a Publisher fails it MUST signal an onError.` should be respected so that if `FluxPublishOn` (which is a `Publisher`) fails, due to an issue with its scheduler, then it must signal this error downstream.  Also, for practical reasons such as message loss and hung subscribers, I think respecting this is more important than violating the expectation regarding the thread it's signalling on.

To make things clearer, it may make sense to signal a specific Reactor exception when this happens with the RejectedExecutionException as its cause.

```
Mono.just(1).publishOn(fromExecutor(executor)).doOnNext(s -> {
	try {
		latch.await();
	}
}).publishOn(fromExecutor(executor)).subscribe(assertSubscriber);

executor.shutdownNow();

assertSubscriber
		.await()
		.assertError(RejectedExecutionException.class)
		.assertNotComplete();
```

In this example, when the executor is shutdown the `onNext` exits throwing an `InterruptedException` which is signalled downstream.  When the second `publishOn` attempts to schedule the onError signal, a `RejectedExecutionException` is thrown but this is not handled in any way and the `assertSubscriber.await()` just hangs as neither error/complete signal ever arrives.

dfeist:
This pull request adds tests for the issue described in #275 as well as a potential fix.  For now, I have only tested/fixed MonoPublishOn behaviour and not touched Flux until I get some feedback.  Once I get some feedback on this, I can add tests for Flux too.

Not catching the RejectedExcutedException also resolves the inconsistency detailed in #274.

### DIFF
--- a/src/main/java/reactor/core/publisher/MonoPublishOn.java
+++ b/src/main/java/reactor/core/publisher/MonoPublishOn.java
@@ -16,6 +16,7 @@
 
 package reactor.core.publisher;
 
+import java.util.concurrent.RejectedExecutionException;
 import java.util.concurrent.atomic.AtomicReferenceFieldUpdater;
 
 import org.reactivestreams.*;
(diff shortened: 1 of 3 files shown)


akarnokd:
There could be a graceful shutdown where there is an ongoing emission still and you call onError concurrently with it. This is not solvable without introducing large overhead on a critical operator.

dfeist:
@akarnokd Did you see my thoughts in the issue #275 ? We are probably best to discuss there as this PR might go away..

PR #284, opened 2016-12-05
Fixes #283 Update .transform() in ParrellelFlux to support Function<P…

### LINKED ISSUE #283: Add .transformEach() in ParallelFlux to support Function<Publisher<U>, Publisher<V>>

dfeist:
…ublisher<U>, Publisher<V>>

### DIFF
--- a/src/main/java/reactor/core/publisher/ParallelFlux.java
+++ b/src/main/java/reactor/core/publisher/ParallelFlux.java
@@ -909,8 +909,8 @@ public abstract class ParallelFlux<T> implements Publisher<T> {
 	 *
 	 * @return the {@link ParallelFlux} returned by the function
 	 */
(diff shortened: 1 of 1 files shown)


akarnokd:
This makes no sense. Returning a Publisher indicates you want to quit the parallel world but then you cast it back to ParallelFlux?

dfeist:
@akarnokd Yes that make no sense, that's because i pushed the wrong thing, wasn't thinking. Look again..

akarnokd:
The problem is that now you apply the default parallelization level to the Publisher source. There is `as` for general purpose conversion and `transform` for parallel-parallel conversion.

dfeist:
@akarnokd I assumed from() had an instance of check, like Flux does. But even so, it would mean that whenever is Publisher from transform is returned parallel kicks in all over again when it shouldn't really.

I'll close out the issue and PR.

PR #298, opened 2016-12-13
Fixes #275 Poor handling of RejectedExecutionException during processing

### LINKED ISSUE #275: Poor handling of RejectedExecutionException during processing

dfeist:
@smaldini This:
- Fixes the tests that broke with your last commit. (due to missing data signal in one usage, and `ExecutorServiceScheduler` throwing `RejectedExecutionException`)
- Adds tests for FluxPublishOn (copied from MonoPublishOnTest)
- Ensures data signal isn't lost in FluxPublishOn.  My quick solution isn't that pretty, but I do believe that whenever `onOperatorError` is called, if it's possible to pass data signal then it should be a must be passed.

Going to run all tests and check this doesn't break anything else now..

### DIFF
--- a/src/main/java/reactor/core/publisher/FluxPublishOn.java
+++ b/src/main/java/reactor/core/publisher/FluxPublishOn.java
@@ -242,7 +242,10 @@ final class FluxPublishOn<T> extends FluxSource<T, T> implements Loopback, Fusea
 				return;
 			}
 			if (sourceMode == Fuseable.ASYNC) {
-				trySchedule();
+				if (!trySchedule()) {
+					throw Exceptions.bubble(Operators.onOperatorError(this,
+							new RejectedExecutionException("Scheduler unavailable"), t));
+				}
 				return;
 			}
 			if (!queue.offer(t)) {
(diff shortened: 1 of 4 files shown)


smaldini:
Can you return Cancellation, boolean would allocate useless mem here.

dfeist:
Pushed, this better?

smaldini:
Very nice thanks for this Daniel!

PR #396, opened 2017-01-28
use computeIfAbsent instead of verbose null check

kamilszymanski:


smaldini:
Nice polish thanks a lot @kamilszymanski

PR #427, opened 2017-02-09
Try WorkQueueProcessor with PublishOn and Retry

dfeist:
@smaldini Can you take a look at these failing tests and tell me what behaviour you expect?


smaldini:
Directly included in #432

Issue #430, opened 2017-02-09
publish() signals old subscriber after retry


dfeist:
This is fixed in https://github.com/reactor/reactor-core/pull/427

@smaldini this should get into 3.0.5 also really, my fix probably needs improving and a test case adding though..

dfeist:
@smaldini this causes signal to be skipped dropping an event on resubscribe, need to reopen and fix.

PR #431, opened 2017-02-10
Add retry tests for #430 (with publishOn case failing)

dfeist:


smaldini:
Directly included in #432

PR #456, opened 2017-03-02
Fix #314 Add onRequest for FluxSink and Flux#push

### LINKED ISSUE #314: Improved support for push/pull Publishers (via request callback on FluxSink?)
Currently `Flux.create` provides support for hot publishers that are unable to respond to demand.   and simply emit events. `Flux.generate` provides support for cold publishers that respond to demand and are able to guarantee a `next()` for each `request()`.

Not all publishers fit into these two categories IMO, and there are some publishers that will be able to respond to demand (and this is preferred) but won't always have a message available to `next()` in response to this demand and may need to instead accumulate this demand until it is able to satisfy it thus has a contract whereby they emit at will up to pending demand.  I see this as a push/pull hot publisher used when the source system supports some kind of basic flow control.

One option that I and @smaldini discussed was adding a request callback to `FluxSink`, there might be other better approach, though.

`Flux.create(s -> s.onRequest(demand -> s.next(xx)) )`

Another way of supporting this might be using something similar to the following, where the consumer implementation can emit events when they become available, up to a maximum of `long` times.

`Flux.generate(BiConsumer<Long, SynchronousSink<T>> generator)`

rajinisivaram:
Add onRequest to FluxSink and MonoSink so that publishers like messaging clients which use push/pull model can respond to demand when messages are available.

### DIFF
--- a/src/main/java/reactor/core/publisher/FluxSink.java
+++ b/src/main/java/reactor/core/publisher/FluxSink.java
@@ -16,6 +16,8 @@
 
 package reactor.core.publisher;
 
+import java.util.function.LongConsumer;
+
 import org.reactivestreams.Subscriber;
 import reactor.core.Cancellation;
 import reactor.core.Disposable;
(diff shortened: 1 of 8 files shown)


src/main/java/reactor/core/publisher/FluxSink.java
```diff
@@ -65,6 +67,14 @@
 	FluxSink<T> serialize();
 
 	/**
+	 * Attaches a {@link LongConsumer} to this {@link FluxSink} that will observe any request
+	 * to this sink.
+	 * @param consumer the consumer to invoke on each request
+	 * @return {@link FluxSink} that observes requests
+	 */
+	FluxSink<T> onRequest(LongConsumer consumer);
```
akarnokd: I'm not in favor of this change for the following reasons:

  - It can't be ruled out that multiple concurrent request() call happens from downstream, triggering the execution of the consumer body that is likely to try and emit concurrently into onNext().
  - Even if there is no concurrent requesting, responding to request must be reentrant safe otherwise a request(1) triggers an onNext() that triggers a request(1) recursively until stack overflow.
  - Create bridges the typically non-backpressured, callback-style world with the backpressured one by providing explicit safeguards against overproduction. Combining the same API with reacting to requests is unnecessary conflation in my mind.
  - The use case actually prompts for implementing a `Flux`/`Mono` directly and following the operator writing guidelines because they are designed for reacting to requests in the first place.

rajinisivaram: @akarnokd @smaldini Thank you for the review. I have updated the code to use a serialized sink when `onRequest` is invoked and the `onRequest` consumer is now a `BiConsumer` which passes in the (serialized) sink as well. Hopefully that addresses some of the concerns above. `MonoSink` has been updated to be consistent with `FluxSink`. Added more unit tests for `FluxCreate` to test with multiple threads.

rajinisivaram:
@smaldini I have updated the PR as we discussed. `Flux#create` now creates a serialized sink on which `onRequest` can be invoked. `Flux#push` generates a non-serialized sink (same as `create` used to be) for push-only use cases. The `serialize()` method has been deprecated. Have updated the docs too.

src/main/java/reactor/core/publisher/FluxSink.java
```diff
@@ -60,11 +62,27 @@
...
+	 * {@link Flux#create(java.util.function.Consumer, OverflowStrategy)} can be
+	 * used to create a serialized {@link FluxSink}.
 	 * @return the serialized {@link FluxSink}
 	 */
+	@Deprecated
 	FluxSink<T> serialize();
 
 	/**
+	 * Attaches a {@link LongConsumer} to this {@link FluxSink} that will be notified of
+	 * any request to this sink. This method is supported only on serialized sinks created
```
dfeist: Isn't this really recognizing there there should probably be a push-only `Sink` (used with `push`) and then something else that extends it and adds on `onRequest()` behaviour?  Having a method that is supported in one scenario, but not in another scenario isn't ideal.

rajinisivaram: @dfeist Thank you for the review. After discussion with @smaldini I have updated the PR to handle `onRequest` for push-only sinks as well. For push-only `FluxSink`, the `onRequest` consumer is invoked once within `onRequest` with a request of `Long.MAX_VALUE` avoiding serialization issues. This enables similar code to be used for handling both push-only and push-pull sinks.

rajinisivaram:
@smaldini I have rebased the PR and reverted the change to `setCancellation`. Can you take a look? Thanks.

PR #615, opened 2017-05-22
fix #611 Include OnAssembly description in Flux/MonoSource#toString

### LINKED ISSUE #611: FluxOnAssembly should override FluxSource#toString if a description is provided
`FluxSource` provides a default `toString` implementation. if a description is provided to a `FluxOnAssembly` instance, its `toString` should included that description

rajinisivaram:


rajinisivaram:
@simonbasle Thank you for the review, updated.

PR #617, opened 2017-05-22
scheduler based on fork join pool

osi:
this is a Scheduler implementation that uses a ForkJoinPool and a
ScheduledExecutorService for scheduling.

it differs from the Parallel scheduler in that Workers are not pre-bound
to a specific thread in the pool. each Worker maintains its own queue of
tasks and when it has tasks to execute it will grab a thread from the
pool to use to execute it.

since the ForkJoinPool does not support time-based scheduling, a
single-threaded ScheduledExecutorService is used to enqueue tasks in the
main FJ pool as necessary.

standard ThreadFactory is only used for the single SES thread.

a custom FJP thread factory is used to rename FJP threads

### DIFF
--- a/reactor-core/src/main/java/reactor/core/scheduler/ForkJoinPoolScheduler.java
+++ b/reactor-core/src/main/java/reactor/core/scheduler/ForkJoinPoolScheduler.java
@@ -1,0 +1,372 @@
+/*
+ * Copyright (c) 2011-2017 Pivotal Software Inc, All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package reactor.core.scheduler;
+
+import java.util.ArrayDeque;
+import java.util.Queue;
+import java.util.concurrent.Executor;
+import java.util.concurrent.ForkJoinPool;
+import java.util.concurrent.ForkJoinTask;
+import java.util.concurrent.Future;
+import java.util.concurrent.RejectedExecutionException;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.ScheduledFuture;
+import java.util.concurrent.ScheduledThreadPoolExecutor;
+import java.util.concurrent.ThreadFactory;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicLong;
+import java.util.function.BooleanSupplier;
+
+import reactor.core.Disposable;
+import reactor.util.Logger;
+import reactor.util.Loggers;
+
+final class ForkJoinPoolScheduler implements Scheduler {
+
+	private final ForkJoinPool             pool;
+	private final ScheduledExecutorService scheduler;
+	/**
+	 * Construct a new instance
+	 *
+	 * @param parallelism Parallelism. Number of fork-join pool threads. See {@link
+	 * ForkJoinPool}
+	 * @param workerThreadFactory Thread factory for fork-join worker threads
+	 * @param threadFactory Thread factory to use for constructing the single threaded
+	 * {@link ScheduledExecutorService} used for time-based scheduling.
+	 */
+	ForkJoinPoolScheduler(int parallelism,
+			ForkJoinPool.ForkJoinWorkerThreadFactory workerThreadFactory,
+			ThreadFactory threadFactory) {
+		pool = new ForkJoinPool(parallelism,
+				workerThreadFactory,
+				this::uncaughtException,
+				true);
+		scheduler = new ScheduledThreadPoolExecutor(1, threadFactory);
+	}
+
+	@Override
+	public Scheduler.Worker createWorker() {
+		return new Worker(pool, scheduler);
+	}
+
+	@Override
+	public void dispose() {
+		Schedulers.executorServiceShutdown(pool, FORK_JOIN_POOL);
+		Schedulers.executorServiceShutdown(scheduler, FORK_JOIN_POOL_TIMER);
+	}
+
+	@Override
+	public boolean isDisposed() {
+		return pool.isShutdown();
+	}
+
+	@Override
+	public Disposable schedule(Runnable runnable) {
+		try {
+			return new DisposableForkJoinTask(pool.submit(runnable));
+		}
+		catch (RejectedExecutionException ignored) {
+			return REJECTED;
+		}
+	}
+
+	@Override
+	public Disposable schedule(Runnable task, long delay, TimeUnit unit) {
+		if (delay == 0) {
+			return schedule(task);
+		}
+		PendingTask pendingTask = new PendingTask(task, NO_PARENT);
+		ScheduledFuture<?> future =
+				scheduler.schedule(new ExecuteTask(pendingTask, pool), delay, unit);
+		return new DisposableFuture(future, pendingTask);
+	}
+
+	@Override
+	public Disposable schedulePeriodically(Runnable task,
+			long initialDelay,
+			long period,
+			TimeUnit unit) {
+		PendingTask pendingTask = new PendingTask(task, NO_PARENT);
+		ScheduledFuture<?> future =
+				scheduler.scheduleAtFixedRate(new ExecuteTask(pendingTask, pool),
+						initialDelay,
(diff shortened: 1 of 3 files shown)


osi:
The tests are copies from the Parallel scheduler. I didn't add any tests specific to this scheduler.

I also did not implement restart of the scheduler. It starts at construction and is unusable after disposal.

osi:
hrm, looks like i rebased onto sand, "reactor.core.publisher.MonoCreateTest > sinkApiErrorAfterErrorBubblesAndDrops FAILED"

simonbasle:
@osi I fixed the test failing in master. This is actually a fix for #233, do you mind rebasing on master again and changing the commit message with `fix #233 ` as a suffix to reflect that? (hopefully that will also unblock codecov)

src/main/java/reactor/core/scheduler/ForkJoinPoolScheduler.java
```diff
@@ -0,0 +1,355 @@
...
+import java.util.concurrent.ScheduledThreadPoolExecutor;
+import java.util.concurrent.ThreadFactory;
+import java.util.concurrent.TimeUnit;
+import java.util.function.BooleanSupplier;
+
+import reactor.core.Disposable;
+import reactor.util.Logger;
+import reactor.util.Loggers;
+
+public class ForkJoinPoolScheduler implements Scheduler {
```
simonbasle: scheduler implementations from reactor-core are usually **package-protected** and `final`.
once it is not public, you'll need to add the `forkJoinPool(...)` default factory to `Schedulers`, as well as a `newForkJoinPool(...)` variant.

src/main/java/reactor/core/scheduler/ForkJoinPoolScheduler.java
```diff
@@ -0,0 +1,355 @@
...
+
+	private final ForkJoinPool             pool;
+	private final ScheduledExecutorService scheduler;
+
+	public ForkJoinPoolScheduler(int parallelism, ThreadFactory threadFactory) {
```
simonbasle: the constructor can also be package-protected. I'd still put javadoc on it to make a note that `threadFactory` is only applicable to the `ScheduledExecutorService` that is used internally for timer-based tasks.

src/main/java/reactor/core/scheduler/ForkJoinPoolScheduler.java
```diff
@@ -0,0 +1,355 @@
...
+	}
+
+	@Override
+	public Scheduler.Worker createWorker() {
+		return new Worker(pool, scheduler);
+	}
+
+	@Override
+	public void dispose() {
+		Schedulers.executorServiceShutdown(pool, "fj sidecar scheduler");
```
simonbasle: The dispose should shutdown both `pool` and `scheduler`.

Use `Schedulers.executorServiceShutdown` in both cases, with the correct string descriptor.

For the string descriptor, add 2 constants in `Schedulers` (next to `ELASTIC`) and use these

```
static final String FORK_JOIN_POOL               = "forkJoinPool"; // Main fork join pool
static final String FORK_JOIN_POOL_TIMER  = "forkJoinPoolTimer"; // Companion to the fork join pool for timed tasks
```

src/main/java/reactor/core/scheduler/ForkJoinPoolScheduler.java
```diff
@@ -0,0 +1,355 @@
...
+
+		@Override
+		public void dispose() {
+			if (shutdown) {
+				return;
+			}
+
+			shutdown = true;
+
+			synchronized (this) {
```
simonbasle: didn't you mean to use the `lock` object here as well?

src/test/java/reactor/core/scheduler/ForkJoinPoolSchedulerTest.java
```diff
@@ -0,0 +1,117 @@
...
+import java.util.concurrent.atomic.AtomicLong;
+
+import org.junit.Test;
+import reactor.core.publisher.Flux;
+import reactor.core.publisher.Mono;
+import reactor.test.StepVerifier;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class ForkJoinPoolSchedulerTest extends AbstractSchedulerTest {
```
simonbasle: I think this indeed needs more tests, notably on what happens when tasks are scheduled ahead of time then the Scheduler is disposed

osi:
Thanks for the feedback! I wasn't sure if a new scheduler would be admitted, so I didn't do all the integration with `Schedulers`. I'll rebase and incorporate your feedback.

src/main/java/reactor/core/scheduler/ForkJoinPoolScheduler.java
osi: The `lock` is to protect the `executing` flag and the `tasks` queue. 

`shutdown` is a volatile since it's only set when disposing

osi:
wrt `synchronized` in the worker, I had opted to use that to keep things simpler. the granularity of the lock seemed small enough that contention should be fairly low (it would only be on concurrent scheduling of tasks for the worker, and the removal of tasks from the queue). similar to how `ConcurrentHashMap` has some `synchronized` blocks once you get fine-grained enough

src/main/java/reactor/core/scheduler/ForkJoinPoolScheduler.java
simonbasle: yes, yet here you synchronize on `this` rather than `lock` before clearing the `tasks` ;)

simonbasle:
@osi did you have a look at the existing implementation in [reactive-streams-commons](https://github.com/reactor/reactive-streams-commons/blob/master/src/main/java/rsc/scheduler/ForkJoinScheduler.java)?

akarnokd:
Isn't there a wrapper for `ExecutorService`-based schedulers in Reactor already? There is only a marginal benefit going native with a FJ service.

simonbasle:
@akarnokd yes there is, it is true the benchmark was brought up in the #233 issue. do you mean that the FJP would be used through the wrapper? the only limitation is that it isn't a `ScheduledExecutorService`...

osi:
@simonbasle I hadn't seen the implementation in commons. That one is very similar to the parallel scheduler where a `Worker` is bound to a single thread and doesn't support the time-based methods

@akarnokd yes, one could use a `ForkJoinPool` as an `ExecutorService`, but then it wouldn't be time-capable. it is also unclear how the `ExecutorServiceScheduler` workers guarantee non-concurrency of tasks if the underlying executor service has multiple threads.

The main benefit I'm trying to bring is not tying a `Worker` to a specific thread. If that's better as an enhancement to the parallel scheduler (and not using the FJ Pool), that's okay with me.

src/main/java/reactor/core/scheduler/ForkJoinPoolScheduler.java
osi: doh! yes, good catch.

osi:
(fwiw, I've been using this instead of the `parallel` scheduler with no ill effects)

smaldini:
Sorry for being late to that party. I'm not ultra keen at first sight on adding a global ForkJoin dedicated global registry versus adding an automatic support for fromExecutorService as in #233 (and using an instanceOf check).

osi:
@smaldini the primary difference in this scheduler is that `Worker` instances are not bound to a specific thread. in my workload not pre-binding them to a thread gave much better utilization of available cores when under load. 

the `ExecutorServiceScheduler` doesn't ensure that tasks on `Worker`s that it creates are run in FIFO order if the supplied `ExecutorService` has multiple threads.

I also think that a scheduler that behaves like this one would be better suited as the default 'parallel' scheduler.

If there's still opposition to merging, is the team amiable to a PR that would allow this to exist as an add-on?

smaldini:
Hey @osi can we remove the global registry changes and just keep the internal Scheduler for now ?
What would be cool is that fromExecutorService auto delegate to it if executorServ is a ForkJoinPool, plus Schedulers.newForkJoin()

osi:
yes no problem with no global registry. that's the current state of this PR, actually. one must create any fork join schedulers explicitly. 

i'll add support to fromExecutorService as you envision too (tomorrow)

-- 
peter royal - (on the go)

smaldini:
Oh i misread the static in Schedulers. could you move them from Schedulers to ForkJoin impl directly https://github.com/reactor/reactor-core/pull/617/files#diff-60668480c66e2171abc4a21085822ae0R316

osi:
no prob 

-- 
peter royal - (on the go)

osi:
regarding delegation in `fromExecutorService`:

- How to handle the `interruptOnCancel` flag? Don't override the overloaded method? Or don't dispatch if it is true?
- A bare `ForkJoinPool` doesn't support time. Should it follow the pattern of `ExecutorServiceScheduler` and not support time? Or should there be another `fromXXX` method that takes a `ForkJoinPool` and a `ScheduledExecutorService` ?

I've got no strong opinion on any of that, so I'm happy to implement whatever the team thinks is best.

PR #618, opened 2017-05-22
Use BiPredicate for DistinctUntilChanged

osi:
Addressing #608, extend the functionality of distinct-until-changed to
accept a BiPredicate<T,T> to determine if the event is "different" from
the last event. The prior key-based system is still supported as an
overlay.

### DIFF
--- a/reactor-core/src/main/java/reactor/core/publisher/FluxDistinctUntilChanged.java
+++ b/reactor-core/src/main/java/reactor/core/publisher/FluxDistinctUntilChanged.java
@@ -18,11 +18,11 @@ package reactor.core.publisher;
 import java.util.Objects;
 import java.util.function.BiPredicate;
 import java.util.function.Function;
+import javax.annotation.Nullable;
 
 import org.reactivestreams.Subscriber;
 import org.reactivestreams.Subscription;
 import reactor.core.Fuseable.ConditionalSubscriber;
-import javax.annotation.Nullable;
 
 /**
  * Filters out subsequent and repeated elements.
@@ -111,9 +111,26 @@ final class FluxDistinctUntilChanged<T, K> extends FluxSource<T, T> {
 				return true;
 			}
 
-			if (null != lastKey && keyComparator.test(lastKey, k)) {
+			if (null == lastKey) {
+				lastKey = k;
+				actual.onNext(t);
+				return true;
+			}
+
+			boolean equiv;
+
+			try {
+				equiv = keyComparator.test(lastKey, k);
+			}
+			catch (Throwable e) {
+				onError(Operators.onOperatorError(s, e, t));
+				return true;
+			}
+
+			if (equiv) {
 				return false;
 			}
+
 			lastKey = k;
 			actual.onNext(t);
 			return true;
@@ -220,9 +237,25 @@ final class FluxDistinctUntilChanged<T, K> extends FluxSource<T, T> {
 				return true;
 			}
 
-			if (null != lastKey && keyComparator.test(lastKey, k)) {
+			if (null == lastKey) {
+				lastKey = k;
+				return actual.tryOnNext(t);
+			}
+
+			boolean equiv;
+
+			try {
+				equiv = keyComparator.test(lastKey, k);
+			}
+			catch (Throwable e) {
+				onError(Operators.onOperatorError(s, e, t));
+				return true;
+			}
+
+			if (equiv) {
 				return false;
 			}
+
 			lastKey = k;
 			return actual.tryOnNext(t);
 		}
--- a/reactor-core/src/main/java/reactor/core/publisher/FluxDistinctUntilChanged.java
+++ b/reactor-core/src/main/java/reactor/core/publisher/FluxDistinctUntilChanged.java
@@ -16,6 +16,7 @@
 package reactor.core.publisher;
 
 import java.util.Objects;
+import java.util.function.BiPredicate;
 import java.util.function.Function;
 
 import org.reactivestreams.Subscriber;
@@ -32,11 +33,15 @@ import javax.annotation.Nullable;
  */
 final class FluxDistinctUntilChanged<T, K> extends FluxSource<T, T> {
 
-	final Function<? super T, K> keyExtractor;
+	final Function<? super T, K>            keyExtractor;
+	final BiPredicate<? super K, ? super K> keyComparator;
 
-	FluxDistinctUntilChanged(Flux<? extends T> source, Function<? super T, K> keyExtractor) {
+	FluxDistinctUntilChanged(Flux<? extends T> source,
+			Function<? super T, K> keyExtractor,
+			BiPredicate<? super K, ? super K> keyComparator) {
 		super(source);
 		this.keyExtractor = Objects.requireNonNull(keyExtractor, "keyExtractor");
+		this.keyComparator = Objects.requireNonNull(keyComparator, "keyComparator");
 	}
 
 	@Override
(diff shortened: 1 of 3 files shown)


simonbasle:
Hey @osi, thanks for the PR. Unfortunately, I don't think this is the right approach:

1. This should be a separate operator and not modify the code of the original one
2. As I said in #608 I think this is closer to `takeUntil` than to `distinctUntilChanged`
3. In any case, the `Flux`-level API should use a method name with a similar prefix to the original operator, but **with a dedicated suffix**. Otherwise this will mean 2 methods with the same name that both take a functional interface parameter, and that is a situation we are actively trying to avoid (notably due to the ambiguity it creates in languages like Kotlin)

osi:
no problem. i'll rework it into a new operator.  

naming wise, i called this functionality "jitter" in my codebase previously. a variant on "takeWhile" sounds reasonable

-- 
peter royal - (on the go)

simonbasle:
what about `takeUntilDiff` or `takeUntilDifference`?

osi:
i suppress repeated elements where the difference between the last emitted element is too small. since this sometimes happens on a complex object, I represent it as a BiPredicate.

a source sequence might look like:

1. 1.0
2. 1.01
3. 1.1
4. 1.09
5. 1.4
6. 1.7

If the predicate was `Math.abs(b-a) >= 0.1` then elements 2 and 4 would not be seen downstream.

That's why I started doing it on top of `distinctUntilChanged`, as sometimes I just use `Objects::equals` as my predicate, which is what `distinctUntilChanged` does if there's no key supplier (essentially, since it uses the hashCode as a key supplier)

osi:
so I'm not sure that there's a better operator name?

in RxJava, this is also `distinctUntilChanged` ... they have the two overloaded methods.

if the desire is to only have one method with a functional interface, I'd propose renaming the key-based one?

http://reactivex.io/RxJava/2.x/javadoc/io/reactivex/Flowable.html#distinctUntilChanged(io.reactivex.functions.BiPredicate)

src/main/java/reactor/core/publisher/Flux.java
```diff
@@ -3326,7 +3327,26 @@ public final T blockLast(Duration timeout) {
 	 * the same key (yet element keys can repeat in the overall sequence)
 	 */
 	public final <V> Flux<T> distinctUntilChanged(Function<? super T, ? extends V> keySelector) {
-		return onAssembly(new FluxDistinctUntilChanged<>(this, keySelector));
+		Objects.requireNonNull(keySelector, "keySelector");
+		Function<? super T, ? extends V> nullSafeKeySelector =
+				keySelector.andThen(key -> Objects.requireNonNull(key,
```
smaldini: should be done inside the target operator if possible keeping the keySelector argument as before. Rx does keep both too

smaldini:
After a rebase (if possible) and addressing comment it should be candidate to be merged indeed.

simonbasle:
@smaldini @osi ok with integrating that as `distinctUntilChanged` iff:
 1. the variant with the comparator is a separate implementation
 2. the API is exposed as `distinctUntilChanged(Function<T, V> keySelector, BiPredicate<V, V> distinctPredicate)`. That is, an even more specialized version of distinctUntilChanged with a different number of arguments.
 3. the `distinctPredicate` should either be non-null or marked as `@Nullable`, in which case it would chose the new implementation only if predicate is not null, classic implementation otherwise.

osi:
@simonbasle cool i'm fine with your second constraint as the behavior i seek can be achieved with a `keySelector` of `Function.identity()`

i'll re-work the PR.

reactor-core/src/main/java/reactor/core/publisher/FluxDistinctUntilChanged.java
```diff
@@ -103,8 +111,7 @@ public boolean tryOnNext(T t) {
 				return true;
 			}
 
-			if (Objects.equals(lastKey, k)) {
-				lastKey = k;
+			if (null != lastKey && keyComparator.test(lastKey, k)) {
 				return false;
```
simonbasle: `lastKey` should still be initialized. Also this block would need to be protected against the keyComparator throwing an exception, like what is done just above.

reactor-core/src/main/java/reactor/core/publisher/FluxDistinctUntilChanged.java
```diff
@@ -210,8 +220,7 @@ public boolean tryOnNext(@Nullable T t) {
 				return true;
 			}
 
-			if (Objects.equals(lastKey, k)) {
-				lastKey = k;
+			if (null != lastKey && keyComparator.test(lastKey, k)) {
```
simonbasle: `lastKey` must still be initialized to `k`. This block must also be protected from the comparator throwing an exception (like what is done just above).

reactor-core/src/test/java/reactor/core/publisher/FluxDistinctUntilChangedTest.java
```diff
@@ -163,6 +169,19 @@ public void withKeySelector() {
...
+		    .distinctUntilChanged(Function.identity(), (a,b) -> b - a < 4)
+		    .subscribe(ts);
+
+		ts.assertValues(1, 5, 9)
+		  .assertComplete()
+		  .assertNoError();
+	}
 
 	@Test
 	public void keySelectorThrows() {
```
simonbasle: also add a similar test `keyComparatorThrows()`

smaldini:
Too bad we can't edit PR directly !

reactor-core/src/main/java/reactor/core/publisher/FluxDistinctUntilChanged.java
osi: When `lastKey` is null (first time around), I do not invoke the comparator as we know that the value is distinct. I broke apart the conditional to make that more explicit when adding the error handling

osi:
re: editing PR, i think you'd have to do changes while merging (vs merging directly). or i can add you as a collaborator in my workspace

simonbasle:
there's a checkbox when doing the PR out of your fork in a branch, allow collaborators to edit. This gives us (maintainers of the source repo) to edit the branch. I think this is now on by default.

osi:
oh, i didn't know - i'll make sure i keep an eye out for it next PR

reactor-core/src/main/java/reactor/core/publisher/FluxDistinctUntilChanged.java
```diff
@@ -103,10 +111,26 @@ public boolean tryOnNext(T t) {
...
+
+			try {
+				equiv = keyComparator.test(lastKey, k);
+			}
+			catch (Throwable e) {
+				onError(Operators.onOperatorError(s, e, t));
+				return true;
+			}
+
+			if (equiv) {
```
simonbasle: I'm not sure this should have been merged as is. This `equiv` block means that the oldest equivalent key will always be the one compared.

If your criteria is "the 2 keys are equivalent if within +- 2 from each other" (`abs(abs(lastKey) - abs(key)) > 2`) and you get `--3--4--5--2--|->" sequence, it will compare 3 and 4, 3 and 5 then _3 and 2_ and will only output `--3--|->`.

Whereas by always replacing the `lastkey` in the equiv block with `k`, you'd get the following comparisons:
 3 and 4, 4 and 5, 5 and 2 (changed!) => output `--3--2--|->`.

I'm not sure which of the 2 is the correct (or more desirable) behavior @osi @smaldini?

If the current behavior seem like the best, keep it that way. Otherwise we'll need to open a follow up issue.

osi: the current behavior is intentional. 

my use-case is emitting numerical values, and this lets me implement a "jitter" to suppress values that are "too close" to the last observed one.

if you always compared against the last key, and not the last emitted key, then in your example if the sequence kept on increasing by 1 then there would only ever be one emission.

you did find the "degenerate" case of this though, where the values dance around right below the allowable difference.

PR #656, opened 2017-06-05
Fix #609 - scanWith will emit in lockstep w/prior

### LINKED ISSUE #609: scan(initial, accumulator) variant that will accumulate-then-emit
To make `Flux#scan(initial, accumulator)` more useful with infinite streams, it would be nice if it behaved like this:

```
        StepVerifier.create(Flux.<Integer>create(sink -> {
            sink.next(3);
            sink.next(4);
            sink.next(5);
        }).scan(10, (a, b) -> a + b))
                    .expectNext(10, 13, 17, 22)
                    .thenCancel()
                    .verify(Duration.ofMillis(250));
```

If scan is being used for a "live" reduction, the accumulation always lags by one. This confused me for awhile since it seems to contradict the flow diagram at http://projectreactor.io/docs/core/release/api/reactor/core/publisher/Flux.html#scan-A-java.util.function.BiFunction- (at least in how the timing works)

osi:
Modify FluxScanSeed so that the seed is sent
immediately, which allows the accumulator to
run in lockstep with the upstream stage.

This fixes a problem where the accumulated value
lagged the upstream by one.

Emitting the seed in `request` was the easiest solution as it minimized state keeping. I checked the code for precedence in doing this, and the initial stages (ie, `just`), acted like this.

I had to add `skip(1)` for the `scenarios_errorFromUpstreamFailure` tests to pass. (I tried a variation that emitted the seed after calling `request` from upstream but that was additional state-keeping and didn't fix the tests. after reading those tests i don't think they can be fixed w/o modifying the tests to say "it is okay to expect one value, because this operator is also kinda a source")

### DIFF
--- a/reactor-core/src/main/java/reactor/core/publisher/FluxScanSeed.java
+++ b/reactor-core/src/main/java/reactor/core/publisher/FluxScanSeed.java
@@ -72,10 +72,10 @@ final class FluxScanSeed<T, R> extends FluxSource<T, R> {
 	static final class ScanSeedCoordinator<T, R>
 			extends Operators.MultiSubscriptionSubscriber<R, R> {
 
-		final         Supplier<R>                 initialSupplier;
-		private final Publisher<? extends T>      source;
-		private final BiFunction<R, ? super T, R> accumulator;
-		volatile      int                         wip;
+		final    Supplier<R>                 initialSupplier;
+		final    Publisher<? extends T>      source;
+		final    BiFunction<R, ? super T, R> accumulator;
+		volatile int                         wip;
 		long produced;
 		private ScanSeedSubscriber<T, R> seedSubscriber;
 
--- a/reactor-core/src/main/java/reactor/core/publisher/FluxScanSeed.java
+++ b/reactor-core/src/main/java/reactor/core/publisher/FluxScanSeed.java
@@ -21,6 +21,7 @@ import java.util.concurrent.atomic.AtomicIntegerFieldUpdater;
 import java.util.function.BiFunction;
 import java.util.function.Supplier;
 
+import org.reactivestreams.Publisher;
 import org.reactivestreams.Subscriber;
 import org.reactivestreams.Subscription;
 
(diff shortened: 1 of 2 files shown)


reactor-core/src/main/java/reactor/core/publisher/FluxScanSeed.java
```diff
@@ -147,90 +125,74 @@ public void onError(Throwable t) {
...
-						return;
-					}
-
-					// transition to HAS_REQUEST_NO_VALUE
-					long u = Operators.addCap(r, n);
-					if (REQUESTED.compareAndSet(this, r, u)) {
-						s.request(n);
-						return;
-					}
+			if (STATE.compareAndSet(this, 0, SENT_SEED)) {
```
akarnokd: This won't work properly with two concurrent requests because both of them may trigger a concurrent call to `onNext`. The safe solution is to use queue-drain where the queue contains a the initial value to begin with: [FlowableScanSeed](https://github.com/ReactiveX/RxJava/blob/2.x/src/main/java/io/reactivex/internal/operators/flowable/FlowableScanSeed.java)

osi: I see. That's bullet 7 under https://github.com/reactive-streams/reactive-streams-jvm#2-subscriber-code ?

akarnokd: Yes, §2.7 that also affects §1.3.

reactor-core/src/main/java/reactor/core/publisher/FluxScanSeed.java
```diff
@@ -187,12 +193,25 @@ public Object scanUnsafe(Attr key) {
...
+				serializedOnNext(value);
+				return true;
+			}
+			return false;
+		}
+
+		// concurrent calls to request at the start could yield concurrent calls to
+		// onNext. this will have low-to-no contention after the initial state has emitted
+		private synchronized void serializedOnNext(R value) {
```
akarnokd: Should not hold lock while calling onXXX as it may lead to deadlock.

osi: booger. thanks for being patient with me. (trying to avoid a queue! i had spiked out a single-element queue but could see how that could still overflow)

osi:
third time's the charm? i wanted to avoid the queue, since it should be doable without a queue

akarnokd:
It is still prone to various race conditions and a queue is essential for it to work correctly. I suggest porting the RxJava operator; it isn't that bad since it uses atomics and stable-prefetch queue-drain.

osi:
Why is a queue essential? The operator can be built from a concat + mapping operation, and that's essentially what I did. 

The only races I can envision would be in the transition from the initial value to the ongoing accumulation, and the logic there is lifted from what `FluxConcatArray` does. Is there something I'm missing?

akarnokd:
The problem manifests when you have a hot source and concat delays the subscription to that hot source resulting in element skip from this source.

osi:
ah, because there's a delay introduced between when the `scanWith` is subscribed to and when the upstream hot source is subscribed to depending on how much the downstream requests (could be a lot if they request 1 then wait awhile).

with that said, is that delay not allowed? isn't there always a processing delay for subscribing to a hot source? if one doesn't want to risk dropped elements at all, there should be some replay introduced?

akarnokd:
If we don't consider reentrance due to the emission of the initial item then a skip would be legal. I defer to @smaldini & @simonbasle on the decision about it.

reactor-core/src/main/java/reactor/core/publisher/FluxScanSeed.java
```diff
@@ -61,21 +59,96 @@
...
-		catch (Throwable e) {
-			Operators.error(s, Operators.onOperatorError(e));
-			return;
+	}
+
+	static final class ScanSeedCoordinator<T, R>
+			extends Operators.MultiSubscriptionSubscriber<R, R> {
+
+		final         Supplier<R>                 initialSupplier;
+		private final Publisher<? extends T>      source;
```
smaldini: minor stuff align visibility in package scope

osi: the 'source' should be package-private instead of private?

(otherwise for alignment I just let IntelliJ auto-format after importing the code style)

smaldini:
I am very pleased/impressed by this quality PR @osi and its absolutely fantastic to see master @akarnokd following through with guidance. I was aware of the revamp of ScanSeed in rxjava after we released but didn't take time to re-investigate it in full, now we have it here.
I certainly think its worth a trade-off to go without a fixed-prefetch queue and we can monitor the results for a while before porting back the latest solution from rx post reactivestreams-commons. Besides my minor comment this should be good to go.

PR #667, opened 2017-06-14
Introduce Kotlin extensions

sdeleuze:


sdeleuze:
If have any compilation issue in IDEA when running test about some 1.6 versus 1.8 bytecode version mismatch, please set Kotlin bytecode level to 1.8 in your IDEA project settings. I have reported this issue in https://youtrack.jetbrains.com/issue/KT-17492.

PR #676, opened 2017-06-16
RefCountGrace doesn't forget subscribers when terminated, yet still reconnects

### LINKED ISSUE #675: a completed flux that's 'hot' via RefCountGrace fails when using next
```
        VirtualTimeScheduler scheduler = VirtualTimeScheduler.create();
        Duration gracePeriod = Duration.ofMillis(10);
        Flux<String> f = Flux.just("hello world")
                             .replay(1)
                             .refCount(1, gracePeriod, scheduler);

        f.subscribe();

        StepVerifier.create(f.next())
                    .expectNext("hello world")
                    .verifyComplete();

        scheduler.advanceTimeBy(gracePeriod);
```

This throws

```
reactor.core.Exceptions$BubblingException: java.util.concurrent.CancellationException: Disconnected
	at reactor.core.Exceptions.bubble(Exceptions.java:112)
	at reactor.core.publisher.Operators.onErrorDropped(Operators.java:263)
	at reactor.core.publisher.LambdaSubscriber.onError(LambdaSubscriber.java:116)
	at reactor.core.publisher.FluxRefCountGrace$RefCountInner.onError(FluxRefCountGrace.java:203)
	at reactor.core.publisher.FluxReplay$ReplaySubscriber.disconnectAction(FluxReplay.java:1180)
	at reactor.core.publisher.FluxReplay$ReplaySubscriber.dispose(FluxReplay.java:1172)
	at reactor.core.publisher.Disposables.dispose(Disposables.java:147)
	at reactor.core.publisher.FluxRefCountGrace.timeout(FluxRefCountGrace.java:137)
	at reactor.core.publisher.FluxRefCountGrace$RefConnection.run(FluxRefCountGrace.java:165)
	at reactor.test.scheduler.VirtualTimeScheduler.advanceTime(VirtualTimeScheduler.java:297)
	at reactor.test.scheduler.VirtualTimeScheduler.advanceTimeBy(VirtualTimeScheduler.java:205)
Caused by: java.util.concurrent.CancellationException: Disconnected
	at reactor.core.publisher.FluxReplay$ReplaySubscriber.disconnectAction(FluxReplay.java:1177)
	... 29 more
```

that felt like a "safe" sequence of operators

osi:
When the stream is terminated, do not forget all state, just mark as disconnected. We then need to adjust the check for minSubscribers to be >=, as there may be a subscriber to the completed stream, yet we still need to reconnect.

Fixes #675

### DIFF
--- a/reactor-core/src/main/java/reactor/core/publisher/FluxReplay.java
+++ b/reactor-core/src/main/java/reactor/core/publisher/FluxReplay.java
@@ -1129,6 +1129,7 @@ final class FluxReplay<T> extends ConnectableFlux<T> implements Scannable, Fusea
 		}
 
 		@Override
+		@SuppressWarnings("unchecked")
 		public void onError(Throwable t) {
 			ReplayBuffer<T> b = buffer;
 			if (b.isDone()) {
@@ -1142,10 +1143,15 @@ final class FluxReplay<T> extends ConnectableFlux<T> implements Scannable, Fusea
 				for (ReplaySubscription<T> rs : a) {
 					b.replay(rs);
 				}
+
+				synchronized (this) {
+					subscribers = EMPTY;
+				}
 			}
 		}
 
 		@Override
+		@SuppressWarnings("unchecked")
 		public void onComplete() {
 			ReplayBuffer<T> b = buffer;
 			if (!b.isDone()) {
@@ -1155,6 +1161,10 @@ final class FluxReplay<T> extends ConnectableFlux<T> implements Scannable, Fusea
 
 				for (ReplaySubscription<T> rs : a) {
 					b.replay(rs);
+				}
+
+				synchronized (this) {
+					subscribers = EMPTY;
 				}
 			}
 		}
--- a/reactor-core/src/test/java/reactor/core/publisher/FluxRefCountGraceTest.java
+++ b/reactor-core/src/test/java/reactor/core/publisher/FluxRefCountGraceTest.java
@@ -25,6 +25,7 @@ import reactor.core.Scannable;
 import reactor.core.scheduler.Schedulers;
 import reactor.test.StepVerifier;
 import reactor.test.publisher.TestPublisher;
+import reactor.test.scheduler.VirtualTimeScheduler;
 import reactor.test.subscriber.AssertSubscriber;
 
 import static org.assertj.core.api.Assertions.assertThat;
@@ -84,7 +85,7 @@ public class FluxRefCountGraceTest {
 		p.subscribe().dispose();
 		p.subscribe().dispose();
 		p.subscribe().dispose();
-		
+
 		AssertSubscriber<Integer> ts1 = AssertSubscriber.create();
 		p.subscribe(ts1);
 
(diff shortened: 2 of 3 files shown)


osi:
i wasn't sure that this was the best fix either.

in `FluxRefCount`, it never disconnects the source after it has terminated, and that's what i borrowed from.

the problem occurs because after the source has completed for the first subscriber, the `RefConnection` is disposed of (never disposing the source). then when the second subscriber comes along, it never reaches the terminal state of the stream, because it only wants the first element. so then after it cancels the timeout kicks in and disposes upstream, which causes the exception.

osi:
here's an alternate fix that modifies `RefCountInner` as suggested that also passes the new test

akarnokd:
This works with RxJava:

```java
import java.util.concurrent.TimeUnit;

import hu.akarnokd.rxjava2.operators.FlowableTransformers;
import io.reactivex.Flowable;
import io.reactivex.schedulers.TestScheduler;

public class RefCountGrace {

        TestScheduler sch = new TestScheduler();

        Flowable<Integer> f = Flowable.just(1).replay(1)
                .compose(FlowableTransformers.refCount(1, 10, TimeUnit.MILLISECONDS, sch));

        f.subscribe();

        sch.advanceTimeBy(10, TimeUnit.MILLISECONDS);

        f.test().assertResult(1);

        sch.advanceTimeBy(10, TimeUnit.MILLISECONDS);

        f.test().assertResult(1);
    }
}
```

osi:
@akarnokd the `FlowableRefCountTimeout` is nearly identical to what `FluxRefCountGrace` did, so perhaps the problem is in `FluxReplay` then

osi:
alrighty, a fresh stab at this.

fix `FluxReplay` this time so that it empties `subscribers` after `onComplete` and `onError`

I had to fix `FluxReplayTest#cancel` to preserve what I believed the intent of the test to be.

reactor-core/src/main/java/reactor/core/publisher/FluxReplay.java
```diff
@@ -1142,10 +1143,15 @@ public void onError(Throwable t) {
 				for (ReplaySubscription<T> rs : a) {
```
smaldini: Use terminate() instead of a

reactor-core/src/main/java/reactor/core/publisher/FluxReplay.java
```diff
@@ -1156,6 +1162,10 @@ public void onComplete() {
...
```
smaldini: Use terminate() instead of a

reactor-core/src/main/java/reactor/core/publisher/FluxReplay.java
osi: that breaks tests in `FluxReplayTest`. 

setting `subscribers` to `TERMINATED` marks the `ReplaySubscriber` as `isTerminated == true`, which is the state it goes to when disposed. that's not correct for this, as the state is mere "completed".

that prevents subscribers from subscribing after the stream has completed successfully and receiving replayed elements.

smaldini: Which test does it break? it seems that doing that synchronization after is open to concurrent subscribe(Subscriber) and EMPTY termination.

PR #759, opened 2017-07-26
fix #730 Use Flux instead of GroupedFlux for FluxWindowPredicate

### LINKED ISSUE #730: Consider GroupedFlux#key non nullable

rajinisivaram:


reactor-core/src/main/java/reactor/core/publisher/FluxWindowPredicate.java
```diff
@@ -847,7 +837,7 @@ public Object scanUnsafe(Attr key) {
 
 		@Override
 		public String toString() {
```
smaldini: you can safely remove that toString

rajinisivaram: @smaldini Done, thanks.

PR #932, opened 2017-11-05
Expose LambdaSubscriber class

cbornet:
LambdaSubscriber is a very useful class for consumers of the reactor library. This PR exposes it so we don't have to duplicate the code.

### DIFF
--- a/reactor-core/src/main/java/reactor/core/publisher/LambdaSubscriber.java
+++ b/reactor-core/src/main/java/reactor/core/publisher/LambdaSubscriber.java
@@ -32,7 +32,7 @@ import reactor.util.context.Context;
  *
  * @param <T> the value type
  */
-final class LambdaSubscriber<T>
+public final class LambdaSubscriber<T>
 		implements InnerConsumer<T>, Disposable {
 
 	final Consumer<? super T>            consumer;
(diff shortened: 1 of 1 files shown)


akarnokd:
`LambdaSubscriber` drives `subscribe(...)`. Why do you need it outside this method call?

cbornet:
I am using it [here](https://github.com/cbornet/grpc-java-contrib/blob/reactor/reactor-grpc/reactor-grpc-stub/src/main/java/com/salesforce/reactorgrpc/stub/ServerCalls.java#L134) in my Reactor plugin for gRPC.
The plugin is a direct port to Reactor of the [RxJava one](https://github.com/salesforce/grpc-java-contrib/tree/master/rxgrpc) which is using `io.reactivex.internal.subscribers.LambdaSubscriber`.
Maybe there's a better way to do it  ?

akarnokd:
`rxResponse` is a `Flux` which has the [3 lambda overload](https://projectreactor.io/docs/core/release/api/reactor/core/publisher/Flux.html#subscribe-java.util.function.Consumer-java.util.function.Consumer-java.lang.Runnable-).

cbornet:
Obviously :sweat_smile:  ! I'll see with the author of the RxJava implementation if it was intentional.

PR #950, opened 2017-11-15
Fix what was in ctx on the second step of the test

naxhh:
Hi. I'm just starting with Reactor. And checking how context passing works I just noticed this test was wrong on what the context had in that step.

If I'm wrong on this please let me know how it really works.


pivotal-issuemaster:
@naxhh Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/reactor?repositoryId=reactor/reactor-core&pullRequestId=950)!

PR #955, opened 2017-11-20
#629 Rework to use BiConsumer, rename Flux methods and add errorStrat…

dfeist:
…egyStop.


dfeist:
@simonbasle Let me know what you think...

PR #1163, opened 2018-04-16
javadoc fixes

OlegDokuka:
* better-explained invocation of BiFunction in case of 0_1 elements
* better-explained behavior for maxConcurrency and prefetch params


OlegDokuka:
cc\ @smaldini @simonbasle

PR #1178, opened 2018-04-30
Unexpected behavior for Flux.interval when scheduled interval does not supported

OlegDokuka:
Got an issue with Flux.interval when granularity of interval is nanoseconds, thus that time is rounded to zero milliseconds, hence default `java.util.concurrent.ScheduledExecutorService#scheduleAtFixedRate` does not support zero time according to documentation 

>  * @throws IllegalArgumentException if period less than or equal to zero
     */
    public ScheduledFuture<?> scheduleAtFixedRate(Runnable command,
                                                  long initialDelay,
                                                  long period,
                                                  TimeUnit unit);

Due to the current implementation, we handle only `RejectedTaskException`, which means that error will be propagated to `FluxInterval` which also catches only `RejectedTaskException`
```
@Override
	public void subscribe(CoreSubscriber<? super Long> actual) {
		Worker w = timedScheduler.createWorker();

		IntervalRunnable r = new IntervalRunnable(actual, w);

		actual.onSubscribe(r);

		try {
			w.schedulePeriodically(r, initialDelay, period, unit);
		}
		catch (RejectedExecutionException ree) {
			if (!r.cancelled) {
				actual.onError(Operators.onRejectedExecution(ree, r, null, null,
						actual.currentContext()));
			}
		}
	}
```

Hence `IllegalArgumentException` causes violation of RS.

### DIFF
--- a/reactor-core/src/test/java/reactor/core/publisher/FluxIntervalTest.java
+++ b/reactor-core/src/test/java/reactor/core/publisher/FluxIntervalTest.java
@@ -179,4 +179,15 @@ public class FluxIntervalTest {
 		            .expectNextCount(6)
 		            .verifyErrorMessage("Could not emit tick 32 due to lack of requests (interval doesn't support small downstream requests that replenish slower than the ticks)");
     }
+
+    @Test
+	public void shouldBeAbleToScheduleIntervalsWithLowGranularity() {
+		StepVerifier.create(Flux.interval(Duration.ofNanos(1)))
+		            .expectSubscription()
+		            .expectNext(0L)
+		            .expectNext(1L)
+		            .expectNext(2L)
+		            .thenCancel()
+		            .verify();
+    }
 }
--- a/reactor-core/src/test/java/reactor/core/scheduler/InstantPeriodicWorkerTaskTest.java
+++ b/reactor-core/src/test/java/reactor/core/scheduler/InstantPeriodicWorkerTaskTest.java
@@ -1,0 +1,247 @@
+/*
+ * Copyright (c) 2011-2018 Pivotal Software Inc, All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package reactor.core.scheduler;
+
+import java.util.List;
+import java.util.concurrent.CopyOnWriteArrayList;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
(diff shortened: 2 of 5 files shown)


OlegDokuka:
WIP, I will port impl from RxJava

simonbasle:
Good catch!
Could catch `NullPointerException || IllegalArgumentException`, or even `Throwable`, but yeah that's probably overkill...
It's minor, but can you change the code style though? (indentation and line returns are not consistent with the rest of the source)

OlegDokuka:
@simonbasle ping, ready for review

OlegDokuka:
@smaldini could you please clarify `code block ends are isolated even for try-catch`, because it seems that I do not get it ^-^. Thanks

reactor-core/src/test/java/reactor/core/scheduler/InstantPeriodicWorkerTaskTest.java
```diff
@@ -0,0 +1,300 @@
...
+    }
+
+    private static List<Throwable> prepareErrorHook() {
+        List<Throwable> errorsCollector = new CopyOnWriteArrayList<>();
+        Schedulers.onHandleError((t, e) -> errorsCollector.add(e));
+        return errorsCollector;
+    }
+
+
+    public static void race(final Runnable r1, final Runnable r2, Scheduler s) {
```
simonbasle: Can't this be replaced by `RaceTestUtils`?

reactor-core/src/test/java/reactor/core/publisher/FluxIntervalTest.java
```diff
@@ -179,4 +179,15 @@ public void tickOverflow() {
 		            .expectNextCount(6)
 		            .verifyErrorMessage("Could not emit tick 32 due to lack of requests (interval doesn't support small downstream requests that replenish slower than the ticks)");
     }
+
+    @Test
+	public void shouldSendErrorThroughOnErrorOnSchedulerRejection() {
```
simonbasle: Doesn't seem to test what the name suggests. Change name?

reactor-core/src/test/java/reactor/core/scheduler/InstantPeriodicWorkerTaskTest.java
OlegDokuka: done

smaldini:
This test might not be reliable testWorkerScheduleSupportZeroPeriod. I highly advise to use StepVerifier.withVirtualTime or manual VirtualTimeScheduler.create + advanceTime to make sure these are deterministic tests and avoid random CI failures 👍

OlegDokuka:
Yeah, working on it. There is a chance that the `dispose` might not immediately dispose an execution, so I decided to wait a little bit and that compare tasks size

OlegDokuka:
Seems to be fixed right now @smaldini

smaldini:
Catch style is still not aligned with the rest otherwise looks kool @OlegDokuka

OlegDokuka:
@smaldini could you please comment on lines which require additional cleanup and where proper style should be applied. Since my IDEA configured with the usage of Reactor code-style file I'm not sure where style is improper

OlegDokuka:
ohhhh, now I got it

OlegDokuka:
isolated catch block ... 
image

OlegDokuka:
done

simonbasle:
Thanks for the PR 🎁, will merge

PR #1367, opened 2018-09-29
implementation for TransmitProcessor

OlegDokuka:
# Motivation

 * low overhead kind of UnicastProcessor with full signal transmission from both sides and full backpressure support


OlegDokuka:
@smaldini aligned with DirectProcessor impl

smaldini:
Nice, I don't think we need a separate inner. Also I'm not sure we are ready to accept a new Processor during 3.2 at all since we plan to deprecate or phase out FluxProcessor API in the next 3.3. That doesn't mean the concept is not interesting (it's a kind of UnicastDirectProcessor) but we have to let it live somewhere else meanwhile.

PR #1369, opened 2018-10-01
fix #1361 Snippet from the docs can't be compiled

### LINKED ISSUE #1361: Some examples from the reference can't be compiled
### Expected behavior
Examples from the reference are compiled and working.

### Actual behavior
Examples can't be compiled

### Steps to reproduce
http://projectreactor.io/docs/core/snapshot/reference/#__code_publishon_code
Examples in this paragraph:

```
Scheduler s = Schedulers.newParallel(4); // no such method in 3.2.0

final Flux<Integer> flux = Flux
    .range(1, 2)
    .map(i -> 10 + i)  
    .publishOn(s)  
    .map(i -> "value " + i);  // map to string but Flux<Integer>

new Thread(() -> flux.subscribe(System.out::println));
```

The same story with http://projectreactor.io/docs/core/snapshot/reference/#__code_subscribeon_code

### Reactor Core version
3.2.0 RELEASE

romangr:
Fix generic type and correct scheduler creation call in the docs.

### DIFF
--- a/src/docs/asciidoc/coreFeatures.adoc
+++ b/src/docs/asciidoc/coreFeatures.adoc
@@ -243,9 +243,9 @@ chained in):
 
 [source,java]
 ----
-Scheduler s = Schedulers.newParallel(4); //<1>
+Scheduler s = Schedulers.newParallel("parallel-scheduler", 4); //<1>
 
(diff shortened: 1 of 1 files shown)


smaldini:
nicesome thanks a lot !

PR #1423, opened 2018-11-05
completed SwitchOnFirst impl

OlegDokuka:
porting of the `SwitchOnFirst` operator's implementation with no fusion for now. This is a complete, tested, merge-ready implementation. Fuseable part will be submitted separately since it is a little bit tricky


reactor-core/src/main/java/reactor/core/publisher/FluxSwitchOnFirst.java
```diff
@@ -0,0 +1,637 @@
+/*
+ * Copyright (c) 2011-2017 Pivotal Software Inc, All Rights Reserved.
```
bsideup:
```suggestion
 * Copyright (c) 2011-2018 Pivotal Software Inc, All Rights Reserved.
```

reactor-core/src/main/java/reactor/core/publisher/Operators.java
```diff
@@ -1099,6 +1099,24 @@ public static boolean validate(long n) {
 		return _actual;
 	}
 
+	@SuppressWarnings("unchecked")
+	public static <T> Fuseable.ConditionalSubscriber<? super  T> toConditionalSubscriber(CoreSubscriber<? super T> actual) {
+		Objects.requireNonNull(actual, "actual");
+
+		Fuseable.ConditionalSubscriber<? super T> _actual;
+
+		if (actual instanceof Fuseable.ConditionalSubscriber){
```
bsideup:
```suggestion
		if (actual instanceof Fuseable.ConditionalSubscriber) {
```

simonbasle:
@OlegDokuka we'll hopefully have time to review in depth for 3.2.4 release, won't be able to in time for thursday's 3.2.3, but we haven't forgotten about this PR 😉

OlegDokuka:
> @OlegDokuka we'll hopefully have time to review in depth for 3.2.4 release, won't be able to in time for thursday's 3.2.3, but we haven't forgotten about this PR 😉

Hope this time it will not be like with AsyncFileFlux ;)

bsideup:
Since it took a bit of time to understand this operator, here is a link to the original PR where this operator was added to rsocket-java:
https://github.com/rsocket/rsocket-java/pull/435#issue-148199400

I find the description quite informative:
> Added SwitchTransform to internal, useful when you need to look at the first message on a channel before deciding how to handle it.

reactor-core/src/test/java/reactor/core/publisher/FluxSwitchOnFirstTest.java
```diff
@@ -0,0 +1,677 @@
...
+                    .expectComplete()
+                    .verify(Duration.ofSeconds(10));
+
+        publisher.assertWasRequested();
+        publisher.assertNoRequestOverflow();
+    }
+
+    @Test
+    public void backpressureTest() {
+        TestPublisher<Long> publisher = TestPublisher.createCold();
```
simonbasle: it should be documented in the API that the source is expected to be cold. Also, in order to make it evident that the source isn't subscribed to twice, an explicit test like below would make sense:

```java
    @Test
    public void sourceSubscribedOnce() {
        AtomicInteger subCount = new AtomicInteger();
        Flux<Integer> source = Flux.range(1, 10)
                                   .hide()
                                   .doOnSubscribe(subscription -> subCount.incrementAndGet());

        StepVerifier.create(source.switchOnFirst((s, f) -> f.filter(v -> v % 2 == s.get())))
                    .expectNext(1, 3, 5, 7, 9)
                    .verifyComplete();
        
        assertThat(subCount).hasValue(1);
    }
```

reactor-core/src/test/java/reactor/core/publisher/FluxSwitchOnFirstTest.java
```diff
@@ -0,0 +1,677 @@
...
+        StepVerifier.create(switchTransformed, 0)
+                    .thenRequest(1)
+                    .expectComplete()
+                    .verify(Duration.ofSeconds(10));
+
+        Assert.assertEquals(10001L, requested.get());
+    }
+
+    @Test
+    public void backpressureDrawbackOnConditionalInTransformTest() {
```
simonbasle: can you comment on the title of this test? not sure what it demonstrates...

reactor-core/src/test/java/reactor/core/publisher/FluxSwitchOnFirstTest.java
```diff
@@ -0,0 +1,677 @@
...
+        Flux<String> switchTransformed = publisher.flux()
+                                                  .switchOnFirst((first, innerFlux) -> innerFlux.map(String::valueOf));
+
+        publisher.next(1L);
+
+        StepVerifier.create(switchTransformed, 0)
+                    .thenRequest(1)
+                    .expectNext("1")
+                    .then(() -> publisher.next(2L))
+                    .expectError()
```
simonbasle: maybe assert the error more deeply (class, message)?

reactor-core/src/test/java/reactor/core/publisher/FluxSwitchOnFirstTest.java
```diff
@@ -0,0 +1,677 @@
...
+                                                     (first, innerFlux) -> innerFlux.map(
+                                                             String::valueOf)));
+
+        StepVerifier.create(switchTransformed)
+                    .expectErrorMessage("hello")
+                    .verify(Duration.ofSeconds(10));
+    }
+
+    @Test
+    public void shouldBeAbleToBeCancelledProperly() {
```
simonbasle: one good alternative version of testing this would be by using eg. `take(1)`. maybe add that as a second test `shouldBeableToBeCancelledProperly2`?

reactor-core/src/main/java/reactor/core/publisher/FluxSwitchOnFirst.java
```diff
@@ -0,0 +1,646 @@
...
+                        first = null;
+                        Operators.onDiscard(f, currentContext());
+                    }
+                }
+            }
+        }
+
+        @Override
+        public void subscribe(CoreSubscriber<? super T> actual) {
+            if (once == 0 && ONCE.compareAndSet(this, 0, 1)) {
```
simonbasle: `ONCE` is more of a `STATE`. maybe have the states as constant would help readability here (STATE_INIT = 0, STATE_SUBSCRIBED_FOR_FIRST = 1, STATE_SUBSCRIBED_FOR_TRANSFORMED = 2 or something like that)

reactor-core/src/main/java/reactor/core/publisher/FluxSwitchOnFirst.java
```diff
@@ -0,0 +1,646 @@
...
+
+        @Override
+        public void request(long n) {
+            if (Operators.validate(n)) {
+                if (once == 1 && ONCE.compareAndSet(this, 1, 2)) {
+                    if (first != null) {
+                        drainRegular();
+                    }
+
+                    if (n != Long.MAX_VALUE && --n > 0) {
```
simonbasle: this part seems to never be covered in tests (from IntelliJ coverage stats). Same in the conditional version.

reactor-core/src/main/java/reactor/core/publisher/FluxSwitchOnFirst.java
```diff
@@ -0,0 +1,646 @@
...
+        public void request(long n) {
+            if (Operators.validate(n)) {
+                if (once == 1 && ONCE.compareAndSet(this, 1, 2)) {
+                    boolean sent = false;
+
+                    if (first != null) {
+                        sent = drainRegular();
+                    }
+
+                    if (sent && n != Long.MAX_VALUE && --n > 0) {
```
simonbasle: this part seems to never be covered in tests (from IntelliJ coverage stats). Same in the non-conditional version.

reactor-core/src/main/java/reactor/core/publisher/FluxSwitchOnFirst.java

reactor-core/src/test/java/reactor/core/publisher/FluxSwitchOnFirstTest.java
OlegDokuka: what do you mean by `cold`? The source could be `hot` as well.

Will add such a test. In case you point to the fact that switch transform does not cause double subscription, that it would make sence

reactor-core/src/test/java/reactor/core/publisher/FluxSwitchOnFirstTest.java
```diff
@@ -0,0 +1,677 @@
...
+    @Test
+    public void shouldSendOnNextSignalConditional() {
+        Signal<? extends Long>[] first = new Signal[1];
+
+        StepVerifier.create(Flux.just(1L)
+                                .switchOnFirst((s, f) -> {
+                                    first[0] = s;
+
+                                    return f;
+                                }))
```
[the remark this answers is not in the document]
OlegDokuka: yeah...

reactor-core/src/test/java/reactor/core/publisher/FluxSwitchOnFirstTest.java
simonbasle: yeah, plus a bunch of these tests use `TestPublisher.createCold()` and don't work with the `TestPublisher.create()` (which can be viewed as hot, it forwards the same data to all subscribers)

simonbasle:
please add more commit as you fix the review elements @OlegDokuka, we'll rebase/squash in the final step (or I'll squash when merging)

reactor-core/src/test/java/reactor/core/publisher/FluxSwitchOnFirstTest.java
OlegDokuka: done

reactor-core/src/main/java/reactor/core/publisher/FluxSwitchOnFirst.java
OlegDokuka: nice catchup! found a 🐛 there! 😄

reactor-core/src/main/java/reactor/core/publisher/FluxSwitchOnFirst.java
OlegDokuka: nice catchup! found a 🐛 there! 😄

reactor-core/src/test/java/reactor/core/publisher/FluxSwitchOnFirstTest.java
OlegDokuka: done

reactor-core/src/test/java/reactor/core/publisher/FluxSwitchOnFirstTest.java
```diff
@@ -580,23 +766,23 @@ public void shouldBeAbleToCancelSubscription() throws InterruptedException {
...
+                Assertions.assertThat(latch.await(5, TimeUnit.SECONDS)).isTrue();
+                Assertions.assertThat(requested.get()).isEqualTo(1L);
                 capturedElements.add(captureElement.get());
                 capturedCompletions.add(captureCompletion.get());
             }
 
-            Assume.assumeThat(capturedElements, hasItem(equalTo(0L)));
-            Assume.assumeThat(capturedCompletions, hasItem(equalTo(false)));
+
+            Assumptions.assumeThat(capturedElements).contains(0L);
```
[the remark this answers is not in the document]
OlegDokuka: No, I don't. In this case, it could happen that racing can end up with all completion instead of cancellation. So I don't want to fail the test because for some system reasons element propagation was faster than the Thread allocation and cancellation task execution

OlegDokuka:
@simonbasle WIP. I'm traveling right now, so trying to push commits in transition 😄

reactor-core/src/test/java/reactor/core/publisher/FluxSwitchOnFirstTest.java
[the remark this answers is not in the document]
OlegDokuka: it is kind of unstable test, so I decided to use assume there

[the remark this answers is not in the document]
simonbasle: please document that very explicitly with a comment then!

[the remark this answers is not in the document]
OlegDokuka: sure buddy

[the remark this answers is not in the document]

reactor-core/src/main/java/reactor/core/publisher/Operators.java
```diff
@@ -1099,6 +1099,17 @@ public static boolean validate(long n) {
 		return _actual;
 	}
 
+	/**
+	 * If the actual {@link CoreSubscriber} is not {@link Fuseable.ConditionalSubscriber},
+	 * it will apply a {@link ConditionalSubscriberAdapter} which directly maps all
```
simonbasle: the class linked is not publicly visible. just say "it will apply an adapter"

reactor-core/src/main/java/reactor/core/publisher/Operators.java
```diff
@@ -1099,6 +1099,17 @@ public static boolean validate(long n) {
...
+	 * {@link Fuseable.ConditionalSubscriber#tryOnNext(T)} to {@link CoreSubscriber#onNext(T)}
+	 * and always returns true as the result
+	 *
+	 * @param <T> passed subscriber type
+	 *
+	 * @param actual the {@link Subscriber} to apply hook on
```
simonbasle: "to apply hook on" -> "to adapt"

reactor-core/src/main/java/reactor/core/publisher/Operators.java
```diff
@@ -1099,6 +1099,17 @@ public static boolean validate(long n) {
...
+	 * @return an eventually transformed {@link Subscriber}
```
simonbasle:
```suggestion
	 * @return a potentially adapted {@link Fuseable.ConditionalSubscriber}
```

reactor-core/src/main/java/reactor/core/publisher/Flux.java
```diff
@@ -7899,6 +7899,51 @@ public final void subscribe(Subscriber<? super T> actual) {
 		return subscriber;
 	}
 
+	/**
+	 * Provide current {@link Flux<T>} transformation when the first available signal
```
simonbasle: replace the next two entire paragraphs with the following suggestion

```
	 * Transform the current {@link Flux<T>} once it emits its first element, making a
	 * conditional transformation possible. This operator first requests one element
	 * from the source then applies a transformation derived from a {@link Signal}.
	 * Note that the source might complete or error immediately instead of emitting,
	 * in which case the {@link Signal} would be onComplete or onError. It is NOT
	 * necessarily an onNext Signal, and must be checked accordingly.
```

reactor-core/src/main/java/reactor/core/publisher/Flux.java
```diff
@@ -7899,6 +7899,51 @@ public final void subscribe(Subscriber<? super T> actual) {
...
+     * first operator proactively requests the first element from the upstream. It does
+     * not mean that upstream is always mandated to return an element, but free to be
+     * empty or error stream, so the given in transform bi-function {@link Signal} can be
+     * either onNext, onError or onComplete.
+     *
+     * Note that any given signal will be propagated to the downstream regardless the
+     * type of the signal. Typically, this operator could be used to provide stream
+     * transformation depends on the first signal in the stream.
+     *
+     * For example, it could be used to define the stream routing, depends on the first
```
simonbasle: Replace the two line paragraph with the following:

```
     * For example, this operator could be used to define a form of routing, depending on the first
     * element (which could contain routing metadata for instance):
```

reactor-core/src/main/java/reactor/core/publisher/Flux.java
```diff
@@ -7899,6 +7899,51 @@ public final void subscribe(Subscriber<? super T> actual) {
...
+	 *          }
+	 *      }
+	 *
+	 *      return Flux.empty();
+	 *  })
+	 * }
+	 * </pre></blockquote>
+	 *
+	 * <p>
+	 * <img class="marble" src="https://raw.githubusercontent.com/reactor/reactor-core/master/docs/marble/switchonfirst.png" alt="">
```
simonbasle: don't include a marble yet, please add your operator in a comment to #1463 instead.

reactor-core/src/main/java/reactor/core/publisher/Flux.java
```diff
@@ -7899,6 +7899,51 @@ public final void subscribe(Subscriber<? super T> actual) {
...
+	 * <p>
+	 *
+	 * @param transformer function transformer. Is executed once the first signal is
```
simonbasle:
```suggestion
	 * @param transformer A {@link BiFunction} executed once the first signal is available and used to transform it conditionally.
```

reactor-core/src/main/java/reactor/core/publisher/Flux.java
```diff
@@ -7899,6 +7899,51 @@ public final void subscribe(Subscriber<? super T> actual) {
...
+     * available. Must return a flux instance as a return type
```
simonbasle:
```suggestion
```

reactor-core/src/main/java/reactor/core/publisher/Operators.java
simonbasle:
```suggestion
	 * it will apply an adapter which directly maps all
```

reactor-core/src/main/java/reactor/core/publisher/Flux.java
OlegDokuka: done

reactor-core/src/main/java/reactor/core/publisher/Flux.java
OlegDokuka: done

reactor-core/src/main/java/reactor/core/publisher/Flux.java
OlegDokuka: done

reactor-core/src/main/java/reactor/core/publisher/Flux.java
OlegDokuka: done

PR #1428, opened 2018-11-09
Fix flaky `distinctUntilChanged*DoesntRetainObjects` tests

bsideup:
This PR introduces Awaitility and changes `distinctUntilChanged*DoesntRetainObjects` to make a few assertion attempts instead of just one after `Thread.sleep`


bsideup:
@simonbasle unless you specify (with `await().atMost(n, TimeUnit.SECONDS)`) it fallbacks to the default value of 10 seconds

PR #1434, opened 2018-11-13
make `expectAccessibleContext` work with `initialContext` from `StepVerifier`

bsideup:
See #1432 #1417


bsideup:
Replaced with #1435 (to target 3.1.x)

PR #1460, opened 2018-12-08
feat(kotlin): add switchIfEmpty with deferred execution

davinkevin:
This allows supporting a deferred execution when using the `switchIfEmpty` operator


davinkevin:
Modification made. I've set version `3.2` on the `@since`.

I've also modified a signature method to accept `Publisher` instead of `Mono` in the `FluxExtensions.kt`.

👍

PR #1481, opened 2019-01-04
Basic tracking support

bsideup:


reactor-core/src/main/java/reactor/core/publisher/ParallelFlux.java
```diff
@@ -1268,4 +1282,37 @@ else if (s2 != null) {
 		return both;
 	}
 
+	private static class MultiSubscriber<T> implements CoreSubscriber<T> {
```
bsideup: This is hacky but didn't know how to do it better

robotmrv:
@bsideup, 
this is very interesting feature. 
Is it still in progress but hidden or rejected forever?

bsideup:
HI @robotmrv,

We decided to split it into smaller chunks (see #1527 and #1546) for now. These PRs duplicate what we did in the tracing one, with an exception of the ContextAware Runnables (see #1524, also closed since most of tracing solutions rely on thread locals anyways), and a common tracking interface (no longer needed given the provided abstractions)

We may consider integrating Reactor with some tracing solutions in future, but that will be another PR :)

PR #1677, opened 2019-04-26
Do not mark `onEachOperatorHook` field as volatile.

bsideup:
Since every operator at assembly time accesses these fields,
we should relax the guarantees and avoid a volatile access.

/cc @akarnokd

### DIFF
--- a/reactor-core/src/main/java/reactor/core/publisher/Hooks.java
+++ b/reactor-core/src/main/java/reactor/core/publisher/Hooks.java
@@ -497,8 +497,8 @@ public abstract class Hooks {
 	}
 
 	//Hooks that are transformative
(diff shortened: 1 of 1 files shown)


akarnokd:
They are volatile for safe publication of the handler function. Without it, a half-constructed callback could wreak havoc.

smaldini:
Maybe we could start with onEachOperator only, the visibility of these callbacks are far less important than the onLastOperator and we might want to deprecate them altogether before the next generation of Reactor since we are building bytecode manipulations alternative for the main use case, debugging.
Also synchronized is used on the log instance which should force a memory barrier read/store and protect the atomicity of the callback updates.

bsideup:
@smaldini changed

PR #1746, opened 2019-06-12
Add "VirtualTimeScheduler#getScheduledTaskCount()"

ttddyy:
Expose "counter" as scheduled task count in VirtualTimeScheduler.

While testing publisher logic that have `publishOn/subscribeOn`, I'd like to have a way to check provided scheduler has performed the execution.
Since `VirtualTimeScheduler` already have `counter` that keeps track of the number of the scheduled tasks, it can be used to check whether tasks have executed on the scheduler.

For example,

```java
mono....
  .publishOn(vts)
  .subscribe();
  
assertThat(vts.getScheduledTaskCount()).isGreaterThanOrEqualTo(1);
```

### DIFF
--- a/reactor-test/src/test/java/reactor/test/scheduler/VirtualTimeSchedulerTests.java
+++ b/reactor-test/src/test/java/reactor/test/scheduler/VirtualTimeSchedulerTests.java
@@ -250,6 +250,71 @@ public class VirtualTimeSchedulerTests {
 		}
 	}
 
+	@Test
+	public void scheduledTaskCount() {
+		VirtualTimeScheduler vts = VirtualTimeScheduler.create();
+		assertThat(vts.getScheduledTaskCount()).as("initial value").isEqualTo(0);
+
+		vts.schedule(() -> {
+		});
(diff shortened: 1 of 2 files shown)


reactor-test/src/test/java/reactor/test/scheduler/VirtualTimeSchedulerTests.java
```diff
@@ -250,6 +250,61 @@ public void racingAdvanceTimeOnVaryingQueue() {
...
+		};
+
+		vts = VirtualTimeScheduler.create();
+		assertThat(vts.getScheduledTaskCount()).as("initial value").isEqualTo(0);
+
+		vts.schedule(task);
+		assertThat(vts.getScheduledTaskCount()).as("a task scheduled").isEqualTo(1);
+
+		// schedule with delay
+		vts = VirtualTimeScheduler.create();
```
bsideup: I would suggest splitting it into a few different tests

ttddyy:
updated the PR.
Split the test to multiple test methods as suggested.

bsideup:
@ttddyy thanks for your contribution! 🎉

PR #1758, opened 2019-06-21
Deprecate `compose` methods

bsideup:
As per #1745, this PR deprecates `compose` method. 

However, the suggested alternative (`transform` + `defer`) has a downside - `onAssembly` will be triggered two times (one in `transform` and another in `defer`).
We could advice using `as` + `defer`, but it may not be an easy-to-find alternative. Perhaps renaming the original method to match `transform` is a better option (instead of dropping the method in favour of `as`+`defer`)

### DIFF
--- a/reactor-core/src/main/java/reactor/core/publisher/Mono.java
+++ b/reactor-core/src/main/java/reactor/core/publisher/Mono.java
@@ -1759,7 +1759,9 @@ public abstract class Mono<T> implements CorePublisher<T> {
 	 * @return a new {@link Mono}
 	 * @see #as as() for a loose conversion to an arbitrary type
 	 * @see #transform(Function)
+	 * @deprecated use {@link #as(Function)} with {@link #defer(Supplier)}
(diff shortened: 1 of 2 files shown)


reactor-core/src/main/java/reactor/core/publisher/Mono.java
```diff
@@ -1759,7 +1759,9 @@ public T block(Duration timeout) {
 	 * @return a new {@link Mono}
 	 * @see #as as() for a loose conversion to an arbitrary type
 	 * @see #transform(Function)
+	 * @deprecated use {@link #as(Function)} with {@link #defer(Supplier)}
```
smaldini: should use transform, as is a full type conversion

bsideup: see my comment about `#transform` in the PR's description - not that trivial

bsideup:
Superseeded by #1782

PR #1869, opened 2019-08-30
Macro-fusion: FluxMap

bsideup:


reactor-core/src/main/java/reactor/core/publisher/FluxMap.java
```diff
@@ -45,7 +63,7 @@
 	 *
 	 * @throws NullPointerException if either {@code source} or {@code mapper} is null.
 	 */
-	FluxMap(Flux<? extends T> source,
+	private FluxMap(Flux<? extends T> source,
```
bsideup: I remember that we prefer internal visibility, but this one is helpful because it forces us to create it through the static factory, and, unless needed, would be better to keep the constructor private

bsideup:
Synthetic benchmark (subsequent `.map` calls) gives pretty impressive results (as expected):
```
$ bat build/reports/jmh*/result.text
───────┬────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
       │ File: build/reports/jmh/result.text
───────┼────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
   1   │ Benchmark               (operatorsCount)  Mode  Cnt    Score   Error  Units
   2   │ TailCallBenchmark.with                 5  avgt    5  126,881 ± 3,206  ns/op
   3   │ TailCallBenchmark.with                10  avgt    5  127,072 ± 2,828  ns/op
───────┴────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
───────┬────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
       │ File: build/reports/jmhBaseline/result.text
───────┼────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
   1   │ Benchmark               (operatorsCount)  Mode  Cnt    Score    Error  Units
   2   │ TailCallBenchmark.with                 5  avgt    5  269,674 ±  6,051  ns/op
   3   │ TailCallBenchmark.with                10  avgt    5  428,167 ± 11,110  ns/op
```

PR #1902, opened 2019-09-23
#1888 detect Context loss

bsideup:
See #1888.

Since loosing the context may lead to some major issues (lost transaction, broken tracing, web security, etc etc), we may want to add an opt-in flag to turn on the "context loss detection".

The implementation adds a random key to the context and replaces the source of `transform` and `transformDeferred` with one that checks whether the key is still present.


reactor-core/src/main/java/reactor/core/publisher/Flux.java
```diff
@@ -8934,6 +8937,17 @@ public final void subscribe(Subscriber<? super T> actual) {
 	 * @see #as for a loose conversion to an arbitrary type
 	 */
 	public final <V> Flux<V> transform(Function<? super Flux<T>, ? extends Publisher<V>> transformer) {
+		if (Hooks.DETECT_CONTEXT_LOSS) {
+			String key = CONTEXT_MARKER_PREFIX + UUID.randomUUID().toString();
```
simonbasle: use identity hashcode rather than UUID

reactor-core/src/main/java/reactor/core/publisher/Flux.java
```diff
@@ -8955,6 +8969,17 @@ public final void subscribe(Subscriber<? super T> actual) {
 	 * @see #as as() for a loose conversion to an arbitrary type
 	 */
 	public final <V> Flux<V> transformDeferred(Function<? super Flux<T>, ? extends Publisher<V>> transformer) {
+		if (Hooks.DETECT_CONTEXT_LOSS) {
+			String key = CONTEXT_MARKER_PREFIX + UUID.randomUUID().toString();
```
simonbasle: use identity hashcode rather than UUID

reactor-core/src/main/java/reactor/core/publisher/Flux.java
```diff
@@ -8955,6 +8969,17 @@ public final void subscribe(Subscriber<? super T> actual) {
...
+			Flux<T> source = this.subscriberContext(ctx -> {
```
simonbasle: ideally there wouldn't be two additional `Subscriber` involved here (`subscriberContext`) => merge the `defer` operator and `subscriberContext` subscribers into a dedicated implementation for this particular case?

reactor-core/src/main/java/reactor/core/publisher/Hooks.java
```diff
@@ -466,6 +464,14 @@ public static void resetOnNextError() {
 		}
 	}
 
+	public static void enableContextLossTracking() {
```
simonbasle: (NB: don't merge without some javadoc ;) )

reactor-core/src/main/java/reactor/core/publisher/Flux.java
```diff
@@ -8934,6 +8937,17 @@ public final void subscribe(Subscriber<? super T> actual) {
 	 * @see #as for a loose conversion to an arbitrary type
 	 */
 	public final <V> Flux<V> transform(Function<? super Flux<T>, ? extends Publisher<V>> transformer) {
```
simonbasle: do we need this on `transform`? keeping things on `transformDeferred` would align well with #1890 and keep the message simple and consistent: if you're using an RS library other than Reactor, scope it into `transformDeferred`.

reactor-core/src/main/java/reactor/core/publisher/Hooks.java
bsideup: added

reactor-core/src/main/java/reactor/core/publisher/ContextTrackingFunctionWrapper.java
```diff
@@ -0,0 +1,78 @@
...
+
+			if (!ctx.hasKey(key)) {
+				throw new IllegalStateException("Context loss after applying " + transformer);
+			}
+
+			Context newContext = ctx.delete(key);
+			return new ContextStartSubscriber<>(actual, newContext);
+		}).andThen(transformer).apply(self);
+
+		return new CorePublisher<V>() {
```
simonbasle: for `newSource`, type is maintained by `liftPublisher`. But here you are returning a `Publisher` where there was potentially a `Flux` or `Mono`. Isn't that problematic?

bsideup: in `transform`, there is `from` on the function's returned value anyways, so that it looks like this:
```java
Mono<V> transform(fn) {
    fn = new Wrapper(fn);
    return onAssembly(from(fn.apply(this)));
}
```

simonbasle: ah, got it. please add a comment for our future selves 😛

reactor-tools/src/main/java/reactor/tools/agent/ContextLossDetectionTransformer.java
```diff
@@ -0,0 +1,100 @@
...
+				int access,
+				String name,
+				String descriptor,
+				String signature,
+				String[] exceptions
+		) {
+			MethodVisitor visitor = super.visitMethod(access, name, descriptor, signature, exceptions);
+
+			switch (name) {
+				case "transform":
```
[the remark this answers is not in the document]
bsideup: it is kinda the same as having `onEachOperator` since we call `onAssembly(from(...))` in `transform` anyways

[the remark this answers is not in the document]
simonbasle: ok, I see your point :+1:

reactor-core/src/main/java/reactor/core/publisher/Hooks.java
```diff
@@ -41,6 +41,8 @@
  */
 public abstract class Hooks {
 
+
```
simonbasle: leftover extra newlines?

reactor-core/src/main/java/reactor/core/publisher/Hooks.java
```diff
@@ -464,6 +466,27 @@ public static void resetOnNextError() {
 		}
 	}
 
+	/**
+	 * Globally enables the {@link reactor.util.context.Context} loss detection that may happen
```
simonbasle: `may` suggests that there is more conditions for it to happen. did you mean `will`?
I'd rephrase along the lines of:
> Globally enables the Context loss detection in operators like transform or transformDeferred when non-Reactor types are used.

reactor-core/src/main/java/reactor/core/publisher/Hooks.java
```diff
@@ -464,6 +466,27 @@ public static void resetOnNextError() {
...
+	 * when you use operators like {@link reactor.core.publisher.Mono#transform(Function)}
+	 * or {@link reactor.core.publisher.Flux#transformDeferred(Function)} with non-Reactor types.
+	 *
+	 * An exception will be thrown if after applying the transformations a new {@link reactor.util.context.Context}
```
simonbasle: Improve the flow of the sentence:
> An exception will be thrown upon applying the transformation if the original Context isn't reachable (ie. it has been replaced by a totally different Context, or no Context at all)

reactor-core/src/main/java/reactor/core/publisher/ContextTrackingFunctionWrapper.java
```diff
@@ -0,0 +1,79 @@
...
+import reactor.core.CorePublisher;
+import reactor.core.CoreSubscriber;
+import reactor.core.publisher.FluxContextStart.ContextStartSubscriber;
+import reactor.util.context.Context;
+
+/**
+ * This {@link Function} wrapper is used by reactor-tools to implement the context loss detection.
+ *
+ */
+@SuppressWarnings("unused")
```
simonbasle: this warning suppression should not be necessary anymore

simonbasle:
ping @bsideup (see my review comments above)

bsideup:
@simonbasle thanks for the ping! Updated 👍

PR #2082, opened 2020-03-25
[test] Add a post-test listener that resets hooks consistently

bsideup:
cc @sbrannen

### DIFF
--- a/reactor-core/build.gradle
+++ b/reactor-core/build.gradle
@@ -64,6 +64,7 @@ dependencies {
 
   testCompile "junit:junit:$jUnitVersion"
   testCompile "org.junit.jupiter:junit-jupiter-api:${jUnitJupiterVersion}"
+  testCompile "org.junit.platform:junit-platform-launcher:1.6.1"
   testCompile "org.junit.jupiter:junit-jupiter-params:${jUnitJupiterVersion}"
   testRuntime "org.junit.jupiter:junit-jupiter-engine:${jUnitJupiterVersion}"
   testRuntime "org.junit.vintage:junit-vintage-engine:${jUnitJupiterVersion}"
--- a/reactor-core/src/test/java/reactor/ReactorTestExecutionListener.java
+++ b/reactor-core/src/test/java/reactor/ReactorTestExecutionListener.java
@@ -1,0 +1,50 @@
+/*
+ * Copyright (c) 2020-Present Pivotal Software Inc, All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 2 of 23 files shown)


reactor-core/build.gradle
```diff
@@ -64,6 +64,7 @@ dependencies {
 
   testCompile "junit:junit:$jUnitVersion"
   testCompile "org.junit.jupiter:junit-jupiter-api:${jUnitJupiterVersion}"
+  testCompile "org.junit.platform:junit-platform-launcher:1.6.1"
```
sbrannen: Have you considered importing the [`junit-bom`](https://junit.org/junit5/docs/current/user-guide/#dependency-metadata-junit-bom) to manage all JUnit 5 artifact versions?

reactor-core/src/test/java/reactor/ReactorTestExecutionListener.java
```diff
@@ -0,0 +1,50 @@
...
+		Schedulers.resetOnHandleError();
+		Schedulers.resetFactory();
+		Schedulers.resetOnScheduleHooks();
+
+		// TODO capture non-default schedulers and shutdown them
+	}
+
+	@Override
+	public void executionFinished(TestIdentifier testIdentifier, TestExecutionResult testExecutionResult) {
+		reset();
```
sbrannen: Do you want to invoke `reset()` for all test identifiers?

FYI: as currently implemented, `reset()` would get invoked after every test method and after every test class.

So, if you just want to do this for test methods, you would need to check if the `TestIdentifier` represents a container or a test.

bsideup:
@sbrannen 

> If this is only for internal use within your own build, and you're only using JUnit Jupiter, 

we have a mix of Jupiter and Vintage :) But we will definitely consider you suggestion when we migrate everything to Jupiter 👍

reactor-core/src/test/java/reactor/ReactorTestExecutionListener.java
bsideup: Since `reset` method is very lightweight, there is no harm in calling it for both test method & class.
We had a concern that some classes may install class-level hooks, but it seems that we don't have any.

reactor-core/build.gradle
bsideup: That's a good suggestion. I will check what we have for importing BOMs in reactor-core's build.

reactor-core/src/test/java/reactor/ReactorTestExecutionListener.java
sbrannen: OK. Thanks for the clarification.

reactor-core/build.gradle
sbrannen: OK. I'll let you investigate that separately, and I'll go ahead and mark my suggestions as approved.

PR #2161, opened 2020-05-18
Ensure SwitchOnFirst serializes requests per rule 2.7

OlegDokuka:
FluxSwitchOnFirst must ensure that requests are issued in serial manner following [rule 2.7](https://github.com/reactive-streams/reactive-streams-jvm/blob/v1.0.3/README.md#2.7)


OlegDokuka:
Delayed until #2164

PR #2183, opened 2020-06-08
Fix #2181 Handle nulls from collectors in Flux.collect()

### LINKED ISSUE #2181: `Flux#collect` does not handle `null` result
While (very?) uncommon, a `Collector` may return `null`. Currently `Flux#collect` does not explicitly consider this case, causing `onNext(null)` to be invoked.

## Steps to Reproduce

```java
@Test
public void reproCase() {
    // Passes.
    StepVerifier.create(Mono.just("foo").map(x -> null))
        .verifyError(NullPointerException.class);

    // Fails.
    StepVerifier.create(Flux.empty().collect(reducing(null, (a, b) -> a)))
        .verifyError(NullPointerException.class);
}
```

The second statement yields:
```
java.lang.AssertionError: expectation "expectError(Class)" failed (expected: onError(NullPointerException); actual: onNext(null))
```

## Possible Solution

Consistent with `.map`, I think a `NullPointerException` should be emitted. The alternative is to emit `onComplete()`, but from an API evolvability POV it seems safer to simply disallow this for now. Another argument for going with the NPE is that this maintains the invariant that `Flux#collect` always emits exactly one element.

##  Environment

* Reactor version: `Dysprosium-SR7`
* JVM version: `AdoptOpenJDK (build 11.0.7+10)`

ericbottard:
Emit an NPE early instead of signaling onNext(null).


reactor-core/src/test/java/reactor/core/publisher/FluxCollectTest.java
```diff
@@ -0,0 +1,43 @@
+/*
+ * Copyright (c) 2011-2020 Pivotal Software Inc, All Rights Reserved.
```
bsideup, line 2:
```suggestion
 * Copyright (c) 2011-Present Pivotal Software Inc, All Rights Reserved.
```

reactor-core/src/main/java/reactor/core/publisher/MonoStreamCollector.java
```diff
@@ -164,7 +164,11 @@ public void onComplete() {
 				return;
 			}
 
-			complete(r);
+			if (r != null) {
+				complete(r);
+			} else {
+				actual.onError(Operators.onOperatorError(new NullPointerException("Collector returned null"), actual.currentContext()));
+			}
```
bsideup, line 171: WDYT about an early return?

```suggestion
			if (r == null) {
				actual.onError(Operators.onOperatorError(new NullPointerException("Collector returned null"), actual.currentContext()));
				return;
			}
			
			complete(r);
```

reactor-core/src/test/java/reactor/core/publisher/FluxCollectTest.java
```diff
@@ -0,0 +1,43 @@
...
+import reactor.test.StepVerifier;
+
+import static java.util.stream.Collectors.reducing;
+
+/**
+ * Tests for {@link Flux#collect(Collector)}.
+ *
+ * @author Eric Bottard
+ */
+public class FluxCollectTest {
```
bsideup, line 32: FYI there is `MonoStreamCollectorTest` that tests `collect(Collector)`

reactor-core/src/test/java/reactor/core/publisher/FluxCollectTest.java
violetagg: This should not be `Pivotal` but `VMware`

Issue #2186, opened 2020-06-10
Race condition in `MonoCollect` leads to NullPointerException 


rstoyanchev:
The example in https://github.com/spring-projects/spring-framework/issues/25216 shows a NullPointerException in the logs:

```
java.lang.NullPointerException: null
at reactor.core.publisher.MonoCollect$CollectSubscriber.onNext(MonoCollect.java:124) [reactor-core-3.3.6.RELEASE.jar:3.3.6.RELEASE]
at reactor.core.publisher.FluxMap$MapSubscriber.onNext(FluxMap.java:114) [reactor-core-3.3.6.RELEASE.jar:3.3.6.RELEASE]
at reactor.core.publisher.FluxPeek$PeekSubscriber.onNext(FluxPeek.java:192) [reactor-core-3.3.6.RELEASE.jar:3.3.6.RELEASE]
at reactor.core.publisher.FluxPeek$PeekSubscriber.onNext(FluxPeek.java:192) [reactor-core-3.3.6.RELEASE.jar:3.3.6.RELEASE]
at reactor.core.publisher.FluxMap$MapSubscriber.onNext(FluxMap.java:114) [reactor-core-3.3.6.RELEASE.jar:3.3.6.RELEASE]
at reactor.netty.channel.FluxReceive.onInboundNext(FluxReceive.java:330) [reactor-netty-0.9.8.RELEASE.jar:0.9.8.RELEASE]
...
```

This happens while collecting buffers in [DataBufferUtils#join](https://github.com/spring-projects/spring-framework/blob/master/spring-core/src/main/java/org/springframework/core/io/buffer/DataBufferUtils.java#L551) and what's `null` is the `value` field (the container collecting the items) and not the `action`. In a debugger this is where the NPE is raised:

```
at org.springframework.core.io.buffer.DataBufferUtils$$Lambda$340.551556520.accept(Unknown Source:-1)
at reactor.core.publisher.MonoCollect$CollectSubscriber.onNext(MonoCollect.java:124)
at reactor.core.publisher.FluxMap$MapSubscriber.onNext(FluxMap.java:114)
at reactor.core.publisher.FluxPeek$PeekSubscriber.onNext(FluxPeek.java:192)
at reactor.core.publisher.FluxPeek$PeekSubscriber.onNext(FluxPeek.java:192)
at reactor.core.publisher.FluxMap$MapSubscriber.onNext(FluxMap.java:114)
[...]

PR #2192, opened 2020-06-16
Align MonoCollect with MonoList

rstoyanchev:
This change aligns the implementation of `MonoCollect` to be comparable to that of `MonoList` in order to address #2186. `MonoList` seems better protected against competing upstream and downstream signals and the two should not be different as far as I can see. A `MonoList` is essentially a `MonoCollect` with an `ArrayList` as the container.


rstoyanchev:
Closing (wrong branch).

rstoyanchev:
Closing (wrong branch).

PR #2218, opened 2020-06-23
Suggesting some changes on the Processor API update

smaldini:
Giving a go at my latest comment on #2188


reactor-core/src/main/java/reactor/core/publisher/Sinks.java
```diff
@@ -203,64 +159,43 @@ public void error(Throwable e) {
...
-		public StandaloneFluxSink<T> next(T t) {
+		public Sink<T> next(T t) {
 			delegateSink.next(t);
 			return this;
 		}
 	}
 
-	//TODO improve synchronization, prefer CAS ?
-	static final class MonoProcessorSink<T> implements StandaloneMonoSink<T> {
+	static final class MonoProcessorSink<T> implements Sink<T> {
```
simonbasle, line 168: if sinks are advertised as thread safe objects, doesn't this require external synchronization of the processor's calls ? especially if we later introduce several flavors of `MonoProcessor` ?

docs/asciidoc/processors.adoc
```diff
@@ -45,7 +45,7 @@ If, after exploring the above alternatives, you still think you need a `Processo
 the <<processor-overview>> section to learn about the different implementations.
 
 [[sinks]]
-= Safely Produce from Multiple Threads by Using `StandaloneFluxSink` and `StandaloneMonoSink`
+= Safely Produce from Multiple Threads by Using `Sink` and `Sink`
```
simonbasle, line 48: IF the mono flavor is removed, this `and 'Sink'` can be removed. but I'm not a huge fan of removing the mono flavor...

reactor-core/src/main/java/reactor/core/publisher/DirectProcessor.java
```diff
@@ -81,7 +81,7 @@
  * @deprecated Prefer clear cut usage of either {@link Processors} or {@link Sinks}, to be removed in 3.5
  */
 @Deprecated
-public final class DirectProcessor<T> extends FluxIdentityProcessor<T> {
+public final class DirectProcessor<T> extends FluxProcessor<T, T> {
```
simonbasle, line 84: could you maybe revert the `FluxIdentityProcessor` removal for now, and offer it in a separate PR ? this distracts from the rest of the PR IMHO, as maybe 90% of these files are only changed due to this removal.

reactor-core/src/main/java/reactor/core/publisher/FluxIdentityProcessor.java
```diff
@@ -1,26 +0,0 @@
...
- */
-
-package reactor.core.publisher;
-
-/**
- * A {@link FluxProcessor} that has the same input and output types.
- * 
- * @author Simon Baslé
- */
-public abstract class FluxIdentityProcessor<T> extends FluxProcessor<T, T> {
```
simonbasle, line 24: the rationale for adding this was that `FluxProcessor<String, String>` feels clunky, and was such a pain to migrate when removing deprecation warnings in the test code. Maybe that pain will be a good motivator for people to move to `Sink`? But maybe on the contrary it will just make them angry 😝

reactor-core/src/main/java/reactor/core/publisher/FluxSink.java
```diff
@@ -113,6 +113,16 @@
 	 */
 	FluxSink<T> onDispose(Disposable d);
 
+	@Override
+	default Flux<T> toFlux() {
+		return Flux.error(new IllegalStateException("A FluxSink does not support back referencing the outer Flux"));
```
simonbasle, line 118: honestly, I'm not a huge fan of offering an API and then having it error in half the cases. this feels like a step backward towards the same kind of inconsistencies that we had in processors.

reactor-core/src/main/java/reactor/core/publisher/MonoSink.java
```diff
@@ -31,7 +33,45 @@
  *
  * @param <T> the value type emitted
  */
-public interface MonoSink<T> extends ScalarSink<T> {
+public interface MonoSink<T> extends Sink<T> {
```
simonbasle, line 36: I think this is very problematic. Offering a flux-like API on the `MonoSink`, even if we introduce a specialized API, is bound to be confusing.

ie this API seemingly allows `next(t).next(t).complete()` or even more confusing, `next(t1).next(t2).success(t3)` type of calls.

reactor-core/src/main/java/reactor/core/publisher/Sinks.java
smaldini: MonoProcessor are safe, all the onXxx methods are atomic - I'll actually have an extra commit that moves further and remove the need for MonoProcessorSink by making MonoProcessor implementing it directly.

docs/asciidoc/processors.adoc
smaldini: I think that is something I'm not too attached to personally if you think having a dedicated mono is valuable. I'll clean the doc in the final pass. The main problem I'm attacking is removing Processors which is more impacting IMO.

reactor-core/src/main/java/reactor/core/publisher/DirectProcessor.java
smaldini: The PR is not finished and there will be more impact due to removing Processors and use of Processor directly wherever possible.

reactor-core/src/main/java/reactor/core/publisher/FluxIdentityProcessor.java
smaldini: Yeah the rationale is to not use processors api at all in the future so no need to spend any more surface api for them.

reactor-core/src/main/java/reactor/core/publisher/FluxSink.java
smaldini: It's not impossible to support it but its easy to make mistakes if supported. At least its explicit that it doesn't work unlike using Processor directly vs sinks. We can look at the final result once I'm done and see what can go back. Right now i'm aiming for one API entry: Sink/Sinks and will add more from this.

reactor-core/src/main/java/reactor/core/publisher/MonoSink.java
smaldini: That is changed in the next commit. Sink has orthogonal methods "boolean emitXxx" (complete|error|next). Chaining only brings value to FluxSink if any and it was impossible for processors to be internally sink due to competing error name. Plus it provides for a new feature, returning false if emission failed because the sink is terminated.

smaldini:
I'm close to happiness with the changes: 
- Only 2 new APIs : Sinks and Sink
- Remove the need to use a Processor altogether except for integration purpose
- Optimized Sink impls to avoid extra wrapper allocation

Follow up: refining Sinks/Sink api if needed:
- the alloc-free builder API is a good idea to make `Sinks` more guided, provide toProcessor 
- do we need extra Sink api: scannable(), isTerminated/hasCancelled, hasDowntream/Subscribers...
- polish doc

smaldini:
Regarding the builder suggestion here is the the proposal:
```java
		Sink<String> test = Sinks.multicast()
								 .onBackpressureBuffer()
								 .get();

		Sink<String> test2 = Sinks.replay()
								  .limit(Duration.ofMillis(100))
								  .get();

		Sink<String> test3 = Sinks.replay()
								  .limit(Duration.ofMillis(100))
								  .get();

		FluxProcessor<String, String> test4 = Sinks.unicast()
												   .onBackpressureBuffer()
												   .fluxProcessor();
```

reactor-core/src/main/java/reactor/core/publisher/Sinks.java
simonbasle: ok. the later might be a better solution actually, since then there is no risk of adding a (non-serialized) flavor of `MonoProcessor` and mistakenly wrapping it with a (non-serialized) sink decorator

reactor-core/src/main/java/reactor/core/publisher/MonoSink.java
simonbasle: wouldn't it make implementing the `tryNext` feature harder ? I'm guessing this overlaps a bit, but it could be interesting to be able to distinguish between a `emitNext` call that returns `false` because termination from one that returns `false` because backpressure. wdyt @dfeist ?

reactor-core/src/test/java/reactor/core/publisher/FluxAutoConnectTest.java
```diff
@@ -43,42 +43,42 @@ public void constructors() {
 	
 	@Test
 	public void connectImmediately() {
-		FluxProcessor<Integer, Integer> e = Processors.multicast();
+		Sink<Integer> e = Sinks.multicast();
```
simonbasle, line 46: since the sink isn't really used, can also be replaced with `TestPublisher` which makes it easier to assert subscribers:

```java
		TestPublisher<Integer> testPublisher = TestPublisher.create();

		testPublisher.flux().publish().autoConnect(0, cancel::set);
		testPublisher.assertSubscribers();
		
		cancel.get().dispose();
		testPublisher.assertNoSubscribers();
```

reactor-core/src/test/java/reactor/guide/GuideTests.java
```diff
@@ -201,21 +201,20 @@ public void advancedCold() {
 
 	@Test
 	public void advancedHot() {
-		Sink<String> hotSource = Sinks.multicastNoWarmup();
+		Sink<String> hotSource = Sinks.replay(0);
```
simonbasle, line 204: let's use `Sinks.multicast()`, `replay(0)` being kind of a corner case.
Also change the corresponding snippet in the asciidoc!

docs/asciidoc/advancedFeatures.adoc
```diff
@@ -193,7 +193,7 @@ Compare the first example to the second example, shown in the following code:
 ====
 [source,java]
 ----
-Sink<String> hotSource = Sinks.multicastNoWarmup();
+Sink<String> hotSource = Sinks.replay(0);
```
simonbasle, line 196: see comment on `GuideTest.java`

reactor-core/src/main/java/reactor/core/publisher/Sinks.java
```diff
@@ -47,25 +53,37 @@ private Sinks() { }
...
+	 */
+	@SuppressWarnings("deprecation")
+	public static <T> Sink<T> multicastNoBackpressure() {
+		DirectProcessor<T> processor = DirectProcessor.create();
+		return SinksHelper.toSink(processor, processor);
+	}
+
+	/**
+	 * A {@link Sink} with the following characteristics:
+	 * <ul>
```
simonbasle, line 78: Add a note that it is better suited to be viewed as a `toMono()` ?

reactor-core/src/main/java/reactor/core/publisher/Sinks.java
```diff
@@ -47,25 +53,37 @@ private Sinks() { }
...
+	 *     <li>Backpressure : this sink does not need any demand since it can only signal error or completion</li>
+	 *     <li>Replaying: Replay the terminal signal (error or complete).</li>
 	 * </ul>
-	 * <p>
-	 * <img class="marble" src="doc-files/marbles/sinkNoWarmup.svg" alt="">
 	 */
 	@SuppressWarnings("deprecation")
-	public static <T> Sink<T> multicastNoWarmup() {
-		return new ProcessorSink<>(ReplayProcessor.create(0));
+	public static Sink<Void> coordinator() {
```
simonbasle, line 85: put it next to `promise`, as both behave closer to `Mono`. That makes me wonder if it would be better backed by a `MonoProcessor`? likely to be used as `toMono()` more often and thus benefit from the 0-overhead conversion? or maybe add a constructor parameter to `DirectProcessor` that allows its `toMono()` method to do `Mono.fromDirect(this)` rather than `next()`?

reactor-core/src/main/java/reactor/core/publisher/Sinks.java
```diff
@@ -112,92 +132,73 @@ private Sinks() { }
...
-		return new ProcessorSink<>(UnicastProcessor.create());
+		UnicastProcessor<T> processor = UnicastProcessor.create();
+		return SinksHelper.toSink(processor, processor);
 	}
 
 	/**
 	 * A {@link Sink} that works like a conceptual Promise: it can be completed
 	 * with or without a value at any time, but only once. This completion is replayed to late subscribers.
-	 * Calling {@link Sink#next(Object)} is enough and does not require a following {@link Sink#complete()}.
+	 * Calling {@link Sink#emitNext(Object)} is enough and does not require a following {@link Sink#emitComplete()}.
```
simonbasle, line 142: add ` (which is implicit).` at the end

simonbasle: also add that as a result the `Sinks` closely follows the `Mono` contract and is better suited to be viewed as `toMono()` (hinting at 0-overhead conversion to mono vs overhead conversion to flux)

reactor-core/src/main/java/reactor/core/publisher/Sinks.java
```diff
@@ -112,92 +132,73 @@ private Sinks() { }
...
-			delegateSink.error(e);
-		}
-
-		@Override
-		public Sink<T> next(T t) {
-			delegateSink.next(t);
-			return this;
-		}
+	/**
+	 * Return unsafe {@link Sink} that are not serialized and expect the signal producer to respect
```
simonbasle, line 149: change to "and expect usage to be externally synchronized so as to respect"

reactor-core/src/main/java/reactor/core/publisher/EmitterProcessor.java
```diff
@@ -192,6 +192,85 @@ public void subscribe(CoreSubscriber<? super T> actual) {
...
+					q = queue;
+					if (q != null) {
+						break;
+					}
+				}
+			}
+		}
+
+		while (!q.offer(t)) {
+			LockSupport.parkNanos(10);
```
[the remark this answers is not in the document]
smaldini: I think we should just not support it at all this is a mistake and it feels like a documentable change for 3.4

reactor-core/src/main/java/reactor/core/publisher/FluxCreate.java
```diff
@@ -131,102 +130,102 @@ public Object scanUnsafe(Attr key) {
...
-		void drainLoop() {
-			Context ctx = sink.currentContext();
-			BaseSink<T> e = sink;
+		final void drainLoop() {
+			Sink<T> e = sink;
 			Queue<T> q = mpscQueue;
 			for (; ; ) {
 
 				for (; ; ) {
-					if (e.isCancelled()) {
```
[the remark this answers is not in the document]
smaldini: double checked but unfortunately there is no easy other way for now and i don't want to scope creep into changing this here. So reverted

reactor-core/src/main/java/reactor/core/publisher/FluxGroupJoin.java
```diff
@@ -207,7 +207,7 @@
 		@Override
 		public Stream<? extends Scannable> inners() {
 			return Stream.concat(
-					lefts.values().stream(),
+					lefts.values().stream().map(Scannable.class::cast),
```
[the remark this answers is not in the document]
smaldini: I'm not too sure, tangent, inners should have its own attribute (and delegate to it on the Scannable contract).
Will update in another PR if we want

reactor-core/src/main/java/reactor/core/publisher/FluxSink.java
```diff
@@ -33,9 +34,29 @@
  */
 public interface FluxSink<T> extends Sink<T> {
 
-	@Override
+	/**
+	 * Emit a non-null element, generating an {@link Subscriber#onNext(Object) onNext} signal.
```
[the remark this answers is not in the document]
smaldini: I'd tend to agree, you still have some chaining value out of next but error and complete are just there to align semantically with next. To be reviewed in a subsequent PR imo

reactor-core/src/main/java/reactor/core/publisher/MonoSink.java
smaldini: If we go that route we always bring back the enum `Emission` with various helper `isOk`, `isTerminated`, `isCancelled`, `isDisposed`, `isBackpressured`

reactor-core/src/main/java/reactor/core/publisher/MonoSink.java
```diff
@@ -55,11 +55,18 @@
 	void success(@Nullable T value);
 
 	/**
-	 * Terminate with the given value without requiring {@link #complete()} to be explicitly called.
+	 * @see #emitError(Throwable)
+	 */
+	void error(Throwable e);
+
+	/**
+	 * Terminate with the given value without requiring {@link #emitComplete()} to be explicitly called.
```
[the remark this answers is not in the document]
smaldini: Agree - Being consistent here with the hook there and processor/sink. That might be a separate PR to Improve documentation on Dropping behaviors.

smaldini:
Besides rebasing, the last touch on this PR : 
- If the guided Sinks api works for everyone, should we extend it to promise, adding its doc too : 
```java
Sink<Void> coordinator = Sinks.promise().coordinator(); // I have implemented a `VoidProcessor extends MonoProcessor` locally

Sink<String> nextPromise = Sinks.promise().next(); //the current MonoProcessor, locally moved under NextProcessor.

Sink<String> latestPromise = Sinks.promise().latest(); // a new processor I wanted to suggest that would emit the latest value only, useful for one off observers.

```
- Adding some common state observers instead of relying Scannable which is not a guaranteed state query: hasSubscribers, isTerminated(), hasError() / getError()
- For this pr or later changing emitterprocessor lock.parkNanos to onError: I don't think we should implement any sort of backpressure management on Sink, also i don't know about adding "offer()" or similar: Flux.create can already do that job in a more appropriate way using onRequest.

In a later iteration I'd like to see a `toConnectableMono` and `toConnectableFlux` on the sink while deprecating autoCancel on emitterProcessor. There should be a common way to address disposing between all sinks that is similar to Flux/Mono : refCount, connect etc. By default all `Sinks` except EmitterProcessor are similar to `share()`

reactor-core/src/test/java/reactor/core/publisher/UnicastProcessorTest.java
```diff
@@ -65,20 +65,19 @@ public void secondSubscriberRejectedProperly() {
 
 	@Test
 	public void multiThreadedProducer() {
-		UnicastProcessor<Integer> processor = UnicastProcessor.create();
-		FluxSink<Integer> sink = processor.sink();
+		Sink<Integer> sink = Sinks.unicast().onBackpressureBuffer();
```
simonbasle, line 68: I'm not sure `xxxProcessor` tests should use the `Sinks` path. After all, these classes are there to explicitly test a specific concrete implementation and would likely survive the switch of said concrete implementation to package-private. On the other hand, if the implementation of `Sinks.unicast()` somehow changes then coverage would drop for `UnicastProcessor`.

This comment applies to other `xxxProcessorTest` cases.

docs/asciidoc/advancedFeatures.adoc
```diff
@@ -193,7 +193,7 @@ Compare the first example to the second example, shown in the following code:
 ====
 [source,java]
 ----
-Sinks.StandaloneFluxSink<String> hotSource = Sinks.multicastNoWarmup();
+Sink<String> hotSource = Sinks.replay().limit(0);
```
simonbasle, line 196: see comment in `GuideTest.java`

reactor-core/src/test/java/reactor/guide/GuideTests.java
```diff
@@ -201,21 +201,20 @@ public void advancedCold() {
 
 	@Test
 	public void advancedHot() {
-		Sinks.StandaloneFluxSink<String> hotSource = Sinks.multicastNoWarmup();
-
-		Flux<String> hotFlux = hotSource.asFlux().map(String::toUpperCase);
+		Sink<String> hotSource = Sinks.replay().limit(0);
```
simonbasle, line 204: reiterating that `replay().limit(0)` is probably a bit too confusing. `multicast().onBackpressureError()` sounds like the best example to give, showcasing only live data makes it past the sink.

Also change in `advancedFeatures.adoc`.

simonbasle:
> ```java
> Sink<Void> coordinator = Sinks.promise().coordinator(); // I have implemented a `VoidProcessor extends MonoProcessor` locally
> 
> Sink<String> nextPromise = Sinks.promise().next(); //the current MonoProcessor, locally moved under NextProcessor.
> 
> Sink<String> latestPromise = Sinks.promise().latest(); // a new processor I wanted to suggest that would emit the latest value only, useful for one off observers.
> ```

I'm good with introducing a spec for mono processors as a follow up, but I think these changes *should be done in a separate PR*. `VoidProcessor` sounds good by itself, but the rework of `MonoProcessor` makes me feel this doesn't belong in this particular PR. Plus there are two dedicated issues for that and `LatestProcessor` : #1053 and #1310 (looks like the second one could actually cover both `next()` and `latest()`). On a side note the names for the API would need some more thought.


> 
>     * Adding some common state observers instead of relying Scannable which is not a guaranteed state query: hasSubscribers, isTerminated(), hasError() / getError()

Ok yeah, why not.

> 
>     * For this pr or later changing emitterprocessor lock.parkNanos to onError: I don't think we should implement any sort of backpressure management on Sink, also i don't know about adding "offer()" or similar: Flux.create can already do that job in a more appropriate way using onRequest.

Since there is a dedicated issue (#2049) it would make sense in another PR. Would greatly appreciate if you have the time to tackle it 🙇 
> 
> 
> In a later iteration I'd like to see a `toConnectableMono` and `toConnectableFlux` on the sink while deprecating autoCancel on emitterProcessor. There should be a common way to address disposing between all sinks that is similar to Flux/Mono : refCount, connect etc. By default all `Sinks` except EmitterProcessor are similar to `share()`

Yeah that sounds like a nice improvement in consistency. The only thing is we don't have `ConnectableMono` :)

reactor-core/src/main/java/reactor/core/publisher/Sinks.java
smaldini: In a separate PR we will have VoidProcessor for this

reactor-core/src/main/java/reactor/core/publisher/FluxGroupJoin.java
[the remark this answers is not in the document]
smaldini: good catch

reactor-core/src/main/java/reactor/core/publisher/DelegateSinkFluxProcessor.java
```diff
@@ -0,0 +1,120 @@
...
+
+	DelegateSinkFluxProcessor(Flux<IN> downstream,
+							  Sink<IN> upstream) {
+		this.downstream = Objects.requireNonNull(downstream, "Downstream must not be null");
+		this.upstream = Objects.requireNonNull(upstream, "Upstream must not be null");
+	}
+
+	@Override
+	public Context currentContext() {
+		if(upstream instanceof CoreSubscriber){
```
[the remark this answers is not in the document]
smaldini: The same existed in FluxProcessor but yeah upstream is the receiving side and downstream the producing side here. Can change to less ambiguous term tho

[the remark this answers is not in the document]
simonbasle: yeah, had to double-back during the review. I think it would benefit readability for our future selves 😄

rstoyanchev:
I like where this has gone from a high-level usage perspective, it looks much cleaner.

One suggestion for consideration. The terminology for the entry points in `Sinks` is a bit mixed. It's not as self-evident as could be. For example `coordinator` and `promise` stand out the most. We probably don't want to introduce such terms that compete with `Mono<T>` and `Mono<Void>`. For such method names it probably won't be too hard to find something more plain.

One thought I had it is something like this:

- `Sinks`
  - `completion()` -> `Sink<Void>`
  - `item()` -> `Sink<T>`
  - `sequence()` -> `SequenceSpec`
    - `unicast()` -> `UnicastSpec`
    - `multicast()` -> `MulticastSpec`
    - `replay()` -> `ReplaySpec`

smaldini:
Hey @rstoyanchev and @simonbasle I'll get back to this over this weekend unless i can push some update today. Locally I stopped on having to separate the work on MonoProcessor to support Mono<Void> in a different PR. 
I agree with the suggestion you made @rstoyanchev in terms of organizing first by cardinality. In fact I wondered about : 
```Sinks.zero(), Sinks.one(), Sinks.many()``` 

The reason is to deemphasize on the item (`Sinks.one` could be at most one where `item()` feels like you need to produce one item). Variants include `Sinks.empty()`, `Sinks.singleOrEmpty()` and `Sinks.many()` but we can adapt naming at will once the concept is in. 
So leftover is to polish with @simonbasle comments and move the rest to a further iteration (scannable and connectable additions).

**Comment by smaldini** [comment]
Better late than never they said - So in this push :
- Split the PR (the other one on MonoProcessor to come)
- The reorg using Sinks.empty, singleOrEmpty and many
- Added the Emission enum.
- Also changed the EmitterProcessor behavior to not lock

File: reactor-core/src/test/java/reactor/guide/GuideTests.java (hunk 21 above)
Comment (simonbasle) [reply reactor-core/src/test/java/reactor/guide/GuideTests.java]: fixed in 7b6b655

File: docs/asciidoc/advancedFeatures.adoc (hunk 20 above)
Comment (simonbasle) [reply docs/asciidoc/advancedFeatures.adoc]: fixed in 7b6b655

File: reactor-core/src/main/java/reactor/core/publisher/SinksSpecs.java [hunk 23]
```diff
@@ -0,0 +1,174 @@
...
+	static final MulticastSpecImpl       MULTICAST_SPEC               = new MulticastSpecImpl(true);
+	static final MulticastReplaySpecImpl MULTICAST_REPLAY_SPEC        = new MulticastReplaySpecImpl(true);
+	static final UnsafeSpecImpl          UNSAFE_SPEC                  = new UnsafeSpecImpl();
+	static final UnicastSpecImpl         UNSAFE_UNICAST_SPEC          = new UnicastSpecImpl(false);
+	static final MulticastSpecImpl       UNSAFE_MULTICAST_SPEC        = new MulticastSpecImpl(false);
+	static final MulticastReplaySpecImpl UNSAFE_MULTICAST_REPLAY_SPEC = new MulticastReplaySpecImpl(false);
+
+}
+
+final class FluxProcessorSink<T> extends FluxCreate.SerializedSink<T, Sink<T>> {
```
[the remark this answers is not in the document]
simonbasle: splitted out and renamed in 6f60e9f

reactor-core/src/main/java/reactor/core/publisher/FluxSink.java
simonbasle: fixed in 6f60e9f

simonbasle:
It should be noted that this PR incidentally fixes #2049

reactor-core/src/main/java/reactor/core/publisher/Sinks.java
```diff
@@ -87,118 +78,315 @@ private Sinks() { }
...
+		<T> Sink<T> onBackpressureBuffer(int bufferSize, boolean autoCancel);
+
+		/**
+		 * A {@link Sink} with the following characteristics:
+		 * <ul>
+		 *     <li>Multicast</li>
+		 *     <li>Without {@link Subscriber}: elements pushed via {@link Sink#emitNext(Object)} are discarded</li>
+		 *     <li>Backpressure : this sink is not able to honor downstream demand and will emit `onError` if there is a mismatch.</li>
+		 *     <li>Replaying: No replay. Only forwards to a {@link Subscriber} the elements that have been
+		 *     pushed to the sink AFTER this subscriber was subscribed. To the exception of the first
```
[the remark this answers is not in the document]
simonbasle: fixed in 8cb64857c

simonbasle:
@reactor/core-team I have reworked the PR according to our internal discussions, can you review and approve?

reactor-core/src/test/java/reactor/core/publisher/FluxFilterWhenTest.java
```diff
@@ -254,13 +254,13 @@ protected void hookFinally(SignalType type) {
 
 	@Test
 	public void cancel() {
-		final FluxIdentityProcessor<Boolean> pp = Processors.multicast();
+		final Sinks.Many<Boolean> pp = Sinks.many().multicast().onBackpressureBuffer();
```
aneveu, line 257: not really important but the name of the variable was already quite unclear before and it becomes completely impossible to link to anything now 😅

PR #2219, opened 2020-06-24
provides stats operators and supporting utility

OlegDokuka:
This PR introduces a Stats utility that allows stats reporting to a logger

## Motivation 

Usually, it is challenging to analyze what is going on within your data pipeline. Metrics are useful to get general "performance" insights for your stream. Debug hooks allow detecting the root cause operator. However, there is a gap in detecting general issues, like timeouts, which operator underperform, etc.

## Solution

In order to simplify the process of the mentioned problems detections, this PR introduces an extra `StatsOperator` which builds a graph from the whole execution pipeline and then collect all important information like:

1. The Number of produced items.
1. The Number of requested items.
1. The average execution time of that particular operator (is different from one provided by metrics).
1. The Last Observed signal (e.g. what happened last before `cancel | error | complete`)
1. The Current State of that operator (e.g. `canceled` | `completed` | `errored` | `none`)

All this information is printed after the execution in the well-formed stack trace like output which looks like the following 

```
10:43:19.381 [parallel-1] INFO  r.c.publisher.DefaultStatsReporter - The following stats are collected:
  	|  Flux.range       ⇢ at reactor.examples.ReportRootHangingTest.timeoutNightmare(ReportRootHangingTest.java:23)
  	|_                  ↳ Stats(Requested: 256; Produced: 100; OnNext ~Time: 450ns; Last Signal:  onNext; State: canceled)
  	|  Flux.map         ⇢ at reactor.examples.ReportRootHangingTest.timeoutNightmare(ReportRootHangingTest.java:24)
  	|_                  ↳ Stats(Requested: 256; Produced: 100; OnNext ~Time: 728ns; Last Signal:  onNext; State: canceled)
 	|  Flux.transform   ⇢ at reactor.examples.ReportRootHangingTest.timeoutNightmare(ReportRootHangingTest.java:25)
  	|_                  ↳ Stats(Requested: 256; Produced:   0; OnNext ~Time:   0ns; Last Signal: request; State: canceled)
  	|  Flux.map         ⇢ at reactor.examples.ReportRootHangingTest.timeoutNightmare(ReportRootHangingTest.java:29)
  	|_                  ↳ Stats(Requested: 256; Produced:   0; OnNext ~Time:   0ns; Last Signal: request; State: canceled)
  	|  Flux.subscribeOn ⇢ at reactor.examples.ReportRootHangingTest.timeoutNightmare(ReportRootHangingTest.java:30)
  	|_                  ↳ Stats(Requested: 256; Produced:   0; OnNext ~Time:   0ns; Last Signal: request; State: canceled)
  	|  Flux.filter      ⇢ at reactor.examples.ReportRootHangingTest.timeoutNightmare(ReportRootHangingTest.java:31)
  	|_                  ↳ Stats(Requested: 256; Produced:   0; OnNext ~Time:   0ns; Last Signal: request; State: canceled)
  	|  Flux.publishOn   ⇢ at reactor.examples.ReportRootHangingTest.timeoutNightmare(ReportRootHangingTest.java:33)
  	|_                  ↳ Stats(Requested:   ∞; Produced:   0; OnNext ~Time:   0ns; Last Signal: request; State: canceled)
  	|  Flux.timeout     ⇢ at reactor.examples.ReportRootHangingTest.timeoutNightmare(ReportRootHangingTest.java:34)
  	|_                  ↳ Stats(Requested:   ∞; Produced:   0; OnNext ~Time:   0ns; Last Signal: request; State:  errored)
  	|  Flux.metrics     ⇢ at reactor.examples.ReportRootHangingTest.timeoutNightmare(ReportRootHangingTest.java:35)
  	|_                  ↳ Stats(Requested:   ∞; Produced:   0; OnNext ~Time:   0ns; Last Signal: request; State:  errored)

```

In order to print such an output, `StatsOperator` uses `AssemblySnapshot`. Also, there is an abstract utility like `StatsReporter` which allows configuring where such output should be printed and what is the format of the output (e.g. one might want to see such an out as Graphviz diagram).

Also, this PR introduced output analyzers, called `StatsMarkers` which can add an extra marker to the output to say that this particular operator is the source of, for instance, timeout:

```
10:43:19.381 [parallel-1] INFO  r.c.publisher.DefaultStatsReporter - The following stats are collected:
...
🔎	|  Flux.transform   ⇢ at reactor.examples.ReportRootHangingTest.timeoutNightmare(ReportRootHangingTest.java:25)
  	|_                  ↳ Stats(Requested: 256; Produced:   0; OnNext ~Time:   0ns; Last Signal: request; State: canceled)
...

```

## HowTo Use 

In order to track advanced debug information and collect reports, you just need to use `Hook` API and enable the following 2 properties:

```java
Hooks.enableStatsRecording();
Hooks.onOperatorDebug();
```

Having that enabled, you may run your app, and if an exception occurs, you will see a detailed report which has extra info on what might cause the error 

### DIFF
--- a/reactor-core/src/main/java/reactor/core/publisher/ConnectableFluxStats.java
+++ b/reactor-core/src/main/java/reactor/core/publisher/ConnectableFluxStats.java
@@ -1,0 +1,107 @@
+package reactor.core.publisher;
+
+import java.util.function.Consumer;
+
+import org.reactivestreams.Publisher;
+import reactor.core.CoreSubscriber;
+import reactor.core.Disposable;
+import reactor.core.Fuseable;
+import reactor.core.Scannable;
+import reactor.util.stats.StatsMarker;
+import reactor.util.stats.StatsReporter;
+
+public class ConnectableFluxStats<T> extends InternalConnectableFluxOperator<T, T> {
+
+	final FluxOnAssembly.AssemblySnapshot assemblySnapshot;
+	final StatsReporter                   statsReporter;
+	final StatsMarker[]                   statsMarkers;
+
+	protected ConnectableFluxStats(ConnectableFlux<T> source,
+			FluxOnAssembly.AssemblySnapshot assemblySnapshot,
+			StatsReporter statsReporter,
+			StatsMarker[] statsMarkers) {
+		super(source);
+		this.assemblySnapshot = assemblySnapshot;
+		this.statsReporter = statsReporter;
+		this.statsMarkers = statsMarkers;
+	}
+
+	@Override
+	public void connect(Consumer<? super Disposable> cancelSupport) {
+		source.connect(cancelSupport);
+	}
+
+	@Override
+	public CoreSubscriber<? super T> subscribeOrReturn(CoreSubscriber<? super T> actual) {
+
+		if (nextOptimizableSource() == null) {
+			return new ConnectableStatsInner<>(actual,
+					assemblySnapshot,
+					source,
+					statsReporter,
+					statsMarkers);
+		}
+		else {
+			return new FluxStats.StatsOperator<>(actual,
+					assemblySnapshot,
+					source,
+					statsReporter,
+					statsMarkers);
+		}
+	}
+
+	static class ConnectableStatsInner<T> extends FluxStats.StatsOperator<T>
+			implements InnerOperator<T, T>, Fuseable.QueueSubscription<T> {
+
+		ConnectableStatsInner(CoreSubscriber<? super T> actual,
+				FluxOnAssembly.AssemblySnapshot snapshotStack,
+				Publisher<?> parent,
+				StatsReporter statsReporter,
+				StatsMarker[] statsMarkers) {
+
+			super(actual, snapshotStack, parent, statsReporter, statsMarkers);
+		}
+
+		@Override
+		public void onNext(T t) {
+			resolveUpstream();
+			super.onNext(t);
+		}
+
+		@Override
+		public void onError(Throwable t) {
+			resolveUpstream();
+			super.onError(t);
+		}
+
+		@Override
+		public void onComplete() {
+			resolveUpstream();
+			super.onComplete();
+		}
+
+		@Override
+		public void cancel() {
+			resolveUpstream();
+			FluxStats.StatsOperator<?> connectableStatsOperator = this.upstreamNode;
+			if (connectableStatsOperator instanceof FluxStats.MultiConsumersStatsOperator) {
+				((FluxStats.MultiConsumersStatsOperator<?>) connectableStatsOperator).disconnect();
+			}
+			super.cancel();
+		}
+
+		void resolveUpstream() {
+			if (this.upstreamNode == null) {
+				FluxStats.MultiConsumersStatsOperator<?>
+						multiConsumersStatsOperator = (FluxStats.MultiConsumersStatsOperator<?>) Scannable.from(s)
+						                                                                                  .parents()
+						                                                                                  .filter(FluxStats.MultiConsumersStatsOperator.class::isInstance)
+						                                                                                  .findFirst()
+						                                                                                  .orElse(null);
+
+				this.upstreamNode = multiConsumersStatsOperator;
+				multiConsumersStatsOperator.connect();
+			}
+		}
+	}
+}
--- a/reactor-core/src/main/java/reactor/core/publisher/ConnectableFluxStats.java
+++ b/reactor-core/src/main/java/reactor/core/publisher/ConnectableFluxStats.java
@@ -10,13 +10,13 @@ import reactor.core.Scannable;
 import reactor.util.stats.StatsMarker;
 import reactor.util.stats.StatsReporter;
 
-public class ConnectableFluxStats<T> extends InternalConnectableFluxOperator<T, T> {
+final class ConnectableFluxStats<T> extends InternalConnectableFluxOperator<T, T> {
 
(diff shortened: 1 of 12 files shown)


reactor-core/src/main/java/reactor/core/publisher/ConnectableFluxStats.java
```diff
@@ -0,0 +1,107 @@
...
+
+import org.reactivestreams.Publisher;
+import reactor.core.CoreSubscriber;
+import reactor.core.Disposable;
+import reactor.core.Fuseable;
+import reactor.core.Scannable;
+import reactor.util.stats.StatsMarker;
+import reactor.util.stats.StatsReporter;
+
+public class ConnectableFluxStats<T> extends InternalConnectableFluxOperator<T, T> {
```
bsideup, line 13: can this and other classes be package-private?

OlegDokuka: fixed

reactor-core/src/main/java/reactor/core/publisher/Hooks.java
```diff
@@ -516,6 +517,27 @@ public static void disableContextLossTracking() {
 		DETECT_CONTEXT_LOSS = false;
 	}
 
+	/**
+	 * Globally enables the {@link Context} loss detection in operators like
```
ericbottard, line 521: These javadocs seem to be a copy/paste of the `ContextLossTracking` feature, *not* `StatsTracking`

reactor-core/src/main/java/reactor/core/publisher/Hooks.java
```diff
@@ -655,21 +684,37 @@ static boolean initStaticGlobalTrace() {
 	}
 
 	static <T, P extends Publisher<T>> Publisher<T> addAssemblyInfo(P publisher, AssemblySnapshot stacktrace) {
-		if (publisher instanceof Callable) {
+		if (GLOBAL_STATS_TRACE) {
```
ericbottard, line 687: I'm not sure I grok all the details of this PR, but shouldn't this be _additive_ to other hooks that were already there?
If it still is with this PR (via code duplication), isn't this problematic in the long run? What if we later add a **third** concern on top of stats collecting? And a fourth, _etc._

OlegDokuka: > I'm not sure I grok all the details of this PR, but shouldn't this be _additive_ to other hooks that were already there?

This feature is *additive* as much as possible. It reuses assembly info in order to collect operators location. It is fully compatible with ReactorDebugAgent which should simplify future users adoption. I'm not still sure that enabling of this future should live separately, but I belive @bsideup can suggest one. I watch thinking about overloaded `onOperatorDebug(boolean enables tats) ` but this will not work with ReactorDebugAgent. Thus the best is a separate hooks method 



> If it still is with this PR (via code duplication). 

Can you please be more certain here? I m not sure about which code duplication your are talking about. 

> , isn't this problematic in the long run? What if we later add a **third** concern on top of stats collecting? And a fourth, _etc._

If you can imagine any extra concerns - all suggestions are welcome. My main goal is to simplify users life. 

Cheers, 
Oleh

reactor-core/src/main/java/reactor/core/publisher/Hooks.java
OlegDokuka: Fixed 🙏

reactor-core/src/main/java/reactor/core/publisher/Hooks.java
OlegDokuka: I guess you are missing a few important points of the full picture. 

1, The main point is that all these methods are internal. 
1. The usage of those internal methods is only within the reactor `onAssembly` or by ReactorDebugAgent. 
1. Those methods clearly state (@Deprecated only for internal usage)
1. From the user perspective, a user has a separate configuration (`enableStatsRecording`) which is SEPARATE from `onOperatorDebug` which enables Stack Trace tracking. These features a coupled. In order to have stats we need operator name and location, so without onOperatorDebug enableStatsReacording should not work. Nevertheless, the implementation of Flux decoration is fully different, so mixing both will be misleading. 

Alright, imagine we follow the suggestions and create a separate method that does the same as `addAssemblyInfo` and call it `addStats`. Now we would need to :

image


1. Add in 6 more places (see image) if-else statement to follow different branches.
1. Every time we would need to  `addAssemblyInfo` -  we would need to maintain this extra if-else branching.
1. The third-party concern would lead to increasing the number of if-else branches in these 6 or more places

------------------------------------------------------------------------------------------


Alright. I guess it is clear that moving that logic from `AddAssemblyInfo` will be even more challenging than support a few if branches in it. I hope we can agree that we should keep the code in this method. 

Anyways, If we wanna code clarity - I can create 2 methods that are being called from `addAssemblyInfo` since all these are assembly info (the first is `addOperatorsStackTraceInfo`  another is `addOperatorStatsInfo`). Should it eliminate the problem being discussed? 

Cheers,
Oleh

ericbottard: > Should it eliminate the problem being discussed?

I think that would be a good thing to do, yes. Thanks !

OlegDokuka: fixed

chemicL:
The last interaction here was over 3 years ago. Closing due to inactivity. Please create a new PR targeting the latest codebase in case you feel this is still applicable.

PR #2254, opened 2020-07-06
Architecture schema for Publishers and Subscribers implementation (#2252)

aneveu:
Please let me know what you want to be improved/added/removed 🙏 

PublishersAndSubscribers


bsideup:
@aneveu nice! I just noticed that `BaseSubscriber` is marked as an interface, while in fact it is an abstract class :)

simonbasle:
A small `legend` block to better tell what plain background vs light background means would be good to have. IIRC plain colors was for public elements, but `CoreSubscriber` is light despite being public, for instance. 

The graph looks quite good ;)

ericbottard:
I understand this is svg when in source form right? If so, would it be possible to add hover texts (is that possible?) that describe what each class does / adds compared to other class (ideally the first sentence of each javadoc, but we may discover that those are not always present)

aneveu:
> I understand this is svg when in source form right? If so, would it be possible to add hover texts (is that possible?) that describe what each class does / adds compared to other class (ideally the first sentence of each javadoc, but we may discover that those are not always present)

The source format is png (no other options). I could probably figure out a way to add hover but I'm afraid it becomes unreadable as there is already a lot of information on it. As mentioned during the weekly last week, [structure 101](https://structure101.com/products/workspace) might handle this more conveniently.

ericbottard:
I meant an hover box *per class* (which svg may have allowed). If the source is png, forget it this won't indeed be readable at all. 

The drawing is already readable as it is, nice work!

aneveu:
> A small `legend` block to better tell what plain background vs light background means would be good to have. IIRC plain colors was for public elements, but `CoreSubscriber` is light despite being public, for instance.
> 
> The graph looks quite good ;)

fixed and preview updated in the comment

aneveu:
> @aneveu nice! I just noticed that `BaseSubscriber` is marked as an interface, while in fact it is an abstract class :)

fixed and preview updated in the comment

simonbasle:
on top of my review, I'd try to explore ways to group / differentiate classes that are about writing operators vs classes that are part of the public API:
 - `GroupedFlux`, `ConnectableFlux`, `FluxProcessor` and `MonoProcessor` are almost at the same level as `Flux`/`Mono`/`ParallelFlux`, in the sense that they are really part of the public API as experienced by end users
 - `FluxOperator`, `FluxFromMonoOperator`, `MonoOperator` and `MonoFromFluxOperator` are more for helping writing custom operators, and as such are "closer" to eg. `InnerOperator` (despite being public, where InnerOperator isn't)
 - `OptimizableOperator` and its child classes are one level further below, package-private and aimed at optimizing core operators

Maybe try to reorganize these differently ? And/or combine that with a background (like the rectangle background behind Reactive Streams classes) ?

aneveu:
New version for preview before moving to Inkscape, with changes requested above and without `processors`
Reactor-Core 3 4@2x

aneveu:
Try to simplify the schema by illustrating the implementation of `OptimizableOperator` through color Reactor-Core 3 4 v2@2x

simonbasle:
in the svg version (and maybe directly in the sketching tool you used @aneveu) the internal optimizable nature could be shown by grouping the blocks on an "optimizable" background the same way Reactive Streams classes have their own zone, wdyt?

aneveu:
I can give it a try, would you prefer to keep this alignment or the previous one?

simonbasle:
@aneveu this one, it is definitely an improvement, less clutter 👍 it's just that the 3rd graphic style in the legend is less visible/distinguishable, hence my suggestion with a background / group instead

aneveu:
Here it is:
Reactor-Core 3 4@2x

simonbasle:
@aneveu nice 👍

simonbasle:
this was a good experiment in mapping out the internal classes of Reactor.
unfortunately, the SVG version turned to be hard to produce due to technical difficulties in running Inkscape. we won't devote more time to this issue but might come back at it later by reusing this basis.

PR #2273, opened 2020-07-17
fix #2264 add blockhound to testset

### LINKED ISSUE #2264: Isolate BlockHound in a dedicated gradle configuration
BlockHound is currently marked as "optional" in the main gradle configuration while being used in a dedicated [test set](https://github.com/unbroken-dome/gradle-testsets-plugin).

This has the effect that related sources don't compile in Intellij. 
Screen Shot 2020-07-10 at 11 45 00
Putting BlockHound in the testset config would fix the problem.

See https://github.com/reactor/reactor-core/pull/2261#discussion_r451607753

ericbottard:
This fixes compilation errors in IntelliJ

### DIFF
--- a/reactor-core/build.gradle
+++ b/reactor-core/build.gradle
@@ -69,6 +69,8 @@ dependencies {
 
 	//Optional BlockHound support
 	optional "io.projectreactor.tools:blockhound:$blockhoundVersion"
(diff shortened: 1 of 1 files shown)


ericbottard:
BH still left as "optional" as it needs to be in the main set **as well** for `ReactorBlockHoundIntegration`.

PR #2348, opened 2020-08-29
[build] Add Gradle wrapper validation

sullis:
https://blog.gradle.org/gradle-wrapper-checksum-verification-github-action

### DIFF
--- a/.github/workflows/gradle-wrapper-validation.yml
+++ b/.github/workflows/gradle-wrapper-validation.yml
@@ -1,0 +1,10 @@
+name: "Validate Gradle Wrapper"
+on: [push, pull_request]
+
+jobs:
+  validation:
+    name: "Gradle wrapper validation"
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: gradle/wrapper-validation-action@v1


simonbasle (approved):
thanks for the PR 🎁 @sullis

.github/workflows/gradle-wrapper-validation.yml
```diff
@@ -0,0 +1,10 @@
+name: "Validate Gradle Wrapper"
+on: [push, pull_request]
+
+jobs:
+  validation:
+    name: "Gradle wrapper validation"
```
bsideup, line 6: nit: there seems to be a duplication that makes it show "Validate Gradle Wrapper / Gradle wrapper validation" in the checks. The job's name could really be "Validate" or something similar. Or maybe even "Gradle Wrapper" as the action's name and "validate" as the job's, so that we get "Gradle Wrapper / validate" in the checks.

/cc @simonbasle 
Screenshot 2020-09-01 at 11 28 17 AM

simonbasle: good point, I'll rename it to `Gradle Wrapper / validate`

PR #2422, opened 2020-10-07
Use BinaryOperator in reduce and scan

camsteffen:
Simplifies the API just a tad. Unfortunately this is not strictly backward compatible.

### DIFF
--- a/reactor-core/src/main/java/reactor/core/publisher/Flux.java
+++ b/reactor-core/src/main/java/reactor/core/publisher/Flux.java
@@ -7363,10 +7363,10 @@ public abstract class Flux<T> implements CorePublisher<T> {
 	}
 
 	/**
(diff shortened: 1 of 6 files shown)


simonbasle (changes requested):
It would have been preferable to use `BinaryOperator` in the first place, I kinda agree... But that change is IMHO low value high impact.

These operators are in the basic toolbox, and there's a high chance a lot of reactive apps use one of these.

Making the switch has the only visible benefit of a shorter method signature due to generics: low value. But it has high impact: potential compilation errors for people that externalize the `BiFunction` into a variable for instance.

Ok the migration would be pretty easy, but given the low benefit even such a small hurdle should be avoided.

camsteffen:
Makes sense

PR #2440, opened 2020-10-15
fix #2428 update to jcstress 0.7

### LINKED ISSUE #2428: Cut down jcstress logging

ericbottard:
This reduces the amount of logging

### DIFF
--- a/reactor-core/build.gradle
+++ b/reactor-core/build.gradle
@@ -26,6 +26,7 @@ buildscript {
 	}
 
 	dependencies {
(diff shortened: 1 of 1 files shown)


ericbottard:
This comes as an alternative to https://github.com/reactor/reactor-core/pull/2435

simonbasle (approved):

PR #2516, opened 2020-11-30
fix #2510 rework asciidoctor integration

### LINKED ISSUE #2510: Asciidoc documentation references images online using raw.githubusercontent.com

ericbottard:
* upgrade to current, new generation of plugin
* fix images to use local versions instead of online
* fix way images are referenced, allows inline view even when a single
  file is rendered in IDE and unaware of attributes
* fix missing project version attribute
* add syntax highlighting in pdf with rouge

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -27,7 +27,8 @@ buildscript {
 
 plugins {
 	id "com.github.johnrengelman.shadow" version "4.0.2"
-	id 'org.asciidoctor.convert' version '1.5.11'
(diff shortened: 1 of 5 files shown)


ericbottard:
Something is wrong with this, somehow my changes to gradle/asciidoctor.gradle are gone Grrr. Will re-open.

PR #2555, opened 2021-01-14
provides debug stats implementation

OlegDokuka:
This PR drafts a simple implementation of stats explained in #2219

### DIFF
--- a/reactor-core/src/main/java/reactor/core/publisher/FluxOnAssembly.java
+++ b/reactor-core/src/main/java/reactor/core/publisher/FluxOnAssembly.java
@@ -15,6 +15,8 @@
  */
 package reactor.core.publisher;
 
+import java.time.Instant;
+import java.time.format.DateTimeFormatter;
 import java.util.Arrays;
 import java.util.LinkedList;
 import java.util.List;
@@ -27,8 +29,6 @@ import reactor.core.Exceptions;
 import reactor.core.Fuseable;
 import reactor.core.Scannable;
 import reactor.util.annotation.Nullable;
-import reactor.util.function.Tuple4;
-import reactor.util.function.Tuples;
 
 /**
  * Captures the current stacktrace when this publisher is created and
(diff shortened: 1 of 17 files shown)


reactor-core/src/main/java/reactor/core/publisher/FluxOnAssembly.java
```diff
@@ -95,14 +97,22 @@ static void fillStacktraceHeader(StringBuilder sb, Class<?> sourceClass, @Nullab
 	@SuppressWarnings("unchecked")
 	static <T> CoreSubscriber<? super T> wrapSubscriber(CoreSubscriber<? super T> actual,
 														Flux<? extends T> source,
-														@Nullable AssemblySnapshot snapshotStack) {
-		if(snapshotStack != null) {
+														@Nullable AssemblySnapshot snapshotStack,
```
bsideup, line 100: nit: something is terribly wrong with the indentation here :D

reactor-core/src/main/java/reactor/core/publisher/FluxOnAssembly.java
```diff
@@ -298,37 +443,123 @@ public String getMessage() {
...
+						}
+					}
 					sb.append("\n");
 				}
 				sb.append("Stack trace:");
 				return sb.toString();
 			}
 		}
+
+		static class Entry {
```
bsideup, line 542: Please give it a meaningful name

reactor-core/src/main/java/reactor/core/publisher/FluxOnAssembly.java
```diff
@@ -298,37 +443,123 @@ public String getMessage() {
...
+
+			final int        key;
+			final int        indent;
+			final String     operator;
+			final String     line;
+			@Nullable
```
bsideup, line 548: What about the idea of No-op `DebugStats` that we discussed?

OlegDokuka: It can be challenging to then check if the DebugStats are real or noops when it comes to the printing of the message (and in the end it will end up to check `if debugStats instanceof NoOpsDebugStats`), thus I decided to stick to null check since that logic is already present in the message printing and JIT should eliminate such checks at runtime in any case

PR #2566, opened 2021-01-22
Add hook to wrap Queues

bsideup:


OlegDokuka (approved):
Looks very promising! Great Job!

simonbasle (approved):

bsideup:
Here is an example of how this can be used btw:
```java
package reactor.patterns;

import org.assertj.core.api.SoftAssertions;

import java.time.Instant;
import java.util.AbstractQueue;
import java.util.Iterator;
import java.util.Queue;
import java.util.concurrent.TimeUnit;

public class WorkStealing {

    private static final ThreadLocal<String> SPAN_ID = ThreadLocal.withInitial(() -> null);

        instrumentRunnables();
        instrumentQueues();

        SoftAssertions.assertSoftly(softly -> {
            try {
                SPAN_ID.set("main");
                Flux.range(0, 20)
                        .doOnNext(i -> {
                            SPAN_ID.set("span-" + i);
                        })
                        .parallel(4, 4)
                        .runOn(Schedulers.parallel())
                        .doOnNext(i -> {
                            String spanId = SPAN_ID.get();

                            softly.assertThat(spanId)
                                    .as("currentSpan of " + i)
                                    .isEqualTo("span-" + i);

                            System.out.printf(
                                    "%s [%s] [%s] processing %d%n",
                                    Instant.now(),
                                    Thread.currentThread().getName(),
                                    spanId,
                                    i
                            );
                        })
                        .then()
                        .toFuture()
                        .get(10, TimeUnit.SECONDS);
            } catch (Exception e) {
            }
        });
    }

    private static void instrumentRunnables() {
        Schedulers.onScheduleHook(
                "tracing",
                runnable -> {
                    // Capture
                    String spanId = SPAN_ID.get();

                    // Propagate
                    return () -> {
                        String old = SPAN_ID.get();

                        try {
                            SPAN_ID.set(spanId);
                            runnable.run();
                        } finally {
                            // Restore
                            SPAN_ID.set(old);
                        }
                    };
                }
        );
    }

    private static void instrumentQueues() {
        Hooks.addQueueWrapper(
                "tracing",
                queue -> {
                    Queue<Envelope> envelopeQueue = (Queue<Envelope>) queue;
                    return new AbstractQueue<Object>() {

                        @Override
                        public int size() {
                            return envelopeQueue.size();
                        }

                        @Override
                        public boolean offer(Object o) {
                            String span = SPAN_ID.get();
                            return envelopeQueue.offer(new Envelope(o, span));
                        }

                        @Override
                        public Object poll() {
                            Envelope envelope = envelopeQueue.poll();
                            if (envelope == null) {
                                return null;
                            }
                            SPAN_ID.set((String) envelope.span);
                            return envelope.body;
                        }

                        @Override
                        public Object peek() {
                            throw new UnsupportedOperationException();
                        }

                        @Override
                        public Iterator<Object> iterator() {
                            throw new UnsupportedOperationException();
                        }
                    };
                }
        );
    }

    static class Envelope {
        final Object body;

        final Object span;

        Envelope(Object body, Object span) {
            this.body = body;
            this.span = span;
        }
    }

}
```

PR #2576, opened 2021-01-27
Fix OverflowException in MultiSubscriptionSubscriber

koldat:
Fixes:

```java
public class FFTest {
    private static final Logger log = LoggerFactory.getLogger(FFTest.class);

    @Test
    public void test() {
        int round = 0;
        try {
            while (true) {
                Flux.range(0, 10)
                    .concatWithValues(10, 11, 12, 13)
                    .concatWith(Flux.range(14, 100 - 14))
                    .limitRate(16, 2)
                    .publishOn(Schedulers.boundedElastic(), 16)
                    .blockLast();
                log.info("Round : " + (round++));
            }
        } catch(Exception e) {
            log.error("", e);
        }
    }
}
```

```
13:06:20.429 [main] INFO com.ca.apm.tas.FFTest - Round : 0
......
13:06:20.442 [main] INFO com.ca.apm.tas.FFTest - Round : 38
13:06:20.442 [main] INFO com.ca.apm.tas.FFTest - Round : 39
13:06:20.450 [main] ERROR com.ca.apm.tas.FFTest - 
reactor.core.Exceptions$OverflowException: Queue is full: Reactive Streams source doesn't respect backpressure
	at reactor.core.Exceptions.failWithOverflow(Exceptions.java:234)
	at reactor.core.publisher.FluxPublishOn$PublishOnSubscriber.onNext(FluxPublishOn.java:232)
	at reactor.core.publisher.FluxConcatArray$ConcatArraySubscriber.onNext(FluxConcatArray.java:183)
	at reactor.core.publisher.FluxRange$RangeSubscription.slowPath(FluxRange.java:154)
	at reactor.core.publisher.FluxRange$RangeSubscription.request(FluxRange.java:109)
```

Problem is that reactor.core.publisher.Operators.MultiSubscriptionSubscriber.drainLoop() is double counting missedRequested in case that missedSubscription happens during evaluation (two cycles). Example values:

First iteration in drainLoop:
missedRequested = 12, missedProduced = 0, requested = 2, missedSubscription = null,
requestTarget = log
requestAmount = 12 

Second iteration in drainLoop:
missedRequested = 0, missedProduced = 0, requested = 14, missedSubscription = range  
requestTarget = range
requestAmount = 26

See that missedRequested (12) was counted twice so requestAmount is 26 instead of 14.

### DIFF
--- a/reactor-core/src/main/java/reactor/core/publisher/Operators.java
+++ b/reactor-core/src/main/java/reactor/core/publisher/Operators.java
@@ -2354,8 +2354,12 @@ public abstract class Operators {
 		@Override
 		@Nullable
 		public Object scanUnsafe(Attr key) {
-			if (key == Attr.TERMINATED || key == Attr.CANCELLED)
-				return once == 1;
+			if (key == Attr.TERMINATED) {
+				return once == 1 || once == 2;
+			}
+			if (key == Attr.CANCELLED) {
+				return once == 2;
+			}
 
 			return InnerProducer.super.scanUnsafe(key);
 		}
--- a/reactor-core/src/main/java/reactor/core/publisher/Operators.java
+++ b/reactor-core/src/main/java/reactor/core/publisher/Operators.java
@@ -2204,6 +2204,7 @@ public abstract class Operators {
 		    int missed = 1;
 
 	        long requestAmount = 0L;
+	        long alreadyInRequestAmount = 0L;
 	        Subscription requestTarget = null;
 
 	        for (; ; ) {
@@ -2258,11 +2259,12 @@ public abstract class Operators {
 	                    }
 		                subscription = ms;
 		                if (r != 0L) {
-			                requestAmount = addCap(requestAmount, r);
+			                requestAmount = addCap(requestAmount, r - alreadyInRequestAmount);
 	                        requestTarget = ms;
 	                    }
 	                } else if (mr != 0L && a != null) {
 	                    requestAmount = addCap(requestAmount, mr);
+	                    alreadyInRequestAmount += mr; 
 	                    requestTarget = a;
 	                }
 	            }
--- a/reactor-core/src/main/java/reactor/core/publisher/Operators.java
+++ b/reactor-core/src/main/java/reactor/core/publisher/Operators.java
@@ -1098,6 +1098,21 @@ public abstract class Operators {
 	}
 
 	/**
+	 * Represents a fuseable Subscription that emits a single constant value synchronously
+	 * to a Subscriber or consumer.
+	 *
+	 * @param subscriber the delegate {@link Subscriber} that will be requesting the value
+	 * @param value the single value to be emitted
+	 * @param stepName step name
+	 * @param <T> the value type
+	 * @return a new scalar {@link Subscription}
+	 */
+	public static <T> Subscription scalarSubscription(CoreSubscriber<? super T> subscriber,
+			T value, String stepName){
+		return new ScalarSubscription<>(subscriber, value, stepName);
+	}
+	
+	/**
 	 * Safely gate a {@link Subscriber} by making sure onNext signals are delivered
 	 * sequentially (serialized).
 	 * Serialization uses thread-stealing and a potentially unbounded queue that might
@@ -2310,11 +2325,19 @@ public abstract class Operators {
 		final T value;
 
 		volatile int once;
+		
+		String stepName;
+		
 		ScalarSubscription(CoreSubscriber<? super T> actual, T value) {
-			this.value = Objects.requireNonNull(value, "value");
-			this.actual = Objects.requireNonNull(actual, "actual");
+			this(actual, value, null);
 		}
 
+		ScalarSubscription(CoreSubscriber<? super T> actual, T value, String stepName) {
+			this.value = Objects.requireNonNull(value, "value");
+			this.actual = Objects.requireNonNull(actual, "actual");
+			this.stepName = stepName;
+		}
+		
 		@Override
 		public void cancel() {
 			if (once == 0) {
@@ -2392,7 +2415,7 @@ public abstract class Operators {
 
 		@Override
 		public String stepName() {
-			return "scalarSubscription(" + value + ")";
+			return stepName != null ? stepName : ("scalarSubscription(" + value + ")");
 		}
 
 		@SuppressWarnings("rawtypes")
--- a/reactor-core/src/main/java/reactor/core/publisher/Operators.java
+++ b/reactor-core/src/main/java/reactor/core/publisher/Operators.java
@@ -1099,11 +1099,12 @@ public abstract class Operators {
 
 	/**
 	 * Represents a fuseable Subscription that emits a single constant value synchronously
-	 * to a Subscriber or consumer.
+	 * to a Subscriber or consumer. Also give the subscription a user-defined {@code stepName}
+	 * for the purpose of {@link Scannable#stepName()}.
 	 *
 	 * @param subscriber the delegate {@link Subscriber} that will be requesting the value
 	 * @param value the single value to be emitted
-	 * @param stepName step name
+	 * @param stepName the {@link String} to represent the {@link Subscription} in {@link Scannable#stepName()}
 	 * @param <T> the value type
 	 * @return a new scalar {@link Subscription}
 	 */
@@ -2324,10 +2325,11 @@ public abstract class Operators {
 
 		final T value;
 
+		@Nullable
+		final String stepName;
+
 		volatile int once;
-		
-		String stepName;
-		
+
 		ScalarSubscription(CoreSubscriber<? super T> actual, T value) {
 			this(actual, value, null);
 		}
@@ -2337,7 +2339,7 @@ public abstract class Operators {
 			this.actual = Objects.requireNonNull(actual, "actual");
 			this.stepName = stepName;
 		}
-		
+
 		@Override
 		public void cancel() {
 			if (once == 0) {
--- a/reactor-core/src/main/java/reactor/core/publisher/FluxJust.java
+++ b/reactor-core/src/main/java/reactor/core/publisher/FluxJust.java
@@ -17,11 +17,9 @@
 package reactor.core.publisher;
 
 import java.util.Objects;
-import java.util.concurrent.atomic.AtomicIntegerFieldUpdater;
 
 import reactor.core.CoreSubscriber;
 import reactor.core.Fuseable;
-import reactor.util.annotation.Nullable;
 
 /**
  * A Stream that emits only one value and then complete.
@@ -67,7 +65,7 @@ final class FluxJust<T> extends Flux<T>
 
 	@Override
 	public void subscribe(final CoreSubscriber<? super T> actual) {
-		actual.onSubscribe(new WeakScalarSubscription<>(value, actual));
+		actual.onSubscribe(Operators.scalarSubscription(actual, value));
 	}
 
 	@Override
@@ -76,82 +74,4 @@ final class FluxJust<T> extends Flux<T>
 		return null;
 	}
 
-	static final class WeakScalarSubscription<T> implements QueueSubscription<T>,
-	                                                        InnerProducer<T>{
-
-		volatile int terminated;
-		@SuppressWarnings("rawtypes")
-		static final AtomicIntegerFieldUpdater<WeakScalarSubscription> TERMINATED =
-				AtomicIntegerFieldUpdater.newUpdater(WeakScalarSubscription.class, "terminated");
-		
-		final T                     value;
-		final CoreSubscriber<? super T> actual;
-
-		WeakScalarSubscription(@Nullable T value, CoreSubscriber<? super T> actual) {
-			this.value = value;
-			this.actual = actual;
-		}
-
-		@Override
-		public void request(long elements) {
-			if (!TERMINATED.compareAndSet(this, 0, 1)) {
-				return;
-			}
-
-			if (value != null) {
-				actual.onNext(value);
-			}
-			actual.onComplete();
-		}
-
-		@Override
-		public void cancel() {
-			TERMINATED.set(this, 1);
-		}
-
-		@Override
-		public int requestFusion(int requestedMode) {
-			if ((requestedMode & Fuseable.SYNC) != 0) {
-				return Fuseable.SYNC;
-			}
-			return 0;
-		}
-
-		@Override
-		@Nullable
-		public T poll() {
-			if (TERMINATED.compareAndSet(this, 0, 1)) {
-				return value;
-			}
-			return null;
-		}
-
-		@Override
-		public boolean isEmpty() {
-			return TERMINATED.get(this) > 0;
-		}
-
-		@Override
-		public int size() {
-			return isEmpty() ? 0 : 1;
-		}
-
-		@Override
-		public void clear() {
-			TERMINATED.set(this, 1);
-		}
-
-		@Override
-		public CoreSubscriber<? super T> actual() {
-			return actual;
-		}
-
-		@Override
-		@Nullable
-		public Object scanUnsafe(Attr key) {
-			if (key == Attr.TERMINATED || key == Attr.CANCELLED) return TERMINATED.get(this) == 1;
-
-			return InnerProducer.super.scanUnsafe(key);
-		}
-	}
 }
--- a/reactor-core/src/main/java/reactor/core/publisher/FluxJust.java
+++ b/reactor-core/src/main/java/reactor/core/publisher/FluxJust.java
@@ -17,8 +17,8 @@
 package reactor.core.publisher;
 
 import java.util.Objects;
+import java.util.concurrent.atomic.AtomicIntegerFieldUpdater;
 
-import org.reactivestreams.Subscriber;
 import reactor.core.CoreSubscriber;
 import reactor.core.Fuseable;
 import reactor.util.annotation.Nullable;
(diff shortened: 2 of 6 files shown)


koldat:
Just pasting another test that this PR fixes (#2567) to be complete. It is about emitting one value twice from Flux.just:
```java
    private static Scheduler reader = Schedulers.newBoundedElastic(32, Integer.MAX_VALUE, "tas-reader");
    
        int round = 0;
        while (true) {
        	Mono.just(0)
        	.flatMapMany(i -> {
        		return Flux.range(0,10)
	                .publishOn(reader)
	                .concatWithValues(10)
	                .concatWithValues(11)
	                .concatWithValues(12)
	                .concatWithValues(13)
	                .concatWithValues(14)
	                .concatWithValues(15)
	                .concatWithValues(16)
	                .concatWithValues(17)
	                .concatWith(Flux.range(18, 100-18));
        	})
            //.log()
            .publishOn(reader, 16)
            //.log()
            .subscribeOn(reader)
            .blockLast();

        	System.out.println("Round : " + (round++));
        }
    }
```

```
28: [tas-reader-2] - reactor.Flux.PublishOn.81446 | onNext(11)
29: [tas-reader-1] - reactor.Flux.MonoFlatMapMany.81445 onNext(13)
30: [tas-reader-2] - reactor.Flux.MonoFlatMapMany.81445 request(12)
31: [tas-reader-1] - reactor.Flux.MonoFlatMapMany.81445 onNext(14) <-- Here is duplicate
32: [tas-reader-2] - reactor.Flux.MonoFlatMapMany.81445 onNext(14) <-- Here is duplicate
33: [tas-reader-1] - reactor.Flux.MonoFlatMapMany.81445 onNext(15)
34: [tas-reader-1] - reactor.Flux.MonoFlatMapMany.81445 onNext(16)
```

OlegDokuka:
@koldat looking at the code now. Thinking if we can do it better. I'm not sure all the volatiles you added should be there

koldat:
@OlegDokuka - all volatiles I added needs to be there as there is no "synchronized" that make a barrier. Threads has no way of finding out that memory was changed during race on places I updated. We only protect area using WIP. We should be better safe then sorry. These took me 3 days until I found correct place. It is very hard to find as it is extremely rare and heavy multithread. But if you want we can take it one by one. But there is always a possibility that two threads (one from up and one from down) will be going to these variables area as we have onComplete (from up) and request (from down or both as it transfer to onNext).

OlegDokuka:
@koldat it is a little tricky. There are guarantees provided by the reactive streams spec which already provides memory barriers. Therefore, I said that the volatiles you have added are redundant.

koldat:
@OlegDokuka I would like to read the source as what Java barrier guarantee is all changes inside synchronized over same monitor. This would hold the lock. But maybe reactor does that different way which I would like to see as I did not found any memory barriers around calls. Can you point me to one example in sources?

OlegDokuka:
@koldat there is a patter so-called "work in progress" (please have a look here and [here](https://github.com/reactor/reactor-core/pull/2576/files#diff-c4b3441dd7a40c384b5dc7ffdc0c7f41bf18f239fe04db5196d68ab87ff7d321R2250)), which uses the volatile field to increment and decrement a counter. As you may notice, there is a check before entering the critical section which checks that nobody is working now, and if somebody works, the counter will be just incremented (according to the Java Memory Model, all the non-volatile writes that [happen before](https://docs.oracle.com/javase/specs/jls/se8/html/jls-17.html#jls-17.4.5) the write memory barrier must be available to the reader. In the same letter, once the work is done, wip field is decremented, chance all the action did within the critical section MUST be available to the next worker which reads on the same volatile field. That said. if we have WIP guarding at any place in our code, we have guaranteed that this section will not have any racing at all. 

Apart, feel free to check places in [the spec](https://github.com/reactive-streams/reactive-streams-jvm/blob/v1.0.3/README.md#glossary) where words `serial` and `thread-safe` are mentioned. The spec says that access to subscription MUST be serial as well as produce MUST call the subscriber in a serial fashion. Though we still have a possibility of racing onNext with subscription#request, we have in may places the `Work in progress` guard which **should** (I say should since we may have a bug as it is in MultisubscriptionSubscriber now) prevent racing in such scenarios

koldat:
You probably refer to reactor.core.Fuseable.THREAD_BARRIER -  	/**
	 * Indicates that the queue will be drained from another thread
	 * thus any queue-exit computation may be invalid at that point.*/

It is just a flag for fusion. In this example two threads are racing there thus needs to follow memory behavior. For example:
use PRODUCED.incrementAndGet(this); instead of produced++;

produced variable is read under WIP, but onNext does not have WIP so two thread can modify the field.

And even if onNext has WIP no one guarantee produced field will be consistent with changes as it is not volatile. And even if it would be volatile one need to use atomic operation as this would break it:
```
					long c = produced;
					if (c != 0L) {
						produced = 0L; <-- Other thread can increment right before this
						produced(c);
					}
```

correct is
```
					long c = PRODUCED.getAndSet(this, 0);
```

etc. Similar is in MultiSubscriptionSubscriber. subscription and requested fields are protected by WIP. But if two threads will go for change (will get WIP access) microseconds in a row then second thread can have old result. This probability is very small, because caches will be most likelly flushed, but I think possibility is there as queue can be drained by two different threads.

OlegDokuka:
@koldat if you refer to FluxConcatArray, then onNext can not race with onXXX method, hence modifying non-volatile filed is safe.

OlegDokuka:
Also, when an operator was fussed, we are under another (internal agreement) guaranteed that says that if the operator was fused in ASYNC or SYNC fashion, then the producer will not be sending any **real** signals over onNext, but rather will be signaling that a queue has values to be drained under the WIP guard. Hence, it will never happen (unless a bug) that onNext modifying a non-volatile field simultaneously with logic in the drain loop

koldat:
@OlegDokuka I agree that FluxConcatArray produced volatile you are right, because most likely switched thread will not have cached such a memory. 

But I do not agree with your view on JVM memory barrier as it has nothing with WIP pattern which is based on atomics and is protecting you only in a way that a thread has an access to variables defined by contract. It does not guarantee you contents of variable that is not volatile in any way. Only the one that are volatile has guaranteed value under WIP.

But if you guarantee it will never happen I can live with that. I would just not be that sure, because threading defects are mostly impossible to find even with such a simple and strict unit test. If it would be a money transaction I would not be happy to count it twice even if everyone will say it cannot happen.

So should I revert volatile changes?

OlegDokuka:
If tests still passes without volatiles - then yes, please

OlegDokuka:
I mean, it does not for me. But I'm still trying to figure out what goes wrong there

koldat:
I have reverted volatile changes. Only fix for double count and Flux.,just is there. Tests pass.

OlegDokuka:
@koldat can you please add the [mentioned](https://github.com/reactor/reactor-core/pull/2576#issuecomment-769021896) test to make sure changes works as expected

simonbasle (changes requested):
is there no way to reproduce the issue in a unit test or a JCStress test ? even a test that was proven to fail locally when using eg. the `repeat until failure` feature of IntelliJ, but now passes? I know, I know, heisenbugs, the lack of failure doesn't prove it can't happen, but it would still be useful to have _something_, especially to demonstrate what you think is the cause and example of a situation that can lead to the issue?

as a side note, and smaller nitpick, we could use this opportunity to get rid of the weird `terminado` wording (which I think was an early inside "easter egg")

@OlegDokuka I'll defer to you for the rest of the review

koldat:
I changed terminado -> terminated. Let me know if you have better name.

I am just strugling where to put the test. I can put there 1000 cycles as it was failing mostly before 200 cycles. Then the test would not need to take that much time.  Are there already any tests like this (stress ones) so I can put it to same place? Any pointer?

simonbasle:
mmh there are stress tests like this inside ordinary test classes per operator (eg in this case `FluxJustTest.java` I guess). In `master` there are also separate stress tests using JCStress (but that has a learning curve, like JMH), but this targets 3.3.x anyway.

koldat:
I have added three tests that runs around 500ms. All of them fails randomly (every 3-5 runs) before fix and do not fail after fix.

simonbasle (approved):
LGTM. @OlegDokuka ?

OlegDokuka (changes requested):
Lets not edit WeakScalarSubscription for now but rather use [ScalarSubscription(https://github.com/reactor/reactor-core/blob/master/reactor-core/src/main/java/reactor/core/publisher/Operators.java#L1095) instead of WeakScalarSubscription there -> https://github.com/reactor/reactor-core/pull/2576/files#diff-1bf720442d34f38fd4c5dfe1348c9c96df93f6d5a0e8db9088f5b2ec59abf68cR70

koldat:
I do not understand. FluxJust is having WeakScalarSubscription inner static class. I cannot use different type. Then field update will not work. You want me to change just implementation? That is not point of the fix.

koldat:
Switch Flux just to use scalar subscription. I had to fix its implementation as well to pass test in FluxJust:

```java
	@Test
	public void scanSubscription() {
		CoreSubscriber<Integer> actual = new LambdaSubscriber<>(null, e -> {}, null, sub -> sub.request(100));
		ScalarSubscription<Integer> test = new ScalarSubscription<>(actual, 1);

		assertThat(test.scan(Scannable.Attr.ACTUAL)).isSameAs(actual);
		assertThat(test.scan(Scannable.Attr.TERMINATED)).isFalse();
		assertThat(test.scan(Scannable.Attr.CANCELLED)).isFalse();
		test.cancel();
		assertThat(test.scan(Scannable.Attr.CANCELLED)).isTrue();
		assertThat(test.scan(Scannable.Attr.TERMINATED)).isTrue();
	}
```

Please take look

simonbasle (approved):

koldat:
There is contradicted condition in reactor.core.publisher.OperatorsTest.scanScalarSubscription()

```java
	@Test
	public void scanScalarSubscription() {
		CoreSubscriber<Integer> actual = new LambdaSubscriber<>(null, null, null, null);
		ScalarSubscription<Integer> test = new ScalarSubscription<>(actual, 5);

		assertThat(test.scan(Scannable.Attr.ACTUAL)).isSameAs(actual);

		assertThat(test.scan(Scannable.Attr.TERMINATED)).isFalse();
		assertThat(test.scan(Scannable.Attr.CANCELLED)).isFalse();
		test.poll();
		assertThat(test.scan(Scannable.Attr.TERMINATED)).isTrue();
		assertThat(test.scan(Scannable.Attr.CANCELLED)).isTrue(); <--- HERE
	}
```

I would vote for putting there isFalse. But not sure. Which test should I fix?

simonbasle:
Mmh yeah I was pondering about that last minute change to `scan` but I tend to agree that there must have been a bug here, given the implementation of `cancel()` setting the `ONCE` to `2`. You can change that line to `isFalse()` I think. It might be worth it to  then add `test.cancel()` and check that `scan(Attr.CANCELLED)` becomes `true`, wdyt?

simonbasle:
ah damn, there's another issue with use of generic `Subscription`: 
image

koldat:
Yes I just seen that as well. I can change it to make test passing. Is it ok?

OlegDokuka:
@koldat go ahead and do that please so we can finally merge that PR 🎉

simonbasle:
> Yes I just seen that as well. I can change it to make test passing. Is it ok?

yes as a stopgap measure, but we'll need to follow up on that, `just` is more useful as a step name than `scalarSubscription(foo)` I think. hopefully this doesn't break anybody, as `stepName` is intended for human consumption 🤞

simonbasle:
created #2611 for the follow up

simonbasle (approved):

OlegDokuka (approved):
LGTM

koldat:
Let me add that attribute to not break some user tests. Give me a minute.

simonbasle:
> Let me add that attribute to not break some user tests. Give me a minute.

As you wish. It can be a `String` field prepopulated with the current `stepName` with an additional constructor and corresponding `Operators.scalarSubscription` method that take a precomputed `stepName` instead. Here is the sketch of the additional `scalarSubscription` I just came up with:

```java
	/**
	 * Represents a fuseable Subscription that emits a single constant value synchronously
	 * to a Subscriber or consumer. Also give the subscription a user-defined {@code stepName}
	 * for the purpose of {@link Scannable#stepName()}.
	 *
	 * @param subscriber the delegate {@link Subscriber} that will be requesting the value
	 * @param value the single value to be emitted
	 * @param stepName the {@link String} to represent the {@link Subscription} in {@link Scannable#stepName()}
	 * @param <T> the value type
	 * @return a new scalar {@link Subscription}
	 */
	public static <T> Subscription scalarSubscription(CoreSubscriber<? super T> subscriber,
			T value, String stepName) {
		return new ScalarSubscription<>(subscriber, value, stepName);
```

koldat:
Pushed, please check

reactor-core/src/main/java/reactor/core/publisher/Operators.java
```diff
@@ -1097,6 +1097,21 @@ public static void reportSubscriptionSet() {
 		return new ScalarSubscription<>(subscriber, value);
 	}
 
+	/**
```
simonbasle, line 1100: can you please reuse the wording I suggested in my earlier comment? 🙇

reactor-core/src/main/java/reactor/core/publisher/Operators.java
```diff
@@ -2310,11 +2325,19 @@ final void drainLoop() {
 		final T value;
 
 		volatile int once;
+		
+		String stepName;
```
simonbasle, line 2329: this field can be final (and as such should be moved above the `volatile` one)

reactor-core/src/main/java/reactor/core/publisher/Operators.java
```diff
@@ -2310,11 +2325,19 @@ final void drainLoop() {
...
+			this(actual, value, null);
+		}
+
+		ScalarSubscription(CoreSubscriber<? super T> actual, T value, String stepName) {
 			this.value = Objects.requireNonNull(value, "value");
 			this.actual = Objects.requireNonNull(actual, "actual");
+			this.stepName = stepName;
 		}
-
+
```
simonbasle, line 2340: please trim whitespaces 😉

reactor-core/src/main/java/reactor/core/publisher/Operators.java
```diff
@@ -2310,11 +2325,19 @@ final void drainLoop() {
...
+		String stepName;
+		
 		ScalarSubscription(CoreSubscriber<? super T> actual, T value) {
+			this(actual, value, null);
+		}
+
+		ScalarSubscription(CoreSubscriber<? super T> actual, T value, String stepName) {
 			this.value = Objects.requireNonNull(value, "value");
 			this.actual = Objects.requireNonNull(actual, "actual");
+			this.stepName = stepName;
```
simonbasle, line 2338: you can `Objects.requireNonNull` here, I prefer the last iteration where the `stepName` was assemble in the constructor.

reactor-core/src/main/java/reactor/core/publisher/Operators.java
```diff
@@ -2392,7 +2415,7 @@ public int size() {
 
 		@Override
 		public String stepName() {
-			return "scalarSubscription(" + value + ")";
+			return stepName != null ? stepName : ("scalarSubscription(" + value + ")");
```
simonbasle, line 2418: see above comment, let's compute `stepName` upfront and make it final

simonbasle (changes requested):

reactor-core/src/main/java/reactor/core/publisher/Operators.java
koldat: I did, but that would affect currect performance potentially as every scalar will calculate string used only in tests.

reactor-core/src/main/java/reactor/core/publisher/Operators.java
koldat: See my second comment. Unnecessary strings in memory. in 99% not used in runtime. I know it is not nice, but runtime is more important I think.

reactor-core/src/main/java/reactor/core/publisher/Operators.java
simonbasle: mmh fair point, it might be slightly more acceptable to slow `stepName` than the constructor... keep in mind `stepName()` isn't necessarily only for tests, it's just that we don't have other prod code that depends on it, but it is very much for user consumption in both their prod and test code.

ok then, let's compute lazily here.

I think the field can still be final though, right?

reactor-core/src/main/java/reactor/core/publisher/Operators.java
simonbasle: still missing the little sentence in the body (line 1102)

reactor-core/src/main/java/reactor/core/publisher/Operators.java
```diff
@@ -2307,10 +2324,18 @@ final void drainLoop() {
 
 		final T value;
 
+		final String stepName;
```
simonbasle, line 2328: this should probably be marked with the reactor `@Nullable` annotation on the line above it

simonbasle (changes requested):
last couple of small adjustments, I promise 😆 
this is looking good, thanks for your effort 👍

reactor-core/src/main/java/reactor/core/publisher/Operators.java
koldat: done

reactor-core/src/main/java/reactor/core/publisher/Operators.java
koldat: done

reactor-core/src/main/java/reactor/core/publisher/Operators.java
koldat: done

reactor-core/src/main/java/reactor/core/publisher/Operators.java
koldat: done

simonbasle (approved):
thanks @koldat ! looks like I can scrap the follow up issue 👍 🙇 🎁

koldat:
yay! :) Thank you guys and sorry for so many turn arounds.

simonbasle:
Forward-merge commit into 3.4.3 is d22f3524

PR #2617, opened 2021-02-16
Fix wrong param name in Sinks.Many#limit(Duration) javadoc

v-terret:


simonbasle (approved):
Good catch, thanks @v-terret

PR #2626, opened 2021-03-02
Improves fuse-enabled operator

OlegDokuka:
This PR is intended to consolide fusion-enabled operators and their relationships with each other if fusion is enabled and the downstream gets access to the queue of the upstream.

The following is simple rules I built up to ensure we are not leaking values, is not getting into dead-lock if multiple threads get access to ScQueue draining/cleaning:

If `ASYNC` fusion is established between upstream and downstream, the downstream get full access to `Queue` draining, and only downstream MUST anyhow interact with that queue until the interaction is considered terminated. 
To terminate interaction and notify the upstream there will no more interaction with `Queue`, the downstream MUST call `queue#clear` (this call must be done in all the terminal scenarios includes COMPLETE / ERROR / CANCEL).
Once, the backfused upstream received `Queue#clear` call, it MUST clear and discard all the value from the queue and set terminate state which indicate that now the upstream is ready to manage `Queue` on its own and no more interactions with that queue are expected from the downstream

### DIFF
--- a/reactor-core/src/main/java/reactor/core/publisher/UnicastProcessor.java
+++ b/reactor-core/src/main/java/reactor/core/publisher/UnicastProcessor.java
@@ -17,6 +17,7 @@ package reactor.core.publisher;
 
 import java.util.Objects;
 import java.util.Queue;
+import java.util.concurrent.CancellationException;
 import java.util.concurrent.atomic.AtomicIntegerFieldUpdater;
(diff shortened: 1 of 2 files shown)


chemicL:
@OlegDokuka please review and potentially close if this is outdated and should be re-introduced as a separate effort if needed.

PR #2645, opened 2021-03-11
improves `MonoProducerBase` and removes `Fusion` from MonoSingle

OlegDokuka:
Some improvements for MonoProducerBase:

1. Ensure infinite loop exits in case of request appears earlier than the value
2. Ensure doOnXXX called exactly once (usually that is what is needed in subclasses) 
3. Removes Fusion interface from MonoSingle to avoid ClassCastException on downstream receiving non `QueueSubscription` type of `Subscription`


ericbottard:
You could have deleted `MonoInnerProducerBaseTest`

PR #2701, opened 2021-05-06
Implement FluxPrefetch operator to control prefetch request time

Tariod:
**Proposal:**
Add prefetch mode (LAZY, EAGER) to reduce the number of discarded values when using resource-constrained data sources with an additional operator.

**Price of changes:**
- Possible overhead due to the use of an additional operator.
- Changing the behavior of some operators due to moving the logic of the prefetch operation into a separate operator.

### DIFF
--- a/reactor-core/src/main/java/reactor/core/publisher/Flux.java
+++ b/reactor-core/src/main/java/reactor/core/publisher/Flux.java
@@ -6015,7 +6015,7 @@ public abstract class Flux<T> implements CorePublisher<T> {
 	 * @return a {@link Flux} limiting downstream's backpressure and customizing the
 	 * replenishment request amount
 	 * @see #publishOn(Scheduler)
-	 * @see #prefetch(int)
+	 * @see #prefetch(int, int)
 	 * @see #limitRequest(long)
 	 */
 	public final Flux<T> limitRate(int highTide, int lowTide) {
@@ -7101,14 +7101,8 @@ public abstract class Flux<T> implements CorePublisher<T> {
 	}
 
 	public final Flux<T> prefetch(int prefetch, int lowTide, boolean lazyMode) {
-		FluxPrefetch.PrefetchMode mode = lazyMode ? FluxPrefetch.PrefetchMode.LAZY :
-				FluxPrefetch.PrefetchMode.EAGER;
-
-		return onAssembly(new FluxPrefetch<>(this,
-				prefetch,
-				lowTide,
-				Queues.get(prefetch),
-				mode));
+		FluxPrefetch.PrefetchMode mode = lazyMode ? FluxPrefetch.PrefetchMode.LAZY : FluxPrefetch.PrefetchMode.EAGER;
+		return onAssembly(new FluxPrefetch<>(this, prefetch, lowTide, Queues.get(prefetch), mode));
 	}
 
 	/**
--- a/reactor-core/src/main/java/reactor/core/publisher/Flux.java
+++ b/reactor-core/src/main/java/reactor/core/publisher/Flux.java
@@ -7076,6 +7076,25 @@ public abstract class Flux<T> implements CorePublisher<T> {
 				Queues.get(prefetch));
 	}
 
+	public final Flux<T> prefetch(int prefetch) {
+		return prefetch(prefetch, FluxPrefetch.RequestMode.EAGER);
+	}
+
+	public final Flux<T> prefetch(int prefetch, FluxPrefetch.RequestMode requestMode) {
+		return prefetch(prefetch, prefetch, requestMode);
+	}
+
+	public final Flux<T> prefetch(int prefetch,
+			int lowTide,
+			FluxPrefetch.RequestMode requestMode) {
+//    TODO: if Callable - FluxPrefetchValue
+		return onAssembly(new FluxPrefetch<>(this,
+				prefetch,
+				lowTide,
+				Queues.get(prefetch),
+				requestMode));
+	}
+
 	/**
 	 * Prepare a {@link ConnectableFlux} which shares this {@link Flux} sequence and
 	 * dispatches values to subscribers in a backpressure-aware manner. Prefetch will
--- a/reactor-core/src/main/java/reactor/core/publisher/Flux.java
+++ b/reactor-core/src/main/java/reactor/core/publisher/Flux.java
@@ -7087,7 +7087,7 @@ public abstract class Flux<T> implements CorePublisher<T> {
 	public final Flux<T> prefetch(int prefetch,
 			int lowTide,
 			FluxPrefetch.RequestMode requestMode) {
-//    TODO: if Callable - FluxPrefetchValue
+//      TODO: if Callable -> FluxPrefetchValue
 		return onAssembly(new FluxPrefetch<>(this,
 				prefetch,
 				lowTide,
--- a/reactor-core/src/main/java/reactor/core/publisher/Flux.java
+++ b/reactor-core/src/main/java/reactor/core/publisher/Flux.java
@@ -7088,21 +7088,21 @@ public abstract class Flux<T> implements CorePublisher<T> {
 		return prefetch(Queues.SMALL_BUFFER_SIZE, lazyMode);
 	}
 
-	public final Flux<T> prefetch(int prefetch) {
-		return prefetch(prefetch, false);
+	public final Flux<T> prefetch(int highTide) {
+		return prefetch(highTide, false);
 	}
 
-	public final Flux<T> prefetch(int prefetch, int lowTide) {
-		return prefetch(prefetch, lowTide, false);
+	public final Flux<T> prefetch(int highTide, int lowTide) {
+		return prefetch(highTide, lowTide, false);
 	}
 
-	public final Flux<T> prefetch(int prefetch, boolean lazyMode) {
-		return prefetch(prefetch, prefetch, lazyMode);
+	public final Flux<T> prefetch(int highTide, boolean lazyMode) {
+		return prefetch(highTide, highTide, lazyMode);
 	}
 
-	public final Flux<T> prefetch(int prefetch, int lowTide, boolean lazyMode) {
+	public final Flux<T> prefetch(int highTide, int lowTide, boolean lazyMode) {
 		FluxPrefetch.PrefetchMode mode = lazyMode ? FluxPrefetch.PrefetchMode.LAZY : FluxPrefetch.PrefetchMode.EAGER;
-		return onAssembly(new FluxPrefetch<>(this, prefetch, lowTide, Queues.get(prefetch), mode));
+		return onAssembly(new FluxPrefetch<>(this, highTide, lowTide, Queues.get(highTide), mode));
 	}
 
 	/**
--- a/reactor-core/src/main/java/reactor/core/publisher/Flux.java
+++ b/reactor-core/src/main/java/reactor/core/publisher/Flux.java
@@ -5970,12 +5970,13 @@ public abstract class Flux<T> implements CorePublisher<T> {
 	 * @param prefetchRate the limit to apply to downstream's backpressure
 	 *
 	 * @return a {@link Flux} limiting downstream's backpressure
-	 * @see #publishOn(Scheduler, int)
+	 * @see #publishOn(Scheduler)
+	 * @see #prefetch(int)
 	 * @see #limitRequest(long)
 	 */
 	public final Flux<T> limitRate(int prefetchRate) {
-		Flux<T> prefetch = onAssembly(this.prefetch(prefetchRate));
-		return onAssembly(prefetch.publishOn(Schedulers.immediate(), true));
+		Flux<T> prefetch = this.prefetch(prefetchRate);
+		return prefetch.publishOn(Schedulers.immediate(), true);
 	}
 
 	/**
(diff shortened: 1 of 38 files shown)


simonbasle:
hi @Tariod ! Thanks for the PR (I knew it would come one day 😉). This one is quite huge and won't probably make sense until 3.5+, so I'm turning it into a draft for now.

Note that this PR is extremely difficult to review as is, even when not considering the sheer size of the change, due to a lot of noise around formatting and reordering of methods... (see the diff of `FluxPublishOn` for example). That's a big time sink, but I'd invest into changing that to make the PR more palatable...

Also, I wonder if it might make sense to have the conditional version of the `FluxPrefetch` subscriber extend the non-conditional one in order to reduce the size of the class (1400 LoC !). I think that goes slightly against the usual pattern of code duplication we have in Reactor, but that is an acceptable tradeoff for such a huge class (and we can always split at the end if we demonstrate it is better).

Tariod:
Hi @simonbasle! I will try to make changes soon to simplify the review and provide benchmarks for different versions as representative as possible it can be on my laptop.

Tariod:
Hi, @simonbasle! I am sorry for not writing for so long. I did clean some noise to make the review easier: methods reordering, variables naming, and so on.

simonbasle:
@Tariod no problem, thanks for the update. I see that two tests that are likely related now fail on the latest commit (publishOn tests), so it will need some investigation.

there is also a copyright header update, but that can be easily fixed by running `./gradlew :spotlessApply`

on a side note, I think this whole effort would be most relevant if targeted for the next big version of Reactor. We're currently thinking about making that next version Reactor-Core 4.0.x, which would give us more leeway to do deeper changes like this one.

reactor-core/src/main/java/reactor/core/publisher/Flux.java
```diff
@@ -6012,11 +6014,13 @@ public int getPrefetch() {
...
 	 * replenishment request amount
-	 * @see #publishOn(Scheduler, int)
+	 * @see #publishOn(Scheduler)
+	 * @see #prefetch(int, int)
 	 * @see #limitRequest(long)
 	 */
 	public final Flux<T> limitRate(int highTide, int lowTide) {
-		return onAssembly(this.publishOn(Schedulers.immediate(), true, highTide, lowTide));
+		Flux<T> prefetch = this.prefetch(highTide, lowTide);
+		return prefetch.publishOn(Schedulers.immediate(), true);
```
simonbasle, line 6023: I wonder if the `publishOn` is needed at all now that the prefetching logic has been moved to the `prefetch` operator. the original idea behind this operator was to re-shape requests according to lowTide/highTide and introduce limit replenishing, and it was using `publishOn` as the best approximation of that (hence the `Schedulers.immediate`).

reactor-core/src/main/java/reactor/core/publisher/Flux.java
```diff
@@ -7076,6 +7080,31 @@ public int getPrefetch() {
...
+
+	public final Flux<T> prefetch(boolean lazyMode) {
+		return prefetch(Queues.SMALL_BUFFER_SIZE, lazyMode);
+	}
+
+	public final Flux<T> prefetch(int prefetch) {
+		return prefetch(prefetch, false);
+	}
+
+	public final Flux<T> prefetch(int prefetch, int lowTide) {
```
simonbasle, line 7095: nit: for consistency with the `lowTide` parameter, I would rename `prefetch` to `highTide` here

reactor-core/src/main/java/reactor/core/publisher/Flux.java
```diff
@@ -7076,6 +7080,31 @@ public int getPrefetch() {
...
+		return prefetch(prefetch, lowTide, false);
+	}
+
+	public final Flux<T> prefetch(int prefetch, boolean lazyMode) {
+		return prefetch(prefetch, prefetch, lazyMode);
+	}
+
+	public final Flux<T> prefetch(int prefetch, int lowTide, boolean lazyMode) {
```
simonbasle, line 7103: nit: for consistency with the `lowTide` parameter, I would rename `prefetch` to `highTide` here

reactor-core/src/main/java/reactor/core/publisher/ParallelRunOn.java
```diff
@@ -79,18 +78,32 @@ public void subscribe(CoreSubscriber<? super T>[] subscribers) {
 			Worker w = scheduler.createWorker();
 
 			if (conditional) {
-				parents[i] = new FluxPublishOn.PublishOnConditionalSubscriber<>(
-						(Fuseable.ConditionalSubscriber<T>)subscribers[i],
-						scheduler, w, true,
-						prefetch, prefetch, queueSupplier);
+				parents[i] =
+						new FluxPrefetch.PrefetchSubscriber<>(
```
simonbasle, line 82: we can keep this, but IIRC the `tryOnNext` path is triggered by sources exclusively, so since `PrefetchSubscriber` isn't conditional it will never be reachable.

reactor-core/src/test/java/reactor/core/publisher/FluxPublishOnTest.java
```diff
@@ -187,14 +181,16 @@ public static void after() {
 	public void failPrefetch() {
 		assertThatExceptionOfType(IllegalArgumentException.class).isThrownBy(() -> {
 			Flux.range(1, 10)
-					.publishOn(Schedulers.immediate(), -1);
+			    .prefetch(-1)
```
simonbasle, line 184: this one should be moved to `FluxPrefetchTest.java` (and shouldn't need the `publishOn`)

reactor-core/src/main/java/reactor/core/publisher/FluxPublishOn.java
```diff
@@ -501,6 +450,52 @@ void runBackfused() {
...
 		}
 
+		void runOneByOne() {
+			int missed = 1;
+
+			final Subscriber<? super T> a = actual;
+
+			long e = produced;
+			for (; ; ) {
+				long r = this.requested;
```
simonbasle, line 460: shouldn't this be `this.requested & Long.MAX_VALUE` ?

reactor-core/src/main/java/reactor/core/publisher/FluxPublishOn.java
```diff
@@ -582,14 +572,12 @@ else if (empty) {
 		@Override
 		@Nullable
 		public Object scanUnsafe(Attr key) {
-			if (key == Attr.REQUESTED_FROM_DOWNSTREAM ) return requested;
-			if (key == Attr.PARENT ) return s;
+			if (key == Attr.REQUESTED_FROM_DOWNSTREAM) return requested;
```
simonbasle, line 575: shouldn't this be `requested & Long.MAX_VALUE` ?

reactor-core/src/main/java/reactor/core/publisher/FluxPublishOn.java
```diff
@@ -582,14 +572,12 @@ else if (empty) {
...
+			if (key == Attr.PARENT) return s;
 			if (key == Attr.CANCELLED) return cancelled;
 			if (key == Attr.TERMINATED) return done;
-			if (key == Attr.BUFFERED) return queue != null ? queue.size() : 0;
```
simonbasle, line 589: should we consider `BUFFERED == 1` when there is something stored in `VALUE` ?

reactor-core/src/main/java/reactor/core/publisher/FluxPublishOn.java
```diff
@@ -893,12 +874,10 @@ void runSync() {
...
 
 			final ConditionalSubscriber<? super T> a = actual;
-			final Queue<T> q = queue;
+			final Queue<T> q = qs;
 
 			long e = produced;
-
 			for (; ; ) {
-
 				long r = requested;
```
simonbasle, line 881: shouldn't this be `requested & Long.MAX_VALUE` ?

reactor-core/src/main/java/reactor/core/publisher/FluxPublishOn.java
```diff
@@ -1055,6 +1030,55 @@ void runBackfused() {
...
 		}
 
+		void runOneByOne() {
+			int missed = 1;
+
+			final ConditionalSubscriber<? super T> a = actual;
+
+			long e = produced;
+			for (; ; ) {
+				long r = requested;
```
simonbasle, line 1040: shouldn't this be `requested & Long.MAX_VALUE` ?

reactor-core/src/main/java/reactor/core/publisher/FluxPublishOn.java
```diff
@@ -1080,10 +1107,8 @@ public Object scanUnsafe(Attr key) {
 			if (key == Attr.PARENT) return s;
 			if (key == Attr.CANCELLED) return cancelled;
 			if (key == Attr.TERMINATED) return done;
-			if (key == Attr.BUFFERED) return queue != null ? queue.size() : 0;
```
simonbasle, line 1083: same as above, this could return `1` if the VALUE is not holding null

reactor-core/src/main/java/reactor/core/publisher/FluxPublishOn.java
```diff
@@ -1080,10 +1107,8 @@ public Object scanUnsafe(Attr key) {
 			if (key == Attr.PARENT) return s;
```
simonbasle, line 1107: above: `REQUESTED_FROM_DOWNSTREAM` should return `this.requested & Long.MAX_VALUE`, not `requested`, right?

reactor-core/src/main/java/reactor/core/publisher/FluxPublishOn.java
```diff
@@ -333,23 +304,19 @@ void runSync() {
...
 
 			final Subscriber<? super T> a = actual;
-			final Queue<T> q = queue;
+			final Queue<T> q = qs;
 
 			long e = produced;
-
 			for (; ; ) {
-
 				long r = requested;
```
simonbasle, line 311: shouldn't this be `requested & Long.MAX_VALUE` ?

simonbasle (changes requested):
this is looking better and better 😄 

found a few issues in `FluxPublishOn` notably, and a few other comments.

I'm not getting into javadoc for the moment, although public methods at least will need to be heavily documented 📖 

also, I have yet to review the main class, `FluxPrefetch` (I have reviewed everything else)

reactor-core/src/main/java/reactor/core/publisher/Flux.java
Tariod: I just revert back to the previous behavior after removing the prefetches

reactor-core/src/main/java/reactor/core/publisher/Flux.java
Tariod: It makes sense. This will help avoid duplicate names.

reactor-core/src/main/java/reactor/core/publisher/FluxPublishOn.java
Tariod: Yes, it needs to cast requested here if the drain is called before the request

reactor-core/src/main/java/reactor/core/publisher/FluxPublishOn.java
Tariod: Yes, it needs to cast requested here if the drain is called before the request

reactor-core/src/main/java/reactor/core/publisher/Flux.java
Tariod: It makes sense. This will help avoid duplicate names.

reactor-core/src/main/java/reactor/core/publisher/ParallelRunOn.java
Tariod: PublishOn interacts with the downstream by the ConditionalSubscriber contract. Also, PublishOn fuses with the  FluxPrefetch and can simply poll values from its queue.

Tariod: Also, PublishOn doesn't implement the ConditionalSubscriber interface and I can't replace PrefetchSubscriber with PrefetchConditionalSubscriber.

simonbasle: ok got it. `PublishOnConditionalSubscriber` naming is a bit confusing, but I don't really have an idea for a better name to convey that it _uses_ the downstream as a `ConditionalSubscriber` 🤔

simonbasle:
I was running all Junit tests under reactor-core.test module for coverage (including loops, that are usually executed during releases), and I see some issues:
 - `FluxPrefetch` only has 40% coverage
 - `FluxPublish` has 82% coverage
 - `FluxPublishOnLoop` running for 11 minutes !

image

simonbasle:
>     * `FluxPublishOnLoop` running for 11 minutes !

I'm running the `main` version to compare, this might not have anything to do with your change @Tariod 😉

reactor-core/src/main/java/reactor/core/publisher/FluxPrefetch.java
```diff
@@ -0,0 +1,1404 @@
...
+			return fusionMode;
+		}
+
+		@Override
+		public int size() {
+			return queue.size();
+		}
+	}
+
+	static final class PrefetchConditionalSubscriber<T>
```
simonbasle, line 744: I have stopped reviewing here for now. Is there any way some code can be shared between the two subscriber classes?

simonbasle (changes requested):
have started to review more in depth, and I have a number of comments. I've set the `PrefetchConditionalSusbcriber` aside for now, as there was already quite a lot of logic to grok in the normal one.

chemicL:
I see there was a lot of work put into this. However, the last interaction here was 2.5 years ago. I will close the PR for now, but please create a new PR targeting the latest codebase in case you feel this is still applicable and would like to make progress on it.

PR #2710, opened 2021-05-20
Polish OperatorDisposablesTest with RepeatedTests annotation

eas5:
Problem:

No problem is found in this test. This is a refactoring suggestion that simplifies test design by eliminating a repetition structure whose increment was not being used during the repetition and can be replaced by an annotation.

Solution:

The @RepeatedTest annotation provides the total number of repetitions desired. Each invocation of a repeated test behaves like the execution of a regular @Test method with full support for the same lifecycle callbacks and extensions, which improves test design and eases future maintenance activities when needed.

### DIFF
--- a/reactor-core/src/test/java/reactor/core/publisher/OperatorDisposablesTest.java
+++ b/reactor-core/src/test/java/reactor/core/publisher/OperatorDisposablesTest.java
@@ -19,6 +19,7 @@ package reactor.core.publisher;
 import java.util.concurrent.atomic.AtomicReferenceFieldUpdater;
 
 import org.junit.jupiter.api.Test;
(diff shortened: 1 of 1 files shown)


simonbasle (changes requested):
that's a good idea, since the iteration # doesn't really matter even in the case one of the iterations fails 👍 

could you just fix the formatting in the lines you changed from using spaces to tabs? you could also remove the `public` modifier of the tests you changed since JUnit 5 doesn't require it anymore (but we didn't mass-remove the modifiers when we migrated from JUnit4 to 5).

eas5:
Thank you for your reply @simonbasle! I made the changes you suggested. Glad to help :)

simonbasle (approved):

PR #2736, opened 2021-07-07
Add some url to Callback Hell

zvischutz:
You can use this one as an example 
https://en.wiktionary.org/wiki/callback_hell


simonbasle (commented):
thanks for the suggestion. it certainly doesn't hurt, but the wiktionary entry is quite terse so I wonder if it brings that much benefit 🤔

PR #2799, opened 2021-09-30
Fix Mono.flatMapIterable javadoc: no prefetch parameter

rs017991:
`Mono.flatMapIterable` does not actually accept a `prefetch` argument.

It appears that this doc was copied from `Flux.flatMapIterable` (which does accept a `prefetch` argument).


simonbasle (approved):
thanks @rs017991, good catch

PR #2834, opened 2021-11-04
Fix flatMapMany description typo

EMellau:


simonbasle (approved):
good catch, thanks

EMellau:
> good catch, thanks
image

Issue #1898, opened 2019-09-19
windowTimeout - Flux disposed prematurely


kbeineke:
When using `windowTimeout(maxElements, duration)` in combination with `concat`/`concatWith` with the following preconditions:
- the number of concatenated elements is greater than `maxElements`,
- `duration` is small and the following operation rather slow.

Then, the following exception is thrown (and not all concatenated elements are processed):

```
11:09:26.874 [parallel-1] DEBUG reactor.core.publisher.Operators - onNextDropped: UnicastProcessor
11:09:26.879 [parallel-1] ERROR reactor.core.publisher.Operators - Operator called default onErrorDropped
reactor.core.Exceptions$ReactorRejectedExecutionException: Scheduler unavailable
	at reactor.core.Exceptions.failWithRejected(Exceptions.java:249)
	at reactor.core.publisher.Operators.onRejectedExecution(Operators.java:807)
	at reactor.core.publisher.FluxWindowTimeout$WindowTimeoutSubscriber.newPeriod(FluxWindowTimeout.java:188)
	at reactor.core.publisher.FluxWindowTimeout$WindowTimeoutSubscriber.drainLoop(FluxWindowTimeout.java:385)
	at reactor.core.publisher.FluxWindowTimeout$WindowTimeoutSubscriber$ConsumerIndexHolder.run(FluxWindowTimeout.java:440)
	at reactor.core.scheduler.PeriodicWorkerTask.call(PeriodicWorkerTask.java:59)
	at reactor.core.scheduler.PeriodicWorkerTask.run(PeriodicWorkerTask.java:73)
	at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:515)
	at java.base/java.util.concurrent.FutureTask.runAndReset(FutureTask.java:305)
	at java.base/java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:305)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
	at java.base/java.lang.Thread.run(Thread.java:834)
Caused by: java.util.concurrent.RejectedExecutionException: Scheduler unavailable
	at reactor.core.Exceptions.<clinit>(Exceptions.java:502)
[...]

PR #3054, opened 2022-05-26
Implement variant of windowTimeout with fairBackpressure

OlegDokuka:


reactor-core/src/main/java/reactor/core/publisher/FluxWindowTimeout.java
```diff
@@ -26,52 +26,1590 @@
...
+					return state;
+				}
+			}
+		}
+
+		boolean sendNext(T t) {
+			int received = this.received + 1 ;
+			this.received = received;
+
+			this.queue.offer(t);
```
simonbasle, line 1014: should we guard against an already cancelled window here and directly discard each element before they get a chance to be added to the queue?

reactor-core/src/main/java/reactor/core/publisher/FluxWindowTimeout.java
```diff
@@ -26,52 +26,1590 @@
...
+					expectedState =
+							(((previousState &~ACTIVE_WINDOW_INDEX_MASK) &~ HAS_UNSENT_WINDOW) ^ (expectedState == previousState ? HAS_WORK_IN_PROGRESS : 0)) |
+									incrementActiveWindowIndex(previousState) |
+									HAS_UNSENT_WINDOW;
+					previousState |= HAS_UNSENT_WINDOW;
+
+					if (isCancelled(expectedState)) {
+						previousWindow.sendCancel();
+						nextWindow.sendCancel();
+						Operators.onDiscard(nextWindow, this.actual.currentContext());
```
simonbasle, line 588: same as above, prefer triggering a discard of the queued values (which I think `sendCancel` should already cover)

reactor-core/src/main/java/reactor/core/publisher/FluxWindowTimeout.java
```diff
@@ -26,52 +26,1590 @@
...
+									(shouldBeUnsent ? HAS_UNSENT_WINDOW : 0);
+							// we need to put unsent flag here because we check at the
+							// beginning of the loop the presence for unsent flag from
+							// the previousState
+							previousState = (previousState &~ HAS_UNSENT_WINDOW) | (shouldBeUnsent ? HAS_UNSENT_WINDOW : 0);
+
+							if (isCancelled(expectedState)) {
+								nextWindow.sendCancel();
+								if (shouldBeUnsent) {
+									Operators.onDiscard(nextWindow, this.actual.currentContext());
```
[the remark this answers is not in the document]
simonbasle: One thing that came to my mind regarding which `currentContext` to use. There are two options:
 1. `this.actual.currentContext()` (only available here at the coordinator's level)
 2. `nextWindow.actual.currentContext()` (available in both cases)

I think that for consistency and predictability it would be better to always use the window's reachable context for the purpose of discarding the queued elements.

That said, there is an interesting aspect to inner Flux in a window operator: unlike in eg. `flatMap`, the subscription to the window can be a few steps down the line. Operators in the main sequence in between might define some `contextWrite` operation that isn't visible to the window as a result.

```java
myFlux.window(/*...*/)
    .doOnDiscard(e -> System.out.println("DISCARD A " + e")) // not invoked
    .concatMap(window -> window
        .doOnDiscard(e -> System.out.println("DISCARD from window: " + e"))
        .map(e -> e.toString())
    );
```


It would probably be worth it to document that in a discard support javadoc paragraph on the `Flux` API.

reactor-core/src/main/java/reactor/core/publisher/FluxWindowTimeout.java
```diff
@@ -26,52 +26,1590 @@
...
+				}
+				else {
+					expectedState = markWorkDone(this, expectedState);
+					previousState = expectedState | HAS_WORK_IN_PROGRESS;
+
+					if (isCancelled(expectedState)) {
+						final InnerWindow<T> currentWindow = this.window;
+						final long previousWindowState = currentWindow.sendCancel();
+						if (!InnerWindow.isSent(previousWindowState)) {
+							Operators.onDiscard(currentWindow, this.actual.currentContext());
```
simonbasle, line 628: same as above, don't discard the window itself

reactor-core/src/main/java/reactor/core/publisher/FluxWindowTimeout.java
```diff
@@ -26,52 +26,1590 @@
...
+				return;
+			}
+
+			this.s.cancel();
+
+			final InnerWindow<T> currentActiveWindow = this.window;
+			if (currentActiveWindow != null) {
+				if (!InnerWindow.isSent(currentActiveWindow.sendCancel())) {
+					if (!hasWorkInProgress(previousState)) {
+						Operators.onDiscard(currentActiveWindow, this.actual.currentContext());
```
simonbasle, line 664: same as above, don't discard the window itself

simonbasle (changes requested):
nice and huge work ! my main concern is around discarding of window queued elements rather than the window itself.

simonbasle:
relates to (fixes?): #1099, #1898 and #2920

reactor-test/src/main/java/reactor/test/util/RaceTestUtils.java
```diff
@@ -177,7 +179,12 @@ public static void race(int timeoutSeconds, Scheduler s, final Runnable... rs) {
 
 		try {
 			if (!cdl.await(timeoutSeconds, TimeUnit.SECONDS)) {
-				throw new AssertionError("RaceTestUtils.race wait timed out after " + timeoutSeconds + "s");
+				throw new AssertionError("RaceTestUtils.race wait timed out after " + timeoutSeconds + "s" + Thread.getAllStackTraces()
```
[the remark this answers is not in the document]
simonbasle: I'd suggest adding an overload with a `boolean collectAllStackTraces` parameter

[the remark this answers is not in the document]
OlegDokuka: done

simonbasle (approved):

PR #3077, opened 2022-06-16
Add StateLogger for structured logging of state machines

OlegDokuka:
Adds state logger useful for state machine debugging purpose


simonbasle (changes requested):
nice !
a few minor polishing suggestions:
 1. even if it is package-private I would consider adding minimal javadoc to `StateLogger`
 2. maybe a minimal amount on `FastLogger` too (it is public, but in jcstress folder)
 3. I would consider adding a constructor for subscribers eg. `SwitchOnFirstConditionalMain` without the `StateLogger` parameter, rather than having to change all unit tests to pass in `null`

OlegDokuka:
@simonbasle done

simonbasle (approved):

PR #3089, opened 2022-06-23
Add disposeGracefully method to Scheduler

chemicL:
Currently, all `Scheduler`s forcefully shutdown the underlying `ExecutorService`s by calling `shutdownNow()` method. In some scenarios, that is undesired as it does not allow proper cleanup.
`Scheduler`s should allow shutting down by not accepting new work, but giving the currently executing tasks a chance to finish without interruption. A `Mono<Void> disposeGracefully()` method has been added for that purpose. Upon subscription, it calls `shutdown()` instead of `shutdownNow()` and creates a background task that does `awaitTermination()` to complete the returned `Mono`. It can be combined with `timeout()` and `retry()` operators in realistic scenarios.


reactor-core/src/main/java/reactor/core/Disposable.java
```diff
@@ -165,4 +167,11 @@ default boolean addAll(Collection<? extends Disposable> ds) {
 		 */
 		int size();
 	}
+
+	// TODO(dj): add javadoc
```
simonbasle, line 171: the semantics of disposeGracefully may vary widely, so I would make the documented contract say that explicitly (eg. "each class implementing this trait should define how subsequent calls behave during the grace period and after it")

simonbasle: one limitation I'm thinking of with this API is that most of the time the underlying resource(s) being disposed gracefully will be atomically swapped out. Which means that even if one re-subscribes to the `Mono`, including with `onErrorResume()` or `retry()`, the underlying resources won't be reachable anymore.

thus, there will be no way of composing operators to fall back to a "hard" shutdown once the graceful shutdown is initiated.

I'm thinking this is fine if documented. The recommendation for implementors should probably be to trigger a hard dispose at the end of the `gracePeriod` THEN propagate a `TimeoutException`, noting that it only serves as a warning / logging but cannot be recovered.

reactor-core/src/main/java/reactor/core/scheduler/BoundedElasticScheduler.java
```diff
@@ -204,6 +210,25 @@ public void dispose() {
...
 
+	@Override
+	public Mono<Void> disposeGracefully(Duration gracePeriod) {
+		BoundedServices services = BOUNDED_SERVICES.get(this);
+		if (services != SHUTDOWN && BOUNDED_SERVICES.compareAndSet(this, services, SHUTDOWN)) {
+			ScheduledExecutorService e = EVICTOR.getAndSet(this, null);
+			Mono<Void> evictorDone;
+			if (e != null) {
+				Sinks.Empty<Void> done = Sinks.empty();
+				Schedulers.shutdownAndAwait(e, gracePeriod, done);
```
simonbasle, line 221: at one point I was thinking that these `EVICTOR` threads in the various pooling schedulers would be a good candidate to run the `awaitTermination` on. that said, it doesn't scale well since `awaitTermination` is blocking: a pool of 3 workers each with their own pending task of 2s would take 6s to terminate gracefully this way, but 2s only if each await is run on a dedicated thread.

reactor-core/src/main/java/reactor/core/scheduler/Schedulers.java
```diff
@@ -1553,4 +1557,51 @@ static final Object scanExecutor(Executor executor, Scannable.Attr key) {
...
+					disposedNotifier.tryEmitError(new TimeoutException("Executor " +
+							"shutdown timed out"));
+				}
+			} catch (InterruptedException e) {
+				disposedNotifier.tryEmitError(e);
+			}
+		});
+	}
+
+	static final void shutdownAndAwait(ScheduledExecutorService[] executors,
```
simonbasle, line 1579: let me think a bit more about a way to get the best of both worlds: awaitTermination + isTerminated mix that consider a bunch of executors and awaits as little as possible, ideally only blocking a single thread. right now, we need to recreate as many threads as there are in the thread pool that is being shut down gracefully, just for the purpose of `awaitTermination`s 😢

simonbasle (commented):
this is shaping up good, but the review triggers a few thoughts on my part and surface additional corner cases 🤔

simonbasle:
@chemicL what do you think of this piece of code to shutdown multiple executors at once, try to await as close to the grace period as possible while still only needing one thread, and finally `shutdownNow` in case the whole thing takes more than `gracePeriod`?

```java

	//this supposes that we somehow can get all the executors and swap them with an empty array
	//with more advanced schedulers like BoundedElasticScheduler, we might not get an ExecutorService array
	//but an array of another resource (like BoundedState[]), hence the Function
	static <RES> void shutdownAndAwait(final RES[] resources, Function<RES, ExecutorService> serviceExtractor, Duration gracePeriod, Sinks.Empty<Void> disposeNotifier) {
		for (int i = 0; i < resources.length; i++) {
			ExecutorService service = serviceExtractor.apply(resources[i]);
			service.shutdown();
		}

		//TODO: use a configurable separate pool?
		final ExecutorService service = Executors.newSingleThreadExecutor();
		service.submit(() -> {
			long nanoStart = System.nanoTime();
			long nanoGraceRemaining = gracePeriod.toNanos();

			boolean allAwaited = true;
			//wait for one executor at a time
			int index = 0;
			while (index < resources.length) {
				ExecutorService toAwait = serviceExtractor.apply(resources[index]);;
				//short case: the current executor has already terminated
				if (toAwait.isTerminated()) {
					index++;
					continue;
				}

				//we're inspecting the next executor and giving it nanoGraceRemaining ns to terminate gracefully
				try {
					if (!toAwait.awaitTermination(nanoGraceRemaining, TimeUnit.NANOSECONDS)) {
						//if it didn't terminate gracefully, the whole graceful operation can be considered a failure
						allAwaited = false;
						break;
					}
					else {
						//update the nanoGraceRemaining so that the global operation is within gracePeriod bounds
						long oldStart = nanoStart;
						nanoStart = System.nanoTime();
						nanoGraceRemaining = Math.max(0, nanoStart - oldStart);
						index++;
					}
				}
					allAwaited = false;
					break;
				}
			}
			if (allAwaited) {
				disposeNotifier.tryEmitEmpty();
			}
			else {
				for (int i = 0; i < resources.length; i++) {
					ExecutorService executorService = serviceExtractor.apply(resources[i]);
					executorService.shutdownNow();
				}
				disposeNotifier.tryEmitError(new TimeoutException("Scheduler didn't shutdown gracefully in time (" + gracePeriod + "), used shutdownNow"));
			}
		});
	}
```

reactor-core/src/main/java/reactor/core/Disposable.java
chemicL: We discussed the possible contracts here with @OlegDokuka. The approach with forceful shutdown before propagating the `TimeoutException` and non-recoverable errors might be confusing to users if they don't read the specific `Scheduler` documentation, but it has some advantages implementation wise.
Another approach could be to propagate a retry-able error and allow re-initiating the shutdown + await procedure, while also allowing for an explicit final call to explicit `shutdownNow()` when desired. I'll go back to the original issue and ask for opinion from the user's perspective to guide the design.

OlegDokuka: My personal opinion is that once we fix specific behavior (e.g. call `shutdownNow()` if timeout or interaction exception) then we probably will end up with everyone doing `scheduler.disposeGracefully(Duration.ofHours(9999999)).subscribe()` or then complaining that they did not wont to have `shutdnowNow` called but rather retry later 

Another thought on `TimoutException` - any exception is useless if we can not do anything useful after that. I'm not sure that logging such an event makes any sense. This exception is just a fact that we forced shutdown process so use just has to take it. Also, taking into account the impl details - all other active subscribers are going to get the same notification but the other late subscriber will not get it, then it is going to be too confusing so even having it documented will not resolve this confusion.

My personal recommendation is to prefer flexibility over fixed behavior. One can always write something like the following to mimic what we can hardcode

```
scheduler.disposeGracefully(Duration.ofMillis(100))
    .retryWhen(Retry.backoff(5, Duration.ofMillis(50)))
    .onErrorResume(e -> Mono.fromRunnable(scheduler::dispose));
```

reactor-core/src/test/java/reactor/core/scheduler/SchedulersTest.java
```diff
@@ -734,34 +734,6 @@ public void immediateTaskIsExecuted() throws Exception {
 		assertThat((end - start) >= 1000).as("Timeout too long").isTrue();
 	}
 
-	@Test
-	public void immediateTaskIsSkippedIfDisposeRightAfter() throws Exception {
```
chemicL, line 738: This is no longer the case. It still could race, but now there's a few more instructions in `dispose()` so the task gets aborted. IMO it's not a _feature_ worth testing - the test was introduced in https://github.com/reactor/reactor-core/commit/6f3383d5a4fc533396a3522095cd4970024849e2 but one should not rely on a race to dispose tasks.

reactor-core/src/main/java/reactor/core/scheduler/BoundedElasticScheduler.java
```diff
@@ -160,61 +126,112 @@ BoundedScheduledExecutorService createBoundedExecutorService() {
...
+					// and that we capture the BoundedStates, otherwise we'd lose references to schedulers
+					// that we need to wait for.
+					final BoundedState[] toAwait = terminating.boundedServicesBeforeShutdown.dispose(false);
+					if (terminating.evictor != null) {
+						terminating.evictor.shutdown();
+					}
+
+					SchedulerState terminated = terminated(terminating, toAwait);
+
+					STATE.set(this, terminated);
```
OlegDokuka, line 214: how about `lazySet` here? It looks like you assume no contention while state is terminating, thus it is safe doing lazySet

reactor-core/src/main/java/reactor/core/scheduler/BoundedElasticScheduler.java
```diff
@@ -160,61 +126,112 @@ BoundedScheduledExecutorService createBoundedExecutorService() {
...
+				if (STATE.compareAndSet(this, currentState, terminating)) {
+					// We need an intermediate state to make sure the disposal happens just once
+					// and that we capture the BoundedStates, otherwise we'd lose references to schedulers
+					// that we need to wait for.
+					final BoundedState[] toAwait = terminating.boundedServicesBeforeShutdown.dispose(false);
+					if (terminating.evictor != null) {
+						terminating.evictor.shutdown();
+					}
+
+					SchedulerState terminated = terminated(terminating, toAwait);
```
OlegDokuka, line 212: looks like terminated with a reference to an old awaiter is used only here, this my suggestion would be to either, use the current login here only, or creating a separate method. Sounds like for all other scneratios the old references is redundant

OlegDokuka (changes requested):
Nice progress overall. Left my comments. Also, there are a set of general polishing points:

1. Let's use plain volatile access (e.g. `this.state` instead of `STATE.get(this)`) plain access is the fastest approach and we use it consistently through the codebase. `AtomicXXXFieldUpdate#get` is an option to get value when plain access is impossible (e.g. shared utils function i.e `Operators#requested`) 
2. Let's make sure imports are not collapsed
3. `ShedulerState#terminated` is not making useful of the old state except bounded elastic, thus lets avoid `terminated(old)`  and use `static final TERMINATED_STATE = new SchedulerState(dead_executor_service, Mono.empty())` where possible ;

reactor-core/src/main/java/reactor/core/scheduler/SchedulerState.java
```diff
@@ -0,0 +1,79 @@
...
+				    })
+				    .replay()
+				    .refCount()
+				    .next()
+		);
+	}
+
+	static SchedulerState terminated(@Nullable SchedulerState base) {
+		return new SchedulerState(TERMINATED,
+				base == null ? Mono.empty() : base.onDispose);
```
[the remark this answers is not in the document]
chemicL: The `base == null` happens when the scheduler is not yet started. The `Mono.empty()` makes sense only in that scenario. Otherwise, the `onDispose` is bound to the used executor, while the current executor is set to the terminated one. That's why `Mono.empty()` can't be used - the concurrent readers of the terminated state need the original `onDispose`, not a fresh and empty one. That's also answering the general comment about no need for `SchedulerState#terminated` - I do think it's needed.

reactor-core/src/main/java/reactor/core/scheduler/SchedulerState.java
```diff
@@ -0,0 +1,79 @@
...
+					    Thread backgroundThread = new Thread(() -> {
+						    while (!Thread.currentThread()
+						                  .isInterrupted()) {
+							    try {
+								    if (executor.awaitTermination(1, TimeUnit.SECONDS)) {
+									    sink.complete();
+									    return;
+								    }
+							    }
+							    catch (InterruptedException e) {
```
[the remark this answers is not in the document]
chemicL: What are you referring to? Catching the IE or the way it's handled? Catching is necessary, handling it with default restoration of the interrupted state and stopping the execution seems acceptable, but not sure the interrupted state is necessary, I think it can be skipped. The reactive chain cancellation triggers interrupt.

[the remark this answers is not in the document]
OlegDokuka: referencing to

```
Thread.currentThread()
								          .interrupt();
                                          ```

File: reactor-core/src/main/java/reactor/core/scheduler/BoundedElasticScheduler.java (hunk 6 above)
Comment (chemicL) [reply reactor-core/src/main/java/reactor/core/scheduler/BoundedElasticScheduler.java:244]: The `terminated` state of `BoundedElasticScheduler` is used 3 times in this class - twice it uses the initial state passed as arguments to preserve the disposal references and once with null upon start. It might be confusing that there's a `SchedulerState` as a general type in the package, while this implementation is different and makes sense for this particular scheduler.

File: reactor-core/src/main/java/reactor/core/scheduler/BoundedElasticScheduler.java [hunk 9]
```diff
@@ -160,61 +146,159 @@ BoundedScheduledExecutorService createBoundedExecutorService() {
...
+		);
 		for (;;) {
-			BoundedServices services = BOUNDED_SERVICES.get(this);
-			if (services != SHUTDOWN) {
-				return;
+			a = this.state;
+			if (a.currentResource.boundedServices == SHUTTING_DOWN) {
+				// keep spinning until previous shuts down
+				continue;
 			}
```
OlegDokuka, line 167: SHUTTING_DOWN should not be a reason to continue spinning. We have to proceed and set a new scheduler or return immediately

reactor-core/src/main/java/reactor/core/scheduler/BoundedElasticScheduler.java
```diff
@@ -160,61 +146,159 @@ BoundedScheduledExecutorService createBoundedExecutorService() {
...
+				final BoundedState[] toAwait =
+						shuttingDown.initialResource.boundedServices.dispose(true);
+				shuttingDown.initialResource.boundedServices.evictor.shutdownNow();
+
+				SchedulerState<BoundedServicesState> shutdown = SchedulerState.transition(
+						disposed(shuttingDown.initialResource.boundedServices, toAwait),
+						SHUTDOWN_STATE,
+						this
+				);
+				STATE.lazySet(this, shutdown);
```
OlegDokuka, line 245: The lazySet is misused. Usually, it is combined with another strong set barrier, ensuring lazySet is flushed out. The nature of the lazySet barrier allows delaying this volatility, thus a thread that is spinning inside an infinite loop may burn a lot of cycles. The best here would be the usage of compareAndSet to ensure the operation is not overwriting state C which happened between your setA and setB

reactor-core/src/main/java/reactor/core/scheduler/BoundedElasticScheduler.java
```diff
@@ -160,61 +146,159 @@ BoundedScheduledExecutorService createBoundedExecutorService() {
...
+					final BoundedState[] toAwait =
+							shuttingDown.initialResource.boundedServices.dispose(false);
+					shuttingDown.initialResource.boundedServices.evictor.shutdown();
+					SchedulerState<BoundedServicesState> shutdown =
+							SchedulerState.transition(
+									disposed(shuttingDown.initialResource.boundedServices, toAwait),
+									SHUTDOWN_STATE,
+									this
+							);
+					STATE.lazySet(this, shutdown);
```
OlegDokuka, line 284: same as above

OlegDokuka (changes requested):

reactor-core/src/main/java/reactor/core/scheduler/BoundedElasticScheduler.java
OlegDokuka: my bad. I figured out that the usage of lazySet is incorrect

reactor-core/src/main/java/reactor/core/scheduler/BoundedElasticScheduler.java
```diff
@@ -160,61 +146,159 @@ BoundedScheduledExecutorService createBoundedExecutorService() {
...
+					return shutdown.onDispose;
+				} // else loop until the winner sets SHUTDOWN
+			}
+		}).timeout(gracePeriod);
+	}
```
OlegDokuka, line 289:
```
@Override
	public void dispose() {
		final SchedulerState<BoundedServicesState> previous = state;

		if (previous.currentResource.boundedServices == SHUTDOWN ) {
			// A dispose process might be ongoing, but we want a forceful shutdown,
			// so we do our best to release the resources without updating the state.
			if (previous.initialResource != null) {
				previous.initialResource.boundedServices.evictor.shutdownNow();
				if (previous.initialResource.disposedBoundedStates != null) {
					for (BoundedState bs : previous.initialResource.disposedBoundedStates) {
						bs.shutdown(true);
					}
				}
			}
			return;
		}

		final BoundedState[] toAwait =
				previous.currentResource.boundedServices.dispose(true);

		SchedulerState<BoundedServicesState> shuttingDown = SchedulerState.transition(
				previous.currentResource, SHUTTING_DOWN_STATE, this
		);
		if (STATE.compareAndSet(this, previous, shuttingDown)) {
			assert shuttingDown.initialResource != null;
			shuttingDown.initialResource.boundedServices.evictor.shutdownNow();

			SchedulerState<BoundedServicesState> shutdown = SchedulerState.transition(
					disposed(shuttingDown.initialResource.boundedServices, toAwait),
					SHUTDOWN_STATE,
					this
			);
			STATE.lazySet(this, shutdown);
		}
	}

	@Override
	public Mono<Void> disposeGracefully(Duration gracePeriod) {
		return Mono.defer(() -> {
			SchedulerState<BoundedServicesState> previous = state;

			if (previous.currentResource.boundedServices == SHUTDOWN) {
				return previous.onDispose;
			}

			final BoundedState[] toAwait =
					previous.currentResource.boundedServices.dispose(false);
			
			if (toAwait == BoundedServices.ALL_SHUTDOWN) {
				return Mono.empty();
			}
			
			previous.currentResource.boundedServices.evictor.shutdown();
			final SchedulerState<BoundedServicesState> shutdown =
					SchedulerState.transition(
							disposed(previous.currentResource.boundedServices, toAwait),
							SHUTDOWN_STATE,
							this
					);

			if (STATE.compareAndSet(this, previous, shutdown)) {
				return shutdown.onDispose;
			}

			// dispose concurrently set state but likely did not observe 
			// BoundedStates, so we need to force terminate it here
			for (BoundedState boundedState : toAwait) {
				boundedState.shutdown(true);
			}
			
			return Mono.empty();
		}).timeout(gracePeriod);
	}
	```

how about this as alternative solution

**Review by simonbasle — COMMENTED** [review]
LGTM from an API / javadoc perspective

File: reactor-core/src/main/java/reactor/core/scheduler/BoundedElasticScheduler.java (hunk 10 above)
Comment (chemicL) [reply reactor-core/src/main/java/reactor/core/scheduler/BoundedElasticScheduler.java:245]: Thanks for explaining. To my understanding there would not be a C in the current implementation, as the `SHUTTING_DOWN_STATE` is guarding any thread from injecting a C value before the lazy or eager set has effect. With the suggestion you made in `start()` it would now be possible, so for sure this logic needs to change. Further discussion follows in the suggested code snippet at the end of `disposeGracefully()`.

File: reactor-core/src/main/java/reactor/core/scheduler/BoundedElasticScheduler.java (hunk 12 above)
Comment (chemicL) [reply reactor-core/src/main/java/reactor/core/scheduler/BoundedElasticScheduler.java:289]: This looks interesting. No busy spinning. The trick with `boundedStates` disposal in case of a race sounds tempting, but no longer delivers the promise of graceful disposal within a given grace period. Another drawback of this approach seems to be the fact that the `dispose()` method would invalidate a concurrent `start()` by the forceful `lazySet` at the end (probably needs a change to `set` as per your other suggestions, or `compareAndSet` in which case it is probably possible to get rid of the undesired invalidation.
All in all, I feel the `boundedState.shutdown(true);` is against users' expectations.

Comment (chemicL) [reply reactor-core/src/main/java/reactor/core/scheduler/BoundedElasticScheduler.java:289]: Also, if there is a race, the forceful shutdown of `boundedState`s returns an empty `Mono` and gives no chance to await the disposal of `evictor` nor the actual disposal of the workers - even though they're forcefully shutdown, they might still take some time to release resources.

Comment (OlegDokuka) [reply reactor-core/src/main/java/reactor/core/scheduler/BoundedElasticScheduler.java:289]: @chemicL I did a mistake, forgot to remove old lazySet from the copypasted code. edited

Comment (OlegDokuka) [reply reactor-core/src/main/java/reactor/core/scheduler/BoundedElasticScheduler.java:289]: > All in all, I feel the boundedState.shutdown(true); is against users' expectations.

that was also a type, copypasted code

**Comment by chemicL** [comment]
The latest changes include improvements for avoiding looping in the `dispose`/`disposeGracefully`/`start` methods as suggested by @OlegDokuka. This was only possible by revisiting `BoundedElasticScheduler`'s inner workings and avoiding atomic replacing of underlying `BoundedState`s by a tombstone. Instead, an encapsulating class was introduced for state management of `BoundedServices`, which in turn simplified the state transitions in `BoundedElasticScheduler` itself with regards to the generic `SchedulerState`. As convoluted as it sounds, the latest set of changes also incorporates a fix for a leakage of `BoundedState` which would not be shutdown if `dispose` happened while an `ExecutorService` was being `pick`ed.

**Review by simonbasle — APPROVED** [review]

File: reactor-core/src/main/java/reactor/core/scheduler/ParallelScheduler.java [hunk 13]
```diff
@@ -79,74 +82,124 @@ public ScheduledExecutorService get() {
...
-            if (a != SHUTDOWN && a != null) {
-                if (b != null) {
-                    for (ScheduledExecutorService exec : b) {
-                        exec.shutdownNow();
-                    }
-                }
-                return;
-            }
+	public void start() {
+		SchedulerState<ScheduledExecutorService[]> a = this.state;
```
OlegDokuka, line 94: this should be within infinite loop

reactor-core/src/main/java/reactor/core/scheduler/SingleScheduler.java
```diff
@@ -71,59 +73,100 @@ public ScheduledExecutorService get() {
...
-		ScheduledExecutorService b = null;
-		for (; ; ) {
-			ScheduledExecutorService a = executor;
-			if (a != TERMINATED && a != null) {
-				if (b != null) {
-					b.shutdownNow();
-				}
-				return;
-			}
+		SchedulerState<ScheduledExecutorService> a = this.state;
```
OlegDokuka, line 84: missed infinite loop

reactor-core/src/main/java/reactor/core/scheduler/BoundedElasticScheduler.java
```diff
@@ -160,61 +139,124 @@ BoundedScheduledExecutorService createBoundedExecutorService() {
...
-		return BOUNDED_SERVICES.get(this) == SHUTDOWN;
+		return state.currentResource == SHUTDOWN;
 	}
 
 	@Override
 	public void start() {
-		for (;;) {
-			BoundedServices services = BOUNDED_SERVICES.get(this);
-			if (services != SHUTDOWN) {
+		SchedulerState<BoundedServices> a = this.state;
```
OlegDokuka, line 147: should not we keep for ;; loop here?

OlegDokuka (approved):
looks good overall with few minor comments

reactor-core/src/jcstress/java/reactor/core/scheduler/RacingDisposeGracefullyStressTest.java
```diff
@@ -0,0 +1,74 @@
...
+
+package reactor.core.scheduler;
+
+import java.time.Duration;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.RejectedExecutionException;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+
+abstract class RacingDisposeGracefullyStressTest<T extends Scheduler> {
```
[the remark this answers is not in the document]
chemicL: This class no longer exists.

reactor-core/src/main/java/reactor/core/scheduler/BoundedElasticScheduler.java
chemicL: I think it's not needed, can you elaborate?

OlegDokuka: yeah. I changed my mind

reactor-core/src/jcstress/java/reactor/core/scheduler/SchedulersStressTest.java
```diff
@@ -0,0 +1,288 @@
...
+		{
+			scheduler.start();
+			// Schedule a task that disallows graceful closure until the arbiter kicks in
+			// to make sure that actors fail while waiting.
+			scheduler.schedule(this::awaitArbiter);
+		}
+
+		@Actor
+		public void disposeGracefully1(ZZZ_Result r) {
+			r.r1 = checkDisposeGracefullyTimesOut();
```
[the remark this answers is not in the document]
chemicL: outdated version

simonbasle (approved):

simonbasle:
great job @chemicL ! finally approved and ready to merge 😄

now the only remaining step is to try to summarize the design of the change for the commit message 📖 
feel free to ping me if you want me to also review that.

PR #3165, opened 2022-08-24
Replace 'long' padding by 'byte' padding to avoid fields re-ordering on modern JDK

lantalex:
Hello,
On modern JDK (15+) good old trick using `long` fields as padding is no longer [reliable](https://shipilev.net/jvm/objects-inside-out/#_superhierarchy_gaps_in_java_15). To prevent false-sharing we could use `byte` padding in the same way, as in original [SpscArrayQueue](https://github.com/JCTools/JCTools/commit/d43b774ae40a34a6449b49d49ae96b8080a69abf#diff-26cefaad7d67f5b07622f85638bafd2496526c5a092da3e66b6c1e1f15df4d97)

### DIFF
--- a/reactor-core/src/main/java/reactor/util/concurrent/SpscArrayQueue.java
+++ b/reactor-core/src/main/java/reactor/util/concurrent/SpscArrayQueue.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2016-2021 VMware Inc. or its affiliates, All Rights Reserved.
+ * Copyright (c) 2016-2022 VMware Inc. or its affiliates, All Rights Reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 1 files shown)


chemicL:
Thank you for the PR. Please update the file header as per output of the initial CI run.

lantalex:
> Thank you for the PR. Please update the file header as per output of the initial CI run.

Fixed

chemicL:
Please consider changing the merge target to 3.4.x instead of main so this improvement can be made available quicker than in 3.5.0.

lantalex:
> Please consider changing the merge target to 3.4.x instead of main so this improvement can be made available quicker than in 3.5.0.

Created [new PR](https://github.com/reactor/reactor-core/pull/3168) for 3.4.x, closing this.

PR #3221, opened 2022-10-08
Make use of CompletionStage#handle instead of `whenComplete`

He-Pin:
Which has better performance.
https://github.com/line/armeria/pull/1596

### DIFF
--- a/reactor-core/src/main/java/reactor/core/publisher/MonoCompletionStage.java
+++ b/reactor-core/src/main/java/reactor/core/publisher/MonoCompletionStage.java
@@ -64,7 +64,7 @@ final class MonoCompletionStage<T> extends Mono<T>
             return;
         }
 
-        future.whenComplete((v, e) -> {
+        future.handle((v, e) -> {
             if (sds.isCancelled()) {
                 //nobody is interested in the Mono anymore, don't risk dropping errors
                 Context ctx = sds.currentContext();
@@ -79,7 +79,7 @@ final class MonoCompletionStage<T> extends Mono<T>
                     Operators.onDiscard(v, ctx);
                 }
 
-                return;
+                return null;
             }
             try {
                 if (e instanceof CompletionException) {
@@ -99,6 +99,7 @@ final class MonoCompletionStage<T> extends Mono<T>
                 Operators.onErrorDropped(e1, actual.currentContext());
                 throw Exceptions.bubble(e1);
             }
+            return null;
         });
     }
 


OlegDokuka (approved):

OlegDokuka:
@chemicL this should be forwardmerged. Lazy one uses same method as well

chemicL:
I just noticed #3222 which addresses the issue on `main`. I suppose we'll need to forward-merge but discard the change and take the one from the other PR instead.

He-Pin:
@chemicL Yes ,I created two PR, one for 3.4.x and one for main, I saw @OlegDokuka did more optimizations of memory usage on the current main branch:)
This is my first time contribution

OlegDokuka:
@He-Pin second is going to be closed since we use forward merge to keep both branches connected

He-Pin:
But in the second one, you fold the `BiConsumer` to `MonoCompletionStageSubscription`. so I do it in a separated PR. or will you backport that memory optimization to 3.4.x?

He-Pin:
image

jmh:run -i 5 -wi 5 -f1 -t1 .CompletionStageBenchmark.

```scala
package bench

import org.openjdk.jmh.annotations._

import java.util.concurrent.CompletableFuture
import java.util.concurrent.CountDownLatch
import java.util.concurrent.TimeUnit

@State(Scope.Benchmark)
@OutputTimeUnit(TimeUnit.SECONDS)
@BenchmarkMode(Array(Mode.Throughput))
class CompletionStageBenchmark {
  val ex = new RuntimeException("ex")

  @Benchmark
  def benchWhenCompleteFailedSync(): Unit = {
    val future = new CompletableFuture[Int]()
    future.completeExceptionally(ex)
    val latch = new CountDownLatch(1)
    future.whenComplete((_, _) => latch.countDown())
    latch.await()
  }

  @Benchmark
  def benchWhenCompleteSuccessSync(): Unit = {
    val future = new CompletableFuture[Int]()
    future.complete(1)
    val latch = new CountDownLatch(1)
    future.whenComplete((_, _) => latch.countDown())
    latch.await()
  }

  @Benchmark
  def benchHandleFailedSync(): Unit = {
    val future = new CompletableFuture[Int]()
    future.completeExceptionally(ex)
    val latch = new CountDownLatch(1)
    future.handle((_, _) => latch.countDown())
    latch.await()
  }

  @Benchmark
  def benchHandleSuccessSync(): Unit = {
    val future = new CompletableFuture[Int]()
    future.complete(1)
    val latch = new CountDownLatch(1)
    future.handle((_, _) => latch.countDown())
    latch.await()
  }

}

```

PR #3260, opened 2022-10-30
fixes incorrect value/error set during `tryEmitXXX` call

### LINKED ISSUE #3257: Mono created from a Sink.one sporadically turns empty
In our code, we create a `Sink.one()` whose value is emitted from onSuccess/onError handlers of some other `Mono`. A value being emitted is never `null`.

## Expected Behavior
Expectation is that a mono view of that sink will never be an empty mono, due to value emitted by a sink being non-null.

## Actual Behavior
Contary to our expectation, a Mono view of that sink _sporadically_ turns empty. It happens both on Windows 10 as well as on Ubuntu 20 machines, so it is likely not something platform-specific.

## Steps to Reproduce
Below is the shortest unit test I could come up with which reflects what we're doing in our production code, and when this test is run from IDE with "repeat until failure" setting, it fails the assertion about non-null Mono result quite consistently within 10-30 seconds after start.

I also added some diagnostic logging around tryEmitValue / tryEmitError, and looks like it's never an error, and never a null value being emitted, neither it's an emission failure.

```java
import java.util.concurrent.*;
import java.util.function.Consumer;

import lombok.extern.slf4j.Slf4j;
import reactor.core.publisher.*;
import reactor.util.Loggers;

import static org.assertj.core.api.AssertionsForClassTypes.assertThat;

@Slf4j
class MonoPlaygroundTest {

    private static ExecutorService executor;

    static {
        executor = Executors.newCachedThreadPool();
        Loggers.useSl4jLoggers();
    }

    @Test
    void shouldConstructNonEmptyMono() {
        var sink = Sinks.<String>one();

        Mono
            .fromSupplier(() -> "value")
            .log(Loggers.getLogger("sink-setter"))
            .subscribeOn(Schedulers.fromExecutorService(executor, "mono-executor"))
            .doOnSuccess(tryEmitValueWithDebugLogging(sink))
            .doOnError(tryEmitErrorWithDebugLogging(sink))
            .subscribe();

        var result = sink
            .asMono()
            .log(Loggers.getLogger("sink-result-receiver"))
            // more map / flatMap here...
            .block();

        assertThat(result).isNotNull();
    }

    private static Consumer<Throwable> tryEmitErrorWithDebugLogging(Sinks.One<String> sink) {
        return t -> {
            sink.tryEmitError(t);
            log.error("Was emitting an error: {}", t);
        };
    }

    private static Consumer<String> tryEmitValueWithDebugLogging(Sinks.One<String> sink) {
        return value -> {
            var result = sink.tryEmitValue(value);
            if (value == null) {
                log.error("Tried to emit null value");
            }
            if (result.isFailure()) {
                log.error("Failed to emit because of {}", result);
            }
        };
    }
}
```

The output from a failed run is like this:
```
23:48:12.635 [pool-1-thread-1] INFO  sink-setter - | onSubscribe([Fuseable] Operators.MonoSubscriber)
23:48:12.635 [pool-1-thread-1] INFO  sink-setter - | request(unbounded)
23:48:12.635 [boundedElastic-1] INFO  sink-result-receiver - onSubscribe(SinkOneMulticast.NextInner)
23:48:12.635 [pool-1-thread-1] INFO  sink-setter - | onNext(value)
23:48:12.635 [boundedElastic-1] INFO  sink-result-receiver - request(unbounded)
23:48:12.635 [pool-1-thread-1] INFO  sink-setter - | onComplete()
23:48:12.635 [boundedElastic-1] INFO  sink-result-receiver - onComplete()

java.lang.AssertionError: 
Expecting actual not to be null
```

## Your Environment

* Reactor version(s) used: 3.4.22
* JVM version (`java -version`): openjdk version "11.0.16.1" 2022-07-19 LTS
* OS and version (eg `uname -a`): Windows 10 / 21H2 build 19044.2130, also Linux 5.4.0-96-generic SMP x86_64 GNU/Linux

OlegDokuka:
closes #3257 

Signed-off-by: OlegDokuka <odokuka@vmware.com>

### DIFF
--- a/reactor-core/src/main/java/reactor/core/publisher/SinkEmptyMulticast.java
+++ b/reactor-core/src/main/java/reactor/core/publisher/SinkEmptyMulticast.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2020-2021 VMware Inc. or its affiliates, All Rights Reserved.
+ * Copyright (c) 2020-2022 VMware Inc. or its affiliates, All Rights Reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -37,9 +37,14 @@ class SinkEmptyMulticast<T> extends Mono<T> implements InternalEmptySink<T> {
 
 	@SuppressWarnings("rawtypes")
 	static final Inner[] EMPTY = new Inner[0];
-
 	@SuppressWarnings("rawtypes")
-	static final Inner[] TERMINATED = new Inner[0];
+	static final Inner[] TERMINATED_EMPTY = new Inner[0];
+	@SuppressWarnings("rawtypes")
+	static final Inner[] TERMINATED_ERROR = new Inner[0];
+
+	static final int STATE_ADDED = 0;
+	static final int STATE_ERROR = -1;
+	static final int STATE_EMPTY = -2;
 
 	@Nullable
 	Throwable error;
@@ -58,12 +63,23 @@ class SinkEmptyMulticast<T> extends Mono<T> implements InternalEmptySink<T> {
 		return this;
 	}
 
+	boolean isTerminated(Inner<?>[] array) {
+		return array == TERMINATED_EMPTY || array == TERMINATED_ERROR;
+	}
+
 	@Override
 	public EmitResult tryEmitEmpty() {
-		Inner<?>[] array = SUBSCRIBERS.getAndSet(this, TERMINATED);
+		Inner<T>[] array;
+		for (;;) {
+			array = this.subscribers;
 
-		if (array == TERMINATED) {
-			return Sinks.EmitResult.FAIL_TERMINATED;
+			if (isTerminated(array)) {
+				return EmitResult.FAIL_TERMINATED;
+			}
+
+			if (SUBSCRIBERS.compareAndSet(this, array, TERMINATED_EMPTY)) {
+				break;
+			}
 		}
 
 		for (Inner<?> as : array) {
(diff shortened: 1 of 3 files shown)


reactor-core/src/main/java/reactor/core/publisher/SinkOneMulticast.java
```diff
@@ -25,58 +25,51 @@
...
 
-		error = cause;
-		value = null;
-
-		for (Inner<O> as : prevSubscribers) {
-			as.error(cause);
-		}
-		return EmitResult.OK;
-	}
+		this.value = value;
```
UgiR, line 49: More of a question for my own understanding: Does the stashing of this value for future subscribers rely on the fact that the 'safe' entry point to this sink will be wrapped in a serialized sink? In other words, is it OK for the below test to fail?

```java
@RepeatedTest(5)
void test() throws InterruptedException {
  final SinkOneMulticast < String > sink = new SinkOneMulticast < > ();
  final AtomicReference < String > successfullyEmittedElement = new AtomicReference < > ();
  final AtomicReference < String > firstSubscriptionOnNext = new AtomicReference < > ();
  final CountDownLatch firstSubscriptionLatch = new CountDownLatch(1);

  // first subscription before element is emitted through sink
  sink.asMono()
    .subscribe(a -> {
      firstSubscriptionOnNext.compareAndSet(null, a);
      firstSubscriptionLatch.countDown();
    });

  // two competing tryEmitValue calls -- I assume this is where I break the spec?
  RaceTestUtils.race(
    () -> {
      if (sink.tryEmitValue("foo").isSuccess()) {
        if (!successfullyEmittedElement.compareAndSet(null, "foo")) {
          fail("Expected only a single successful emit");
        }
      }
    },
    () -> {
      if (sink.tryEmitValue("bar").isSuccess()) {
        if (!successfullyEmittedElement.compareAndSet(null, "bar")) {
          fail("Expected only a single successful emit");
        }
      }
    }
  );

  firstSubscriptionLatch.await();
  assertThat(firstSubscriptionOnNext.get()).isEqualTo(successfullyEmittedElement.get());

  // second subscription *after* sink is 'terminated'
  final String secondSubscriptionOnNext = sink.asMono().block();
  assertThat(secondSubscriptionOnNext).isEqualTo(successfullyEmittedElement.get());

}
```

I assume it is my responsibility in this case to not break the spec by racing two `tryEmitValue` calls, but just want to double check.

OlegDokuka: Absolutely correct. It is only possible to get there if you use Sinks.unsafe() spec which implies more advanced capabilities but unsafe

simonbasle (approved):

reactor-core/src/main/java/reactor/core/publisher/SinkEmptyMulticast.java
```diff
@@ -77,23 +93,36 @@ public EmitResult tryEmitEmpty() {
...
 
-		Inner<T>[] prevSubscribers = SUBSCRIBERS.getAndSet(this, TERMINATED);
-		if (prevSubscribers == TERMINATED) {
+		Inner<T>[] prevSubscribers = this.subscribers;
+
+		if (isTerminated(prevSubscribers)) {
 			return EmitResult.FAIL_TERMINATED;
 		}
 
 		error = cause;
```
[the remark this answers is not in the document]
OlegDokuka: This is not the goal of this implementation to cover such cases

PR #3264, opened 2022-11-01
add Flux.mergePriority that does not wait for all sources to emit

osi:
this is a solution for  #2827

modify FluxMergeComparing to support not waiting for all sources to emit an item before comparing and sending downstream.

if this approach is amenable, i will update the marble diagrams.


OlegDokuka (approved):

simonbasle (changes requested):
sounds like a workable approach, with minor comments on documentation. @osi do you need this in a `3.4.x` release? (in which case this is targeting the wrong branch). also did you want to make a dedicated marble diagram?

osi:
@simonbasle awesome - yes, i'm happy to target 3.4.x as that's the version we're using. 

i wanted to ensure the approach was reasonable before updating the marble diagrams. i'll make the suggested changes and target that branch

osi:
I'm closing this in favor of #3267 which targets 3.4.x. I've addressed the comments there, but still need to update the marble diagrams.

PR #3266, opened 2022-11-02
Add Mono::fromCompletionStage overrides suppressing cancellation

chemicL:
Recently, Mono::fromFuture operators were enhanced to support suppresion of subscription cancellation being propagated to the source Future. In case of CompletionStage, this overrides can be also useful, since cancelling the source was a behaviour change and users might need to default to the old behaviour.

### DIFF
--- a/reactor-core/src/main/java/reactor/core/publisher/Mono.java
+++ b/reactor-core/src/main/java/reactor/core/publisher/Mono.java
@@ -547,7 +547,30 @@ public abstract class Mono<T> implements CorePublisher<T> {
 	 * @return A {@link Mono}.
 	 */
 	public static <T> Mono<T> fromCompletionStage(CompletionStage<? extends T> completionStage) {
-		return onAssembly(new MonoCompletionStage<>(completionStage, false));
+		return fromCompletionStage(completionStage, false);
+	}
+
+	/**
+	 * Create a {@link Mono}, producing its value using the provided {@link CompletionStage}.
+	 *
+	 * <p>
+	 * <img class="marble" src="doc-files/marbles/fromFuture.svg" alt="">
+	 * <p>
+	 * Note, use {@link #fromFuture(CompletableFuture, boolean)} with {@code
+	 * suppressCancellation} set to {@code true} if you need to suppress cancellation
+	 * propagation
+	 *
+	 * @param completionStage {@link CompletionStage} that will produce a value (or a null to
+	 * complete immediately)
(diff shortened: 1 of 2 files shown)


reactor-core/src/main/java/reactor/core/publisher/Mono.java
```diff
@@ -547,7 +547,30 @@ public static <T> Mono<T> fromCallable(Callable<? extends T> supplier) {
...
+		return fromCompletionStage(completionStage, false);
+	}
+
+	/**
+	 * Create a {@link Mono}, producing its value using the provided {@link CompletionStage}.
+	 *
+	 * <p>
+	 * <img class="marble" src="doc-files/marbles/fromFuture.svg" alt="">
+	 * <p>
+	 * Note, use {@link #fromFuture(CompletableFuture, boolean)} with {@code
```
violetagg, line 559:
```suggestion
	 * Note, use {@link #fromCompletionStage(CompletableFuture, boolean)} with {@code
```

reactor-core/src/main/java/reactor/core/publisher/Mono.java
```diff
@@ -567,7 +590,34 @@ public static <T> Mono<T> fromCompletionStage(CompletionStage<? extends T> compl
...
+	}
+
+	/**
+	 * Create a {@link Mono} that wraps a {@link CompletionStage} on subscription,
+	 * emitting the value produced by the {@link CompletionStage}.
+	 *
+	 * <p>
+	 * <img class="marble" src="doc-files/marbles/fromFutureSupplier.svg" alt="">
+	 * <p>
+	 * Note, use {@link #fromFuture(CompletableFuture, boolean)} with {@code
```
violetagg, line 603:
```suggestion
	 * Note, use {@link #fromCompletionStage(CompletableFuture, boolean)} with {@code
```

violetagg (changes requested):

simonbasle (changes requested):
as discussed I don't think this is the right course of action since that parameter doesn't make sense for _all_ possible inputs (namely, `CompletionStage` instances that don't implement `Future`)

chemicL:
Closing. It does make sense when the `CompletionStage` also implements `Future`, but does not support conversion to `CompletableFuture` (in which case the `fromFuture` operator could be used). But we can come back to this PR and improve the javadoc, which is inaccurate at the moment, if users run into this problem and want to disable cancellation propagation.

PR #3282, opened 2022-11-15
Add migration notes to reference guide

simonbasle:
- polish: add anchor for observation, underline H5
- Add migration notes appendix with 3.5 migration notes


simonbasle:
@reactor/team this is a first draft at more fleshed out migration notes from 3.4.x to `3.5.0`.

I'm not sure about adding this to the reference guide. It could as well be a wiki page, which we can reference from the guide and the release notes.

wdyt?

chemicL:
This is long overdue. As I believe users will still find this information available, I published this as a wiki: https://github.com/reactor/reactor-core/wiki/Migration-Guides.

PR #3283, opened 2022-11-16
Update retryWhen docs on deprecated currentContext

cbrunsdon:
This should now be pointing to contextView as currentContext is deprecated

@pivotal-cla This is an Obvious Fix


simonbasle (changes requested):
nice catch ! I think there are other places where a similar change could be implemented, notably the `FluxSink` and `MonoSink` javadocs deprecation notice for `currentContext()`.

also, the notices and deprecations have been put as early as the `3.4.x` branch. would you mind retargetting your PR to that branch instead? 🙇

cbrunsdon:
Retargetted to 4.3 and picked up the other two instances, thanks.

Couldn't find anything else wrongly referencing `getContextView`, all remaining instances (on the 4.3.x branch anyway) are for the `Signal` method.

OlegDokuka (approved):

OlegDokuka:
@cbrunsdon great work! Thanks for your improvements!

PR #3288, opened 2022-11-17
Drop-in replacement for enableMetrics and disableMetrics

### LINKED ISSUE #3285: No way to meter Schedulers used internally by Reactor
Either I'm missing something or by deprecating `Schedulers.enableMetrics()` there is a regression in functionality.

When I played with https://github.com/reactor/reactor-monitoring-demo I noticed that it was possible to meter Schedulers that were used internally by Reactor eg:

```java
	public static Flux<Long> interval(Duration period) {
		return interval(period, Schedulers.parallel());
	}
```

or `boundedElastic` Scheduler that was used to handle incoming requests from netty.

Currently, there is no way to get metrics for such schedulers because I cannot wrap them in `TimedScheduler`.

kkocel:
fixes #3285

### DIFF
--- a/reactor-core-micrometer/src/main/java/reactor/core/observability/micrometer/Micrometer.java
+++ b/reactor-core-micrometer/src/main/java/reactor/core/observability/micrometer/Micrometer.java
@@ -132,7 +132,8 @@ public final class Micrometer {
 	}
 
     /**
-     * Install a {@link Schedulers.Factory} with metrics support.
+     * Install a {@link Schedulers.Factory} with metrics support. All new schedulers will be instrumented with
+	 * the provided {@link MeterRegistry} and naming meters using the provided {@code metricsPrefix}.
      *
      * @param metricsPrefix the prefix to use in meter names. Must not end with a dot, which is automatically added.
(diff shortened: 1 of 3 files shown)


simonbasle (changes requested):
as described in the issue, the approach is a bit hacky: I think the ability to distinguish between individual schedulers is important to avoid aggregation issues and worse (like some meter overlap), but the way we obtain a "name" from the `ThreadFactory` isn't really officially supported.

and beyond that, I strongly feel that instrumenting _all_ schedulers at once like that is a mistake, or at least something that should be a very thought out decision from the user, and that as a result Reactor shouldn't offer "the easy way" of such a `Factory`

kkocel:
As a user, I would like to enable metrics easily for schedulers as I used to do with `enableMetrics()` and most users want to do just that, get Schedulers metered. 

Having them create their own `Schedulers.Factory`, learn about `timedScheduler`, and wiring this all together you 
expose them to a lot of Reactor internals (which should not be required) and give them a steep learning curve. 

I am aware of all the pitfalls you described but please allow users do to common and simple tasks easily. And I strongly believe that enabling monitoring for schedulers is such task.

PR #3297, opened 2022-11-24
ensures given iterable is interacted once per subscriber

### LINKED ISSUE #3295: FluxIterable may call iterator twice, that may break constrainOnce contract of Kotlin Sequence
FluxIterable may call `iterator` method twice. I don't think it is forbidden, but it seams a poor practice because of performance and/or possible side effects. Use with Kotlin Sequence, it may cause an IllegalStateException: This sequence can be consumed only once.

## Expected Behavior

The following code throws an IllegalStateException. I expect the flux is consumed without any exception.

    listOf('1', '2', '3').iterator().asSequence().toFlux().blockLast()

## Actual Behavior

Now, an  IllegalStateException is thrown with the message: "This sequence can be consumed only once".

## Steps to Reproduce
This Kotlin code fails systematically with an exception: `IllegalStateException: This sequence can be consumed only once`.

```java
@Test
void reproCase() {
    // This should not throw exception. But it does...
    listOf('1', '2', '3').iterator().asSequence().toFlux().blockLast()
}
```

In my understanding Kotlin asSequence extension puts a constrainOnce contract:

    public fun <T> Iterator<T>.asSequence(): Sequence<T> = Sequence { this }.constrainOnce()

In FluxExtensions.kt, we have:

    fun <T : Any> Sequence<T>.toFlux(): Flux<T> = Flux.fromIterable(object : Iterable<T> {
        override fun iterator(): Iterator<T> = this@toFlux.iterator()
    })

And in FluxIterable, we have:

    static <T> boolean checkFinite(Iterable<T> iterable) {
        return iterable instanceof Collection || iterable.spliterator().hasCharacteristics(Spliterator.SIZED);
    }
    public void subscribe(/* ... */) {
        /* ... */
        knownToBeFinite = FluxIterable.checkFinite(iterable);
        it = iterable.iterator();
        /* ... */
    }

Moreover, Kotlin `Sequence` does not overload the default implementation of `spliterator()`, that's why `iterator` method is called twice in my case. So the constrainOnce is broken.

## Possible Solution

I think `FluxIterable` should call the `iterator` method only once. A possible solution is to always build a `Spliterator` to both iterate and test `Characteristics`.

If you prefer to keep `FluxIterable` unchanged, then I suggest to change the implementation of `Sequence<T>.toFlux()` extension, to rely on `Flux.fromStream` instead of `Flux.fromIterable`.

## Your Environment

* Reactor version(s) used: 3.4.6 or 3.5.0
* Other relevant libraries versions (eg. `netty`, ...): Kotlin-extension 1.1.8 or 1.2.0
* JVM version (`java -version`): Java 8 or 11.

OlegDokuka:
close #3295

### DIFF
--- a/reactor-core/src/main/java/reactor/core/publisher/FluxIterable.java
+++ b/reactor-core/src/main/java/reactor/core/publisher/FluxIterable.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2016-2021 VMware Inc. or its affiliates, All Rights Reserved.
+ * Copyright (c) 2016-2022 VMware Inc. or its affiliates, All Rights Reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -16,11 +16,9 @@
 
 package reactor.core.publisher;
 
-import java.util.Collection;
-import java.util.Iterator;
-import java.util.Objects;
-import java.util.Spliterator;
+import java.util.*;
 import java.util.concurrent.atomic.AtomicLongFieldUpdater;
+import java.util.function.Consumer;
 
 import org.reactivestreams.Subscriber;
 
@@ -49,12 +47,12 @@ final class FluxIterable<T> extends Flux<T> implements Fuseable, SourceProducer<
 	 * A {@link Collection} is assumed to be finite, and for other iterables the {@link Spliterator}
 	 * {@link Spliterator#SIZED} characteristic is looked for.
 	 *
-	 * @param iterable the {@link Iterable} to check.
+	 * @param spliterator the {@link Iterable} to check.
 	 * @param <T>
 	 * @return true if the {@link Iterable} can confidently classified as finite, false if not finite/unsure
 	 */
-	static <T> boolean checkFinite(Iterable<T> iterable) {
-		return iterable instanceof Collection || iterable.spliterator().hasCharacteristics(Spliterator.SIZED);
+	static <T> boolean checkFinite(Spliterator<T> spliterator) {
+		return spliterator.hasCharacteristics(Spliterator.SIZED);
 	}
 
 	final Iterable<? extends T> iterable;
@@ -73,18 +71,18 @@ final class FluxIterable<T> extends Flux<T> implements Fuseable, SourceProducer<
 	@Override
 	public void subscribe(CoreSubscriber<? super T> actual) {
 		boolean knownToBeFinite;
-		Iterator<? extends T> it;
+		Spliterator<? extends T> sp;
 
 		try {
-			knownToBeFinite = FluxIterable.checkFinite(iterable);
-			it = iterable.iterator();
+			sp = iterable.spliterator();
+			knownToBeFinite = FluxIterable.checkFinite(sp);
 		}
 		catch (Throwable e) {
 			Operators.error(actual, Operators.onOperatorError(e, actual.currentContext()));
 			return;
 		}
 
-		subscribe(actual, it, knownToBeFinite, onClose);
+		subscribe(actual, sp, knownToBeFinite, onClose);
 	}
 
 	@Override
@@ -103,10 +101,10 @@ final class FluxIterable<T> extends Flux<T> implements Fuseable, SourceProducer<
 	 * Common method to take an {@link Iterator} as a source of values.
 	 *
 	 * @param s the subscriber to feed this iterator to
-	 * @param it the {@link Iterator} to use as a predictable source of values
+	 * @param sp the {@link Iterator} to use as a predictable source of values
 	 */
-	static <T> void subscribe(CoreSubscriber<? super T> s, Iterator<? extends T> it, boolean knownToBeFinite) {
-		subscribe(s, it, knownToBeFinite, null);
+	static <T> void subscribe(CoreSubscriber<? super T> s, Spliterator<? extends T> sp, boolean knownToBeFinite) {
+		subscribe(s, sp, knownToBeFinite, null);
 	}
 
 	/**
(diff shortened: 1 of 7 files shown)


reactor-core/src/main/java/reactor/core/publisher/Operators.java
```diff
@@ -618,6 +618,30 @@ public static void onDiscardMultiple(@Nullable Iterator<?> multiple, boolean kno
...
+							hook.accept(o);
+						}
+						catch (Throwable t) {
+							log.warn("Error while discarding element from an Iterator, continuing with next element", t);
+						}
+					}
+				});
+			}
+			catch (Throwable t) {
+				log.warn("Error while discarding Iterator, stopping", t);
```
chemicL, line 640: Consider Splitterator to make debugging easier.

reactor-core/src/main/java/reactor/core/publisher/FluxIterable.java
```diff
@@ -49,12 +47,12 @@
 	 * A {@link Collection} is assumed to be finite, and for other iterables the {@link Spliterator}
 	 * {@link Spliterator#SIZED} characteristic is looked for.
 	 *
-	 * @param iterable the {@link Iterable} to check.
+	 * @param spliterator the {@link Iterable} to check.
```
chemicL, line 50: Iterable -> Spliterator

reactor-core/src/main/java/reactor/core/publisher/FluxIterable.java
```diff
@@ -103,10 +101,10 @@ public Object scanUnsafe(Attr key) {
 	 * Common method to take an {@link Iterator} as a source of values.
 	 *
 	 * @param s the subscriber to feed this iterator to
-	 * @param it the {@link Iterator} to use as a predictable source of values
+	 * @param sp the {@link Iterator} to use as a predictable source of values
```
chemicL, line 104: Iterator -> Spliterator; also in other javadocs below

reactor-core/src/main/java/reactor/core/publisher/FluxIterable.java
```diff
@@ -119,18 +117,18 @@ static <T> void subscribe(CoreSubscriber<? super T> s, Iterator<? extends T> it,
...
 			Operators.error(s, new NullPointerException("The iterator is null"));
 			return;
 		}
 
-		boolean b;
+		long size;
 
 		try {
-			b = it.hasNext();
+			size = knownToBeFinite ? it.estimateSize() : -1;
```
chemicL, line 131: I'd love to see a comment explaining why this is feasible, e.g. "If the Spliterator is finite and we know it has not been traversed, the estimate is accurate." - but this raises a question - should we check whether the Spliterator has not been traversed yet? Is there a way to do so?

chemicL (approved):

reactor-core/src/main/java/reactor/core/publisher/FluxIterable.java
chemicL, line 50:
```suggestion
	 * @param spliterator the {@link Spliterator} to check.
```

reactor-core/src/main/java/reactor/core/publisher/Operators.java
chemicL, line 640:
```suggestion
				log.warn("Error while discarding Spliterator, stopping", t);
```

reactor-core/src/main/java/reactor/core/publisher/FluxIterable.java
OlegDokuka: I think it is irrelevant. At this stage, we just retrieved a new spliterator instance. So does it make sense to care about "check whether the Spliterator has not been traversed yet"? 

Also, the code it selves is taken from the method `Spliterator#getExactSizeIfKnown()` so this avoids extra work since we `knowToBeFinite`

reactor-core/src/main/java/reactor/core/publisher/FluxIterable.java
OlegDokuka: fixed

reactor-core/src/main/java/reactor/core/publisher/FluxIterable.java
OlegDokuka: fixed

reactor-core/src/main/java/reactor/core/publisher/Operators.java

reactor-core/src/main/java/reactor/core/publisher/FluxIterable.java
chemicL: Right, I failed to record the fact the spliterator was retrieved in the constructor. Makes me think though: what happens in case of -1?

OlegDokuka: it is not used explicitly. I tend to think that indeed we need to replace that with simplified code. e.g. 

```java
hasValues = !knownToBeFinite || it.estimateSize() != 0
```

OlegDokuka: using the above we can say: ok if it is unknown to be finite - we can do nothing, assume has values, if it is known to be finite and estimated size is not 0 then it has values

chemicL: Sounds good, thanks.

OlegDokuka:
removed `instanceof Collection` check in `checkFinite` method

instanceof check brings significant overhead. All JMH benchmarks with observations are available at the following link https://gist.github.com/OlegDokuka/2a5dc81ba89d1f3b8f81153a78a4dd96

chemicL (approved):

PR #3312, opened 2022-12-08
fixes #3307 case

### LINKED ISSUE #3307: Flux#next throws NoSuchElementException on a non-finite empty iterator
Spring Security's [`R2dbcReactiveOAuth2AuthorizedClientServiceTests`](https://github.com/spring-projects/spring-security/blob/main/oauth2/oauth2-client/src/test/java/org/springframework/security/oauth2/client/R2dbcReactiveOAuth2AuthorizedClientServiceTests.java) [started failing](https://ge.spring.io/s/ujpedsru73sh4/tests/:spring-security-oauth2-client:test/org.springframework.security.oauth2.client.R2dbcReactiveOAuth2AuthorizedClientServiceTests/loadAuthorizedClientWhenExistsThenReturnAuthorizedClient()?top-execution=1) on the `3.5.1-SNAPSHOT` recently.

It appears to coincide with the changes in this commit (67cc87ab7e32c4163f7097c4ab1fc35f9b1ee4b0).

Here is a quick test that fails that is a minimal representation of what the Spring Security tests are doing:

```java
@Test
public void testFluxIterable() {
    Iterable<String> iterable = mock(Iterable.class);
    given(iterable.spliterator()).willReturn(mock(Spliterator.class));
    Flux.fromIterable(iterable).next().block();
}
```

The Javadoc for `Flux#next` says

> Emit only the first item emitted by this Flux, into a new Mono. If called on an empty Flux, emits an empty Mono.

I would expect the above test to pass, but it instead throws a `NoSuchElementException`.

OlegDokuka:
closes #3307 

### DIFF
--- a/reactor-core/src/main/java/reactor/core/publisher/FluxIterable.java
+++ b/reactor-core/src/main/java/reactor/core/publisher/FluxIterable.java
@@ -158,11 +158,51 @@ final class FluxIterable<T> extends Flux<T> implements Fuseable, SourceProducer<
 		}
 
 		if (s instanceof ConditionalSubscriber) {
-			s.onSubscribe(new IterableSubscriptionConditional<>((ConditionalSubscriber<? super T>) s,
-					sp, knownToBeFinite, onClose));
+			IterableSubscriptionConditional<? extends T> isc =
+					new IterableSubscriptionConditional<>((ConditionalSubscriber<? super T>) s,
+							sp,
+							knownToBeFinite,
+							onClose);
+
+			boolean hasNext;
+			try {
+				hasNext = isc.hasNext();
+			}
+			catch (Throwable ex) {
+				Operators.error(s, ex);
+				isc.onCloseWithDropError();
+				return;
+			}
+
+			if (!hasNext) {
+				Operators.complete(s);
+				isc.onCloseWithDropError();
+				return;
+			}
+
+			s.onSubscribe(isc);
 		}
 		else {
-			s.onSubscribe(new IterableSubscription<>(s, sp, knownToBeFinite, onClose));
+			IterableSubscription<? extends T> is =
+					new IterableSubscription<>(s, sp, knownToBeFinite, onClose);
+
+			boolean hasNext;
+			try {
+				hasNext = is.hasNext();
+			}
+			catch (Throwable ex) {
+				Operators.error(s, ex);
+				is.onCloseWithDropError();
+				return;
+			}
+
+			if (!hasNext) {
+				Operators.complete(s);
+				is.onCloseWithDropError();
+				return;
+			}
+
+			s.onSubscribe(is);
 		}
 	}
 
@@ -183,7 +223,7 @@ final class FluxIterable<T> extends Flux<T> implements Fuseable, SourceProducer<
 				AtomicLongFieldUpdater.newUpdater(IterableSubscription.class,
 						"requested");
 
-		int state = STATE_CALL_HAS_NEXT;
+		int state;
 
 		/**
 		 * Indicates that the iterator's hasNext returned true before but the value is not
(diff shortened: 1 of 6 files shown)


reactor-core/src/test/java/reactor/core/publisher/FluxIterableTest.java
```diff
@@ -44,11 +50,104 @@
...
+		    .expectSubscription()
+		    .expectComplete()
+		    .verify();
+	}
+
+	@Test
+	public void testFluxIterableSlowPath() {
+		Iterable<String> iterable = mock(Iterable.class);
+		Mockito.when(iterable.spliterator()).thenReturn(mock(Spliterator.class));
+		Flux.fromIterable(iterable).limitRate(1).next().block();
```
chemicL, line 148: It looks like this case is covered in the factory.

reactor-core/src/test/java/reactor/core/publisher/FluxIterableTest.java
```diff
@@ -44,11 +50,104 @@
 
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.assertThatExceptionOfType;
+import static org.mockito.Mockito.mock;
 
 public class FluxIterableTest {
 
 	final Iterable<Integer> source = Arrays.asList(1, 2, 3, 4, 5, 6, 7, 8, 9, 10);
 
+	static Stream<Function<Flux, Flux>> factory() {
```
chemicL, line 59: Consider not duplicating the code present in both this file and the flatten scenario. Otherwise, best to add a note as a comment that these should be kept in sync upon change.

reactor-core/src/main/java/reactor/core/publisher/FluxIterable.java
```diff
@@ -296,16 +299,17 @@ void slowPath(long n) {
 						return;
 					}
 
-					s.onNext(t);
-
-					if (cancelled) {
+					if (!b) {
+						s.onComplete();
+						onCloseWithDropError();
```
chemicL, line 304: I just noticed onClose is never actually provided. Maybe it's worth creating an issue to remove the implementation for it? It looks like the constructor accepting onClose is an internal one, so this wouldn't be a breaking change.

chemicL (changes requested):

reactor-core/src/test/java/reactor/core/publisher/FluxIterableTest.java
OlegDokuka: right. duplication. fixed

reactor-core/src/main/java/reactor/core/publisher/FluxIterable.java
OlegDokuka: it is used by FluxStream

chemicL: Oh. `FluxIterable#subscribe` with `onClose` is a static method used by `FluxStream`. I stand corrected, I only looked at constructors of `FluxIterable`.

chemicL (approved):

chemicL (approved):

reactor-core/src/main/java/reactor/core/publisher/FluxIterable.java
```diff
@@ -158,11 +154,51 @@ static <T> void subscribe(CoreSubscriber<? super T> s, Spliterator<? extends T>
...
+							knownToBeFinite,
+							onClose);
+
+			boolean hasNext;
+			try {
+				hasNext = isc.hasNext();
+			}
+			catch (Throwable ex) {
+				Operators.error(s, ex);
+				isc.onCloseWithDropError();
```
chemicL, line 169: is it expected even before onSubscribe?

chemicL (approved):

reactor-core/src/main/java/reactor/core/publisher/FluxIterable.java
OlegDokuka: look at `Operators.error(s, ex);` it looks like the following 

```java
s.onSubscribe(EmptySubscription.INSTANCE);
s.onError(e);
```

so it is after

chemicL: Got it, thanks.

OlegDokuka: in that case, instantiated IterableSubscription is rather an existing thing that can close the resource instead of writing  

```java
             if (onClose != null) {
				try {
					onClose.run();
				}
				catch (Throwable t) {
					Operators.onErrorDropped(t, s.currentContext());
				}
			}
			```

PR #3374, opened 2023-03-06
Exclude `ContextPropagation#isContextPropagationAvailable` from the preprocessing for `native-image` support

violetagg:
Reflection is not used for identifying whether this library is available or not, thus this field can be excluded from the preprocessing for `native-image` support.

Related to https://github.com/spring-projects/spring-framework/issues/30058

### DIFF
--- a/reactor-core/src/main/java/reactor/core/publisher/ContextPropagation.java
+++ b/reactor-core/src/main/java/reactor/core/publisher/ContextPropagation.java
@@ -49,7 +49,8 @@ final class ContextPropagation {
 
 	static final Logger LOGGER;
 
-	static final boolean isContextPropagationAvailable;
+	// Note: If a reflection is used for this field, then the name of the field should end with 'Available'
+	static final boolean isContextPropagationOnClasspath;
 	static boolean propagateContextToThreadLocals = false;
 
 	static final Predicate<Object> PREDICATE_TRUE = v -> true;
@@ -80,7 +81,7 @@ final class ContextPropagation {
 					" The feature is considered disabled due to this:", t);
 		}
 
-		isContextPropagationAvailable = contextPropagation;
+		isContextPropagationOnClasspath = contextPropagation;
 		WITH_GLOBAL_REGISTRY_NO_PREDICATE = contextCaptureFunction;
 	}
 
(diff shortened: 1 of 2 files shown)


reactor-core/src/main/java/reactor/core/publisher/ContextPropagation.java
```diff
@@ -49,7 +49,8 @@ final class ContextPropagation {
 
 	static final Logger LOGGER;
 
-	static final boolean isContextPropagationAvailable;
+	// Note: If a reflection is used for this field, then the name of the field should end with 'Available'
```
chemicL, line 52:
```suggestion
	// Note: If reflection is used for this field, then the name of the field should end with 'Available'
```

chemicL: It would be also great to have an answer to "why" - is this processing based on variable name suffix documented anywhere?

chemicL:
Thanks @violetagg ! Just a minor documentation request and let's 🚢

reactor-core/src/main/java/reactor/core/publisher/ContextPropagation.java
violetagg: No this is not documented and the functionality is located in Spring Framework.
The functionality in Spring Framework, related to Reactor projects, is a short term solution. Ultimately the long term solution should be in each Reactor project that needs such.

chemicL (approved):

reactor-core/src/main/java/reactor/core/publisher/ContextPropagation.java
chemicL: Thanks for the clarification and adding the comment!

reactor-core/src/main/java/reactor/core/publisher/ContextPropagation.java
```diff
@@ -49,7 +49,11 @@ final class ContextPropagation {
 
 	static final Logger LOGGER;
 
-	static final boolean isContextPropagationAvailable;
+	// Note: If reflection is used for this field, then the name of the field should end with 'Available'.
+	// The preprocessing for native-image support is Spring Framework, and is a short term solution.
+	// The field should end with 'Available'. See org.springframework.aot.nativex.feature.PreComputeFieldFeature.
```
eduanb, line 54: I think this comment is wrong. The PR changes the field name to not end with `Availible` however, the comment says it should end with `Availible`.

chemicL: The comment is fine. It begins with "if reflection is used" and currently reflection is NOT used. So if we introduce reflection here, we can rename it to end with "Available".

violetagg:
thanks for the review

PR #3405, opened 2023-03-16
Propagate ThreadLocals in tap

### LINKED ISSUE #3395: Micrometer.observation failure with automatic context propagation
Context is not properly propagated through publishers when using `Hooks.enableAutomaticContextPropagation();`. I've created a simple [spring boot 3 application](https://github.com/maciej-gromul/boot-3-problems-with-context-propagation) that should be working using micrometer observations to create tracing context. I can see spans being created but their context is not propagated which results in something like this:

image

while expected trace should be nested like this:

```
http get /trace
  |
  -- some-random-name
       |
       -- simple-text-wrapping-service
            |
            -- other-layer
```

I've confirmed that TapSubscriber properly registers in reactor context new tracing context from created span, but it get's lost somewhere and in the next tap operator that creates a new observation with new span later on it always attaches to the root span.

## Expected Behavior
Context get's propagated and spans are being properly connected to each other.

## Actual Behavior
Root span created by spring webflux is always being treated as parent root, even if we create new span following the migration guide for sleuth, it get's created but parent stays as current span/context.

## Steps to Reproduce
Here's the project with example to reproduce the problem: https://github.com/maciej-gromul/boot-3-problems-with-context-propagation. There's docker compose configuration file to run grafana stack allowing easy verification of generated tracing.

## Possible Solution
Don't know what precisely causes this.

chemicL:
Aside from `contextWrite()`, the tap operator also has write access to the `Context`.
New classes have been added to handle the automatic context propagation:

* `MonoTapRestoringThreadLocals`
* `FluxTapRestoringThreadLocals`

They replace the corresponding operators in tap when `Hooks.enableAutomaticContextPropagation()` is called, and when the `context-propagation` library is on the class path.

As a result of these changes, `FluxTapTest` was also moved to the `withMicrometerTest` source set, where the `context-propagation` library is present.

Fixes #3395.

### DIFF
--- a/reactor-core-micrometer/src/test/java/reactor/core/observability/micrometer/MicrometerTest.java
+++ b/reactor-core-micrometer/src/test/java/reactor/core/observability/micrometer/MicrometerTest.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2022 VMware Inc. or its affiliates, All Rights Reserved.
+ * Copyright (c) 2022-2023 VMware Inc. or its affiliates, All Rights Reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -25,6 +25,7 @@ import io.micrometer.observation.contextpropagation.ObservationThreadLocalAccess
 import org.junit.jupiter.api.Test;
 import org.mockito.Mockito;
 
+import reactor.core.publisher.Hooks;
 import reactor.core.scheduler.Scheduler;
 
 import static org.assertj.core.api.Assertions.assertThat;
--- a/reactor-core-micrometer/src/test/java/reactor/core/observability/micrometer/MicrometerTest.java
+++ b/reactor-core-micrometer/src/test/java/reactor/core/observability/micrometer/MicrometerTest.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2022-2022 VMware Inc. or its affiliates, All Rights Reserved.
+ * Copyright (c) 2022 VMware Inc. or its affiliates, All Rights Reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/reactor-core-micrometer/src/test/java/reactor/core/observability/micrometer/MicrometerTest.java
+++ b/reactor-core-micrometer/src/test/java/reactor/core/observability/micrometer/MicrometerTest.java
@@ -25,7 +25,6 @@ import io.micrometer.observation.contextpropagation.ObservationThreadLocalAccess
 import org.junit.jupiter.api.Test;
 import org.mockito.Mockito;
 
-import reactor.core.publisher.Hooks;
 import reactor.core.scheduler.Scheduler;
 
 import static org.assertj.core.api.Assertions.assertThat;
--- a/reactor-core-micrometer/src/test/java/reactor/core/observability/micrometer/MicrometerTest.java
+++ b/reactor-core-micrometer/src/test/java/reactor/core/observability/micrometer/MicrometerTest.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2022-2023 VMware Inc. or its affiliates, All Rights Reserved.
+ * Copyright (c) 2022-2022 VMware Inc. or its affiliates, All Rights Reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/reactor-core/src/main/java/reactor/core/publisher/ContextPropagation.java
+++ b/reactor-core/src/main/java/reactor/core/publisher/ContextPropagation.java
@@ -150,6 +150,17 @@ final class ContextPropagation {
 			.updateContext(target);
 	}
 
+	/**
+	 * When <a href="https://github.com/micrometer-metrics/context-propagation">context-propagation library</a>
+	 * is available on the classpath, the provided {@link BiConsumer handler} will be
+	 * called with {@link ThreadLocal} values restored from the provided {@link Context}.
+	 * @param handler user provided handler
+	 * @param contextSupplier supplies the potentially modified {@link Context} to
+	 *                           restore {@link ThreadLocal} values from
+	 * @return potentially wrapped {@link BiConsumer} or the original
+	 * @param <T> type of handled values
+	 * @param <R> the transformed type
+	 */
 	static <T, R> BiConsumer<T, SynchronousSink<R>> contextRestoreForHandle(BiConsumer<T, SynchronousSink<R>> handler, Supplier<Context> contextSupplier) {
 		if (propagateContextToThreadLocals || !ContextPropagation.isContextPropagationAvailable()) {
 			return handler;
@@ -165,6 +176,21 @@ final class ContextPropagation {
 		};
 	}
 
+	/**
+	 * When <a href="https://github.com/micrometer-metrics/context-propagation">context-propagation library</a>
+	 * is available on the classpath, the provided {@link SignalListener} will be wrapped
+	 * with another one that restores {@link ThreadLocal} values from the provided
+	 * {@link Context}.
+	 * <p><strong>Note, this is only applied to {@link FluxTap}, {@link FluxTapFuseable},
+	 * {@link MonoTap}, and {@link MonoTapFuseable}.</strong> The automatic propagation
+	 * variants: {@link FluxTapRestoringThreadLocals} and
+	 * {@link MonoTapRestoringThreadLocals} do not use this method.
+	 * @param original the original {@link SignalListener} from the user
+	 * @param contextSupplier supplies the potentially modified {@link Context} to
+	 *                           restore {@link ThreadLocal} values from
+	 * @return potentially wrapped {@link SignalListener} or the original
+	 * @param <T> type of handled values
+	 */
 	static <T> SignalListener<T> contextRestoreForTap(final SignalListener<T> original, Supplier<Context> contextSupplier) {
 		if (!ContextPropagation.isContextPropagationAvailable()) {
 			return original;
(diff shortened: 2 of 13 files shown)


reactor-core/src/main/java/reactor/core/publisher/ContextPropagation.java
```diff
@@ -166,7 +166,7 @@ static <T, R> BiConsumer<T, SynchronousSink<R>> contextRestoreForHandle(BiConsum
 	}
 
 	static <T> SignalListener<T> contextRestoreForTap(final SignalListener<T> original, Supplier<Context> contextSupplier) {
-		if (propagateContextToThreadLocals || !ContextPropagation.isContextPropagationAvailable()) {
+		if (!ContextPropagation.isContextPropagationAvailable()) {
```
OlegDokuka, line 169: Any reasons why this is changed?

reactor-core/src/main/java/reactor/core/publisher/FluxTapRestoringThreadLocals.java
```diff
@@ -0,0 +1,353 @@
...
+		/**
+		 * Cancel the active subscription, pass the listener error to {@link SignalListener#handleListenerError(Throwable)}
+		 * and then terminate the downstream directly with same error (without invoking any other handler).
+		 *
+		 * @param listenerError the exception thrown from a handler method
+		 */
+		protected void handleListenerErrorAndTerminate(Throwable listenerError) {
+			s.cancel();
+			listener.handleListenerError(listenerError);
+			actual.onError(listenerError); //TODO wrap ? hooks ?
```
OlegDokuka, line 156: decorated with scope?

reactor-core/src/main/java/reactor/core/publisher/FluxTapRestoringThreadLocals.java
```diff
@@ -0,0 +1,353 @@
...
+		 * Cancel the prepared subscription, pass the listener error to {@link SignalListener#handleListenerError(Throwable)}
+		 * and then terminate the downstream directly with same error (without invoking any other handler).
+		 *
+		 * @param listenerError the exception thrown from a handler method before the subscription was set
+		 * @param toCancel      the {@link Subscription} that was prepared but not sent downstream
+		 */
+		protected void handleListenerErrorPreSubscription(Throwable listenerError, Subscription toCancel) {
+			toCancel.cancel();
+			listener.handleListenerError(listenerError);
+			Operators.error(actual, listenerError);
```
OlegDokuka, line 144: scope?

reactor-core/src/main/java/reactor/core/publisher/ContextPropagation.java
OlegDokuka: how about adding comment that this method is used in non-context restoring versions of tap operator

chemicL: yep, will do

reactor-core/src/main/java/reactor/core/publisher/FluxTapRestoringThreadLocals.java
```diff
@@ -0,0 +1,353 @@
...
+		 * (without invoking any other handler).
+		 *
+		 * @param listenerError the exception thrown from a handler method
+		 * @param originalError the exception that was about to occur when handler was invoked
+		 */
+		protected void handleListenerErrorMultipleAndTerminate(Throwable listenerError, Throwable originalError) {
+			s.cancel();
+			listener.handleListenerError(listenerError);
+			RuntimeException multiple = Exceptions.multiple(listenerError, originalError);
+			actual.onError(multiple); //TODO wrap ? hooks ?
```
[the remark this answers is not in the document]
chemicL: Right, good eye. I fixed the three you noticed and also found 2 more in the subscribe method, so together with the Mono counterpart that makes 4 more places to add. Will follow up shortly.

reactor-core/src/main/java/reactor/core/publisher/MonoTapRestoringThreadLocals.java
```diff
@@ -0,0 +1,95 @@
...
+		//if the SignalListener cannot be created, all we can do is error the subscriber.
+		//after it is created, in case doFirst fails we can additionally try to invoke doFinally.
+		//note that if the later handler also fails, then that exception is thrown.
+		SignalListener<T> signalListener;
+		try {
+			//TODO replace currentContext() with contextView() when available
+			signalListener = tapFactory.createListener(source, actual.currentContext().readOnly(), commonTapState);
+		}
+		catch (Throwable generatorError) {
+			try (ContextSnapshot.Scope ignored =
```
OlegDokuka, line 53: i guess this is redundant since we don't have new TL set

reactor-core/src/main/java/reactor/core/publisher/FluxTapRestoringThreadLocals.java
```diff
@@ -0,0 +1,368 @@
...
+		//if the SignalListener cannot be created, all we can do is error the subscriber.
+		//after it is created, in case doFirst fails we can additionally try to invoke doFinally.
+		//note that if the later handler also fails, then that exception is thrown.
+		SignalListener<T> signalListener;
+		try {
+			//TODO replace currentContext() with contextView() when available
+			signalListener = tapFactory.createListener(source, actual.currentContext().readOnly(), commonTapState);
+		}
+		catch (Throwable generatorError) {
+			try (ContextSnapshot.Scope ignored =
```
OlegDokuka, line 56: this is redundant here since there is no new TL restored yet

reactor-core/src/main/java/reactor/core/publisher/MonoTapRestoringThreadLocals.java
```diff
@@ -0,0 +1,95 @@
...
+		}
+
+		// invoked AFTER doFirst
+		Context alteredContext;
+		try {
+			alteredContext = signalListener.addToContext(actual.currentContext());
+		}
+		catch (Throwable e) {
+			signalListener.handleListenerError(new IllegalStateException("Unable to augment tap Context at construction via addToContext", e));
+			try (ContextSnapshot.Scope ignored =
```
OlegDokuka, line 76: any reasons for doing that?

reactor-core/src/main/java/reactor/core/publisher/MonoTapRestoringThreadLocals.java
```diff
@@ -0,0 +1,95 @@
...
+					     ContextPropagation.setThreadLocals(actual.currentContext())) {
+				alteredContext = actual.currentContext();
+			}
```
OlegDokuka, line 79: interesting. the return statement is missing and likely a bug in both existing implementations.

reactor-core/src/main/java/reactor/core/publisher/MonoTapRestoringThreadLocals.java
OlegDokuka: I mean we dont have new TL set yet

reactor-core/src/main/java/reactor/core/publisher/FluxTapRestoringThreadLocals.java
```diff
@@ -0,0 +1,368 @@
...
+		}
+
+		// invoked AFTER doFirst
+		Context alteredContext;
+		try {
+			alteredContext = signalListener.addToContext(actual.currentContext());
+		}
+		catch (Throwable e) {
+			signalListener.handleListenerError(new IllegalStateException("Unable to augment tap Context at construction via addToContext", e));
+			alteredContext = actual.currentContext();
```
OlegDokuka, line 82: likely there should be 
`Operators.error(actual, listenerError);` and `return;` seems to be a bug in both existing implementations

reactor-core/src/main/java/reactor/core/publisher/FluxTapRestoringThreadLocals.java
chemicL: my bad

reactor-core/src/main/java/reactor/core/publisher/FluxTapRestoringThreadLocals.java
```diff
@@ -0,0 +1,368 @@
...
+			}
+			return;
+		}
+
+		try {
+			signalListener.doFirst();
+		}
+		catch (Throwable listenerError) {
+			signalListener.handleListenerError(listenerError);
+			try (ContextSnapshot.Scope ignored =
```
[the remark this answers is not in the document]
chemicL: right, my mistake

reactor-core/src/main/java/reactor/core/publisher/FluxTapRestoringThreadLocals.java
chemicL: I don't know, I suppose. But let's not make the fix part of this PR, but a follow up.

reactor-core/src/main/java/reactor/core/publisher/MonoTapRestoringThreadLocals.java
chemicL: That must have been a mistake, `currentContext()` doesn't need TLs to be present at all, removing.

reactor-core/src/main/java/reactor/core/publisher/FluxTapRestoringThreadLocals.java
chemicL: I created #3414 to address this concern.

reactor-core-micrometer/src/test/java/reactor/core/observability/micrometer/MicrometerTest.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2022 VMware Inc. or its affiliates, All Rights Reserved.
+ * Copyright (c) 2022-2023 VMware Inc. or its affiliates, All Rights Reserved.
```
OlegDokuka, line 2: redundant change?

OlegDokuka (approved):

Issue #3340, opened 2023-01-31
`Sinks.many().replay().latest()` holds two values in cache


MikkelHJuul:
The caching used for `Sinks.many().replay().latest()` holds both `head` and `tail` in the cache, although only one is available for replay (as it should). This means that the extra reference holds unnecessary memory, and holds onto a reference that cannot be reclaimed.

## Expected Behavior
The cache holds only a single reference in memory.

## Actual Behavior
The  https://github.com/reactor/reactor-core/blob/main/reactor-core/src/main/java/reactor/core/publisher/FluxReplay.java#L774 SizeBoundReplayBuffer holds two items in cache.

## Steps to Reproduce

```java
package reactor.core.publisher;

@Test
void reproCase() {
     var replayBuffer = new FluxReplay.SizeBoundReplayBuffer<Integer>(1);
     replayBuffer.add(11);
     replayBuffer.add(22);
     assert replayBuffer.head != replayBuffer.tail;
     assert replayBuffer.head.value != replayBuffer.tail.value;
}
```

## Possible Solution
```java
package reactor.core.publisher;

@Test
void fixCase() {
     var replayBuffer = new FluxReplay.SizeBoundReplayBuffer<Integer>(0);
     replayBuffer.add(11);
     replayBuffer.add(22);
     assert replayBuffer.head == replayBuffer.tail;
     assert replayBuffer.head.value == replayBuffer.tail.value;
     assert replayBuffer.head.value == 22;
}
```

## Your Environment
doesn't bloody matter

* Reactor version(s) used: 3.5.0

MikkelHJuul:
I will just add that the possible solution stated here fixes the head != tail issue but the buffer will not emit its value, it looks like this is a "simple" off-by-one error.

Consider making a singleton implementation of the buffer. And moving to using the Java's AtomicReferenceArray in stead of the linked list (Node extends AtomicReference<Node>). This should improve memory and performance.

MikkelHJuul:
My least effort implementations of a SingletonReplayBuffer and an AtomiReferenceArray-based implementation

There is bound to be some issues with these, but they both passed a naive test:

```
    public static <T> Sinks.Many<T> replayLatestSink() {
        var sink = new SinkManyReplayProcessor<>(new AtomicArraySizeBoundReplayBuffer<T>(1));
        return sink; // and wrapped: new SinkManySerialized<>(sink, sink);
    }
    public static <T> Sinks.Many<T> replayLatestSinkSingleton() {
[...]

PR #3410, opened 2023-03-22
[issues/3340] fix for leak

MikkelHJuul:
This is a simple fix for the leak of memory. It is slightly unsatisfactory as the memory required for the extra node (especially in the `#replay(1)`case) is still very high. With the extra `Node`, taking up 24 unnecessary bytes (27% of the `#replay(1)`-buffer's memory) 

I have also added some extra tests, that I initially added for my other attempts at fixing this issue.

I have also fixed the same issue for the `SizeAndTimeBoundReplayBuffer`.

### DIFF
--- a/reactor-core/src/main/java/reactor/core/publisher/FluxReplay.java
+++ b/reactor-core/src/main/java/reactor/core/publisher/FluxReplay.java
@@ -155,7 +155,7 @@ final class FluxReplay<T> extends ConnectableFlux<T>
 		final Scheduler scheduler;
 		int size;
 
-		volatile TimedNode<T> head;
+		final TimedNode<T> head;
 
 		TimedNode<T> tail;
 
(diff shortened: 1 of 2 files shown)


MikkelHJuul:
this is obviously related to https://github.com/reactor/reactor-core/issues/3340, and fixes the leak, for all cases, including TimeAndSizeBoundReplayBuffer

MikkelHJuul:
On a side-note; I have run the stress test that I also added in relation to the PR #3364, and this doesn't fix the concurrency bug, and will still throw null pointers during `#add`

MikkelHJuul:
@OlegDokuka or others is there anything more I can add here to get this PR through. I haven't had much time for a long time now but it would be nice to push this and possibly the other related one (for the single case)

It seems there needs be a license header in the files added, is this simply copy paste work? What about the other failing check I cannot see logs anymore

PR #3418, opened 2023-03-28
Propagate ThreadLocals for non-Reactor upstream sources

### LINKED ISSUE #3366: ThreadLocal values lost after non-Reactor upstream sources continue execution
In cases like:

- `Flux.from(Publisher<? extends T> source)`
- `Mono.fromFuture(CompletableFuture<? extends T> future)`

where the source can make `Thread` context switches, we'd still like to provide users the means to restore `ThreadLocal` values using the automatic context-propagation mechanism. Currently, the user would need to manually create a boundary, like this:

    Mono.fromFuture(future).contextWrite(Function.identity())

to be able to restore `ThreadLocal`s from current `Context`.

These can be implemented as something like "onFirstOperator" hook, or "onExternalSource".

---

Making the external source aware of and able to use ThreadLocals registered with `ContextRegistry` from context-propagation library is outside of the scope of this issue. For that, we could prepare a documentation section with best practices and guidance as to how to deal with these:
- for simple cases that only use an `Executor` to run tasks -> explain how to wrap the `Executor` with the context-propagation library
- for simple cases where user code is run with `Runnable` or `Callable` primitives, explain how to wrap these with the context-propagation library
- for advanced cases, like using a different reactive-streams implementation, or event-loops, explain how context-propagation can be used in these cases and what are the risks of non-linear async execution

chemicL:
Factory methods for creating `Flux` and `Mono` from non-Reactor sources now restore `ThreadLocal` values when `Hooks.enableAutomaticContextPropagation()` was called in the following cases:

* `Flux.from(Publisher)`
* `Mono.from(Publisher)`
* `Mono.fromDirect(Publisher)`
* `Mono.fromFuture(CompletableFuture)`
* `Mono.fromCompletionStage(CompletionStage)`

and relevant overloads.

Fixes #3366.

### DIFF
--- a/reactor-core/src/main/java/reactor/core/publisher/MonoCompletionStage.java
+++ b/reactor-core/src/main/java/reactor/core/publisher/MonoCompletionStage.java
@@ -24,9 +24,9 @@ import java.util.concurrent.Future;
 import java.util.concurrent.atomic.AtomicIntegerFieldUpdater;
 import java.util.function.BiFunction;
 
+import io.micrometer.context.ContextSnapshot;
 import reactor.core.CoreSubscriber;
 import reactor.core.Exceptions;
-import reactor.core.Fuseable;
 import reactor.core.Scannable;
 import reactor.util.annotation.Nullable;
 import reactor.util.context.Context;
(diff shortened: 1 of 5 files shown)


reactor-core/src/main/java/reactor/core/publisher/MonoCompletionStage.java
```diff
@@ -62,8 +68,6 @@ public Object scanUnsafe(Attr key) {
     }
 
     static class MonoCompletionStageSubscription<T> implements InnerProducer<T>,
-                                                               Fuseable,
-                                                               QueueSubscription<T>,
```
chemicL, line 66: Note, `MonoCompletionStage` also has `Fuseable` removed. They are not needed, as the fusion support is `NONE`. And it makes the subscriber wrapping to the restoring subscriber possible (it is not `Fuseable`).

chemicL:
Verified against the reproducer provided in https://github.com/spring-projects/spring-boot/issues/34374 - trace info is present with this branch.

OlegDokuka (approved):

PR #3420, opened 2023-03-30
Transparent contextCapture in block operators

### LINKED ISSUE #3406: Apply contextCapture automatically in block operators
## Motivation

When `WebClient` is used in Spring MVC with `block` operator, users always need to call `contextCapture` to grab the context on the thread to properly populate the context in addition to `Hooks.enableAutomaticContextPropagation()`.

## Desired solution

The expectation from many developers in imperative programming world would be that context are automatically propagated such `WebClient` operation chain, and wherever writing logs, MDC is properly populated.

The `block` operators could be a trigger to capture context. So that, `WebClient` users with blocking code can automatically get context propagation without always specifying `contextCapture`.

## Additional context

It is a headache for developers to always ask to call `contextCapture` when they use `WebClient` in imperative code with `block` operators.
Many cases, they have switched `RestTemplate` to `WebClient` and realized MDC logging is not working.
Their expectation is simply `WebClient` call with `block` to work as same with `RestTemplate`...

chemicL:
Operators that block to retrieve the result from the reactive chain are usually used in imperative contexts. If `ThreadLocal` values are present at call site, they can be captured automatically. This change frees the user from calling `contextCapture` everywhere that they block.

The following operators were altered to capture automatically if the context-propagation library is on the classpath:

* `Mono#block`,
* `Mono#blockOptional`,
* `Flux#blockFirst`,
* `Flux#blockLast`,
* `Flux#toIterable`.

Fixes #3406.


navaneeth-spotnana:
@chemicL thanks a lot for this PR. 

It will be great if this can be merged soon, as I am migration our codebase to springboot 3 and without this I will have to add `.contextCapture()` in a few hundred places.

OlegDokuka (approved):
Nice!

rstoyanchev (commented):
I like that this behavior is connected with the `enableAutomaticContextPropagation` hook. I'm wondering whether a separate `autoCaptureInBlock` flag is even needed to start?

Arguably making `ThreadLocal` values available automatically goes along capturing automatically at the start of the chain. So unless we have something specific in mind, I can suggest dropping the extra flag for now. We can always add it later if the need arises. At that point it may be more clear whether it should be a boolean, or a key predicate, some sort of function providing more control.

PR #3429, opened 2023-04-10
remove jcstress tests from publish workflow

OlegDokuka:
this pr removes JCstress tests from publish workflow


pderop (approved):
LGTM, thanks.

PR #3445, opened 2023-04-25
[#3446] Optimize ContextN putAllInto

koo-taejin:
Optmize ContextN putAllInto .

Hello.
I often use Coroutines and Reactors together.

To use these two together
I often use the `Await.kt` from `https://github.com/Kotlin/kotlinx.coroutines`
Unfortunately, there are performance issues.

Extract the ReactorContext from `Await.kt`.
Publisher has the extracted ReactorContext, and Subscriber has an empty ReactorContext because it creates a new one.
And finally combine these two ReactorContext in `reactor.util.context.ContextN#putAllInto`.

In my case, I always have 20 to 30 ReactorContexts.

In this case, base is empty, so the code below runs.
```
Context[] holder = new Context[]{base};
forEach((k, v) -> holder[0] = holder[0].put(k, v));
return holder[0];
```

If this code assumes that there are 20 contexts like me, many Contexts will be created.
In addition, because the previous value is copied again

image

```
1(Context1) (copy previous value 0)
2(Context2) (copy previous value  1)
3(Context3) (copy previous value  2)
4(Context4) (copy previous value  3)
5(Context5) (copy previous value  4)
6(ContextN) (copy previous value  5)
...
20 
```
To create ContextN, my application has to copy a lot.

If you change this part like the changed code, resource usage can be reduced.
Anyway, since `this` is already `ContextN`, the result will be made into ContextN.
Only the number of Contexts will be copied, and Context objects are created only once, so it is judged that it will be possible to save resources much more than before.

Thanks :)


OlegDokuka:
@koo-taejin can you please rebase this on 3.4.x branch and change the base as well, this has to be fixed in both, 3.4.x and 3.5.x

OlegDokuka:
Also, dont forget to apply `./gradlew applySpotless' as suggested there https://github.com/reactor/reactor-core/actions/runs/4795306427/jobs/8530817921?pr=3445#step:6:21

OlegDokuka:
@koo-taejin it has to be rebased still

OlegDokuka:
Ohh, I see. Closing in favor of #3445

Issue #3446, opened 2023-04-25
Optmize ContextN putAllInto


koo-taejin:
Hello.

## Motivation
I often use Coroutines and Reactors together.

To use these two together
I often use the `Await.kt` from `https://github.com/Kotlin/kotlinx.coroutines`
Unfortunately, there are performance issues.

Extract the ReactorContext from `Await.kt`.
Publisher has the extracted ReactorContext, and Subscriber has an empty ReactorContext because it creates a new one.
And finally combine these two ReactorContext in `reactor.util.context.ContextN#putAllInto`.

In my case, I always have 20 to 30 ReactorContexts.

In this case, base is empty, so the code below runs.
```
Context[] holder = new Context[]{base};
forEach((k, v) -> holder[0] = holder[0].put(k, v));
return holder[0];
```

If this code assumes that there are 20 contexts like me, many Contexts will be created.
In addition, because the previous value is copied again

![image](https://user-images.githubusercontent.com/10057874/234218576-40562ae6-f23f-4cc2-be3d-77fe37ae8f57.png)

```
1(Context1) (copy previous value 0)
2(Context2) (copy previous value  1)
3(Context3) (copy previous value  2)
4(Context4) (copy previous value  3)
5(Context5) (copy previous value  4)
6(ContextN) (copy previous value  5)
...
20 
```
To create ContextN, my application has to copy a lot.

## Desired solution
- as-is 
```
    Context[] holder = new Context[]{base};
    forEach((k, v) -> holder[0] = holder[0].put(k, v)); // Context is created as many times and copy previous values.
    return holder[0];
```	

- desired solution 
```
    final ContextN newContext = new ContextN(base.size() + this.size()); // ContextN object is created only once.
    newContext.putAll((Map<Object, Object>) this);
    base.forEach((k, v) -> newContext.accept(k, v));
    return newContext;
```

## Additional context
PR : https://github.com/reactor/reactor-core/pull/3445

If you change this part like the changed code, resource usage can be reduced.
Anyway, since `this` is already `ContextN`, the result will be made into ContextN.
Only the number of Contexts will be copied, and Context objects are created only once, so it is judged that it will be possible to save resources much more than before.

Below is the code comparing when base is ContextN or not.
When running with AsyncProfiler individual tests on my MACBook

CPU
- as-is: 1,483,000
- to-be: 186,000
Memory Allocation
- as-is: 28g
- to-be : 13g

```
class AwaitTest {

    val count = 10000000;

[...]

PR #3448, opened 2023-04-25
Backport: [#3446] Optimize ContextN putAllInto

koo-taejin:


OlegDokuka:
@koo-taejin can you please rework [`defaultPutAllOtherIsAbstractContext`](https://github.com/reactor/reactor-core/blob/96cde5be59fbd64b867a1006cc29d2104c7e48bf/reactor-core/src/test/java/reactor/util/context/ContextTest.java#L468) so it does not fail

koo-taejin:
@OlegDokuka Ok. I will check it.

koo-taejin:
I have updated. :)

OlegDokuka:
@koo-taejin have noticed a few more things. let's work through it to ensure we are not breaking the original method behavior documented in the `CoreContext` class. Also, please rollback the test, since it was checking putAllInto literally puts context of Context A into ContextB

reactor-core/src/test/java/reactor/util/context/ContextTest.java
```diff
@@ -482,10 +538,10 @@ void defaultPutAllOtherIsAbstractContext() {
 		right.accept(6, "F");
 
 		Context combined = left.putAll(right.readOnly());
-		assertThat(combined).isInstanceOf(ForeignContext.class);
-		ForeignContext combinedN = (ForeignContext) combined;
+		assertThat(combined).isInstanceOf(ContextN.class);
+		ContextN combinedN = (ContextN) combined;
```
OlegDokuka, line 542: Lets roll that back

koo-taejin:
@OlegDokuka 

It failed.

The reason is that have already made new ConextN object.

Documentation doesn't seem to be a problem(documentation doesn't say that the same class comes out. And CoreContext are already being retyped.) , but it seems best to maintain compatibility. 😭 

In order to keep it as similar as possible.
Changing ContextN's PutAllInto to the following looks good:

```
@Override
public Context putAllInto(Context base) {
    if (base instanceof CoreContext) { // ContextN -> CoreContext 
        ContextN newContext = new ContextN(base.size() + this.size());
        newContext.putAll(base. readOnly()); // newContext.putAll((Map<Object, Object>) base); ->  newContext.putAll(base. readOnly());
        newContext.putAll((Map<Object, Object>) this);
        return newContext;
    }

    Context[] holder = new Context[]{base};
    forEach((k, v) -> holder[0] = holder[0].put(k, v));
    return holder[0];
}

```

I would wait your advice.

Thanks :)

koo-taejin:
@OlegDokuka 
I've rolled that back to what you said. 
As mentioned above, it will work as before for custom Context to maintain compatibility. 
However, for CoreContext, it will work more efficiently than before.

reactor-core/src/test/java/reactor/util/context/ContextTest.java
```diff
@@ -464,6 +465,61 @@ void defaultPutAllMap() {
...
+		ContextN left = new ContextN(leftMap);
+		ContextN right1 = new ContextN(rightMap);
+		Context1 right2 = new Context1(key, newValue);
+
+		final Context combined1 = left.putAllInto(right2);
+		final Context combined2 = left.putAllInto(right1);
+
+		// The result is to have the value of the key included as an argument, if the same key exists.
+		Assertions.assertEquals((String) combined1.get(key), (String) combined2.get(key));
+	}
```
OlegDokuka, line 521: Can we also use Mockito constructor instrumentation to see that we allocate an object exactly once?

OlegDokuka (approved):
Look good to me! Good job!

One minor comment to resolve and we can merge that

reactor-core/src/test/java/reactor/util/context/ContextTest.java
koo-taejin: I have checked that the test fails, so I am going to fix the code to make it work. 😭 

I don't know if it can be verified with Mockito's mock or spy, because creating a new ContextN object internally. 
I'll check the requirements more.

koo-taejin: Mockito's spy() and mock() can't transform the final class, so I used a Proxy. 

And I changed the production code like below.

```
newContext.putAll(base.readOnly()); -> ((CoreContext) base).unsafePutAllInto(newContext);
```

Thanks :)

OlegDokuka:
Good work, @koo-taejin! Look forward to seeing more contributions from you ;)

PR #3459, opened 2023-05-10
Avoid IllegalStateException: Cannot resolve type description for io.micrometer.context.ContextRegistry

pderop:
@chemicL, as we have discussed, this is the PR from my old patch, which seems to avoid the `IllegalStateException: Cannot resolve type description for io.micrometer.context.ContextRegistry` when context-propagation jar is not in the classpath and when Blockhound is used.
Can you take a look ? 

This patch avoids direct access to the `ContextPropagation` class if we know that the context-propagation jar is not in the classpath; so in this case, it will avoid Blockhound instrumenting the class; hence it will avoid any IllegalStateException.

tested with reactor-netty, as well as with https://github.com/violetagg/GH-3334

The PR should also fix https://github.com/reactor/BlockHound/issues/356

side notes: 

- as we have discussed, I have left unchanged the `ContextPropagationSupport.dontRestoreContextForHandle()` method; maybe you will want to refactor it. 
- I have left the extra checks in the ContextPropagation various methods, maybe you will want to remove them.
-  some static final inner classes from the reactor.core.publisher package are referring to some context propagation types, like the `FluxSource.FluxSourceRestoringThreadLocalsSubscriber` which is using `ContextSnapshot.Scope`. So, when context-propagation is unavailable, these inner classes are not used, but are part of classes which are used. So for the moment, we don't seem to have any problems, but who knows ... so, if we still have any IllegalStateExceptions even with this PR, then most likely that the static inner classes that are using the context-propagation types will have to be moved to the ContextPropagation class. Let's do not rush to do this for the moment, but we may consider to do this at some point in the future if the problem remains.

thanks.

### DIFF
--- a/reactor-core/src/main/java/reactor/core/publisher/ContextPropagation.java
+++ b/reactor-core/src/main/java/reactor/core/publisher/ContextPropagation.java
@@ -17,7 +17,6 @@
 package reactor.core.publisher;
 
 import java.util.AbstractQueue;
-import java.util.Collection;
 import java.util.HashMap;
 import java.util.Iterator;
 import java.util.Map;
@@ -33,8 +32,6 @@ import io.micrometer.context.ContextSnapshot;
 
 import io.micrometer.context.ThreadLocalAccessor;
 import reactor.core.observability.SignalListener;
-import reactor.util.Logger;
-import reactor.util.Loggers;
 import reactor.util.annotation.Nullable;
 import reactor.util.context.Context;
 import reactor.util.context.ContextView;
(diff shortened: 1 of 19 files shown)


reactor-core/src/main/java/reactor/core/publisher/ContextPropagationSupport.java
```diff
@@ -0,0 +1,62 @@
...
+        return isContextPropagationOnClasspath;
+    }
+
+    static boolean shouldPropagateContextToThreadLocals() {
+        return isContextPropagationOnClasspath && propagateContextToThreadLocals;
+    }
+
+    static boolean dontRestoreContextForHandle() {
+        return propagateContextToThreadLocals || !isContextPropagationOnClasspath;
+    }
```
chemicL, line 61: I'd invert the meaning and rename:

```suggestion
    static boolean shouldRestoreThreadLocalsInSomeOperators() {
        return isContextPropagationOnClasspath && !propagateContextToThreadLocals;
    }
```

And invert all the usages of course.

chemicL:
Thank you for the PR @pderop! Just a few minor suggestions and we can 🚢

reactor-core/src/main/java/reactor/core/publisher/ContextPropagation.java
```diff
@@ -120,7 +72,7 @@ public static Function<Runnable, Runnable> scopePassingOnScheduleHook() {
 	 * @return the {@link Context} augmented with captured entries
 	 */
 	static Function<Context, Context> contextCapture() {
-		if (!isContextPropagationOnClasspath) {
+		if (!ContextPropagationSupport.isContextPropagationOnClasspath) {
```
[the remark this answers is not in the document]
pderop: I have applied your feedback in commit https://github.com/reactor/reactor-core/pull/3459/commits/ed0146b9fe8927a4d2ee194817f649d36e24401f

reactor-core/src/main/java/reactor/core/publisher/ContextPropagationSupport.java
pderop: Please carefully check https://github.com/reactor/reactor-core/pull/3459/commits/a5d20c7275f56421339741402190abdb43e0cd51

reactor-core/src/main/java/reactor/core/publisher/ContextPropagation.java
```diff
@@ -162,7 +114,7 @@ static Function<Context, Context> contextCapture(Predicate<Object> captureKeyPre
 	 * @param <R> the transformed type
 	 */
 	static <T, R> BiConsumer<T, SynchronousSink<R>> contextRestoreForHandle(BiConsumer<T, SynchronousSink<R>> handler, Supplier<Context> contextSupplier) {
-		if (propagateContextToThreadLocals || !ContextPropagation.isContextPropagationAvailable()) {
+		if (ContextPropagationSupport.propagateContextToThreadLocals || !ContextPropagationSupport.isContextPropagationAvailable()) {
```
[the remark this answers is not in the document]
pderop: @chemicL , let me know if you are fine with the new commits. thanks !

chemicL (approved):

reactor-core/src/main/java/reactor/core/publisher/ContextPropagationSupport.java
```diff
@@ -0,0 +1,62 @@
...
+import reactor.util.Loggers;
+
+final class ContextPropagationSupport {
+    static final Logger LOGGER = Loggers.getLogger(ContextPropagationSupport.class);
+
+    // Note: If reflection is used for this field, then the name of the field should end with 'Available'.
+    // The preprocessing for native-image support is in Spring Framework, and is a short term solution.
+    // The field should end with 'Available'. See org.springframework.aot.nativex.feature.PreComputeFieldFeature.
+    // Ultimately the long term solution should be provided by Reactor Core.
+    static final boolean isContextPropagationOnClasspath;
```
OlegDokuka, line 29: is it intentionally that comment says field should end on `Available` and this one does not?

OlegDokuka (approved):

OlegDokuka:
@pderop good overall with one small comment/question

reactor-core/src/main/java/reactor/core/publisher/ContextPropagationSupport.java
pderop: no, it is not intentional. I actually tried to reduce the modifications as much as possible in order to avoid regressions; so I kept this comment that was already present in the current version [here](https://github.com/reactor/reactor-core/blob/main/reactor-core/src/main/java/reactor/core/publisher/ContextPropagation.java#L52-L56).

honestly, I don't know if this comment should be changed , or even removed, or we we need to keep it ?

OlegDokuka: I see

pderop:
@OlegDokuka , @chemicL , thank you for the reviews !

PR #3460, opened 2023-05-11
Allow null previous value

marcingrzejszczak:
without this change we can't distinguish in Micrometer (e.g. via `ObservationThreadLocalAccessor`) whether we're introducing an empty value (`setValue(null)`) or resetting `reset()`. 

with this change we always set a value regardless whether it's `null` or not

This change is a prerequisite to fix current issues with OTLA and Reactor and scope leaks and memory leaks etc.


chemicL:
@OlegDokuka this follows from https://github.com/micrometer-metrics/context-propagation/pull/103 and once context-propagation is released (the week of June 5-9), we can include this in the next release (June 13).

Once we merge https://github.com/micrometer-metrics/context-propagation/pull/105 and release (I don't expect this to be a part of the nearest release, but rather a month later) we can further improve reactor's codebase to not include a custom implementation, but use the context-propagation library directly as it should then allow clearing the missing ThreadLocals.

OlegDokuka (approved):

OlegDokuka:
@chemicL sounds good to me

PR #3466, opened 2023-05-22
Improve documentation about subscribeOn

### LINKED ISSUE #3452: Documentation: emphasize that the order of subscribeOn matters
Following from #3450, it should be clarified in the documentation that the order of `subscribeOn` in the chain is significant, as opposed to what the documentation states in https://projectreactor.io/docs/core/release/reference/#schedulers.

chemicL:
The documentation can be misleading with regards to ordering guarantees of subscibeOn with regards to other operators. This change improves the wording.

Resolves #3452.

### DIFF
--- a/docs/asciidoc/coreFeatures.adoc
+++ b/docs/asciidoc/coreFeatures.adoc
@@ -318,9 +318,9 @@ the part of the chain after them.
 * Picks one thread from the `Scheduler`
 
 - NOTE: Only the closest `subscribeOn` call in the downstream chain effectively
- schedules subscription and request signals to upstream operators that can intercept them
- (`doFirst`, `doOnRequest`). Using multiple `subscribeOn` calls will introduce unnecessary
- Thread switches that have no value.
+ schedules subscription and request signals to the source or operators that can
+ intercept them (`doFirst`, `doOnRequest`). Using multiple `subscribeOn` calls will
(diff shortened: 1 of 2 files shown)


docs/asciidoc/coreFeatures.adoc
```diff
@@ -306,17 +306,17 @@ The print happens on the latest execution context, which is the one from `publis
...
+
+However, this does not affect the
 behavior of subsequent calls to `publishOn` -- they still switch the execution context for
 the part of the chain after them.
 
 * Changes the `Thread` from which the *whole chain* of operators subscribes
 * Picks one thread from the `Scheduler`
 
-NOTE: Only the earliest `subscribeOn` call in the chain is actually taken into account.
-
```
OlegDokuka, line 319: How about saying "if multiple subscribeOn are in the chain, only the earliest is applied"

OlegDokuka: I guess in that case, the earliest for 

```
.doFirst(() -> System.out.println("Downstream: " + Thread.currentThread() .getName()))
```

is the one with bounded elastic. Got your point but removing can cause more questions. Maybe 

"Only the closest `subscribeOn` call in the downstream chain is actually taken into account for the operators that can intercept subscription or request signals" something in that spirit

OlegDokuka (approved):

PR #3469, opened 2023-05-23
Removes the Operators.serialize in MonoTimeout#subscribeOrReturn

manzhizhen:
When I debug the code, I find a small problem, the operation of the MonoTimeout#subscribeOrReturn class causes the CoreSubscriber to be serialized twice, which I think can be optimized has.

So I PR the removes the Operators.serialize operation in MonoTimeout#subscribeOrReturn that is unnecessary for CoreSubscriber, because FluxTimeout. TimeoutMainSubscriber has carries on the Operators. Serialize operation


manzhizhen:
> @manzhizhen thank you for the proposal. It looks correct from my perspective. Can you please target 3.4.x branch instead of main?

ok，I will do it

chemicL:
Superseded by #3470

PR #3470, opened 2023-05-23
Remove redundant Operators.serialize in MonoTimeout#subscribeOrReturn

manzhizhen:
When I debug the code, I find a small problem, the operation of the MonoTimeout#subscribeOrReturn class causes the CoreSubscriber to be serialized twice, which I think can be optimized has.

So I PR the removes the Operators.serialize operation in MonoTimeout#subscribeOrReturn that is unnecessary for CoreSubscriber, because FluxTimeout. TimeoutMainSubscriber has carries on the Operators. Serialize operation


chemicL (approved):

OlegDokuka (approved):

chemicL:
@manzhizhen thank you for the contribution!

PR #3482, opened 2023-05-31
Add thread interrupt state propagate (#3451)

### LINKED ISSUE #3451: Interrupt state not propagated when Thread interrupted during Blocking method call
If an InterruptedException is thrown during the operation, the Interrupt state is hidden by the BlockingOptionalMonoSubscriber classes. (Other implementations may have similar issues)

This implementation prevents the Thread.interrupted() method from determining if a thread is in the Interrupt state when an exception is thrown on the thread that called the block method.

## Expected Behavior
If the thread is interrupted while the block method is in the wait state, the Thread.interrupted() method should return true in the catch clause.

## Actual Behavior
The Thread.interrupted() method returns false, even though the thread was interrupted and an exception was thrown in the block method. 

## Steps to Reproduce
In the test code, the Thread.interrupted() assertion in the catch clause fails, even though we interrupted the thread that called the blockOptional method.

```java
@Test
public void interruptStateTest() throws InterruptedException {
	final Object lock = new Object();
	final Thread thread = new Thread(() -> {
		try {
			Mono.create(sink -> {
				try {
					Thread.sleep(1000);
					sink.success();
				}
					Thread.currentThread().interrupt();
				}
			}).blockOptional();
		}
		catch (Exception e) {
			e.printStackTrace();
			assertTrue(Thread.interrupted());
		}
		finally {
			synchronized (lock) {
				lock.notify();
			}
		}
	});
	
	thread.start();
	thread.interrupt();
	
	synchronized (lock) {
		lock.wait();
	}
}
```

## Possible Solution
When an InterruptedException exception is thrown, add `Thread.currentThread().interrupt();` to restore interrupt state the current thread and throw a custom exception.

### Example code
https://github.com/reactor/reactor-core/blob/324a64c3d7e21b9ec51c4e421a88c8a8689263b5/reactor-core/src/main/java/reactor/core/publisher/BlockingOptionalMonoSubscriber.java#L107-L116

Modify the above code as follows
```java
try {
	await();
}
catch (InterruptedException ex) {
	dispose();
	RuntimeException re = Exceptions.propagate(ex);
	//this is ok, as re is always a new non-singleton instance
	re.addSuppressed(new Exception("#blockOptional() has been interrupted"));

	// Restore interrupt state
	Thread.currentThread().interrupt();
	throw re;
}
```

## Your Environment
* Reactor version(s) used: 3.5.5
* Other relevant libraries versions (eg. `netty`, ...): None
* JVM version (`java -version`)
```
openjdk version "1.8.0_322"
OpenJDK Runtime Environment (Temurin)(build 1.8.0_322-b06)
OpenJDK 64-Bit Server VM (Temurin)(build 25.322-b06, mixed mode)
```
* OS and version (eg `uname -a`): Windows 10, 22H2 19045.2913

10sa:
Fix interrupt state not being propagated when interrupted while thread is in wait state by calling blocking method

Fixes #3451

### DIFF
--- a/reactor-core/src/main/java/reactor/core/publisher/BlockingIterable.java
+++ b/reactor-core/src/main/java/reactor/core/publisher/BlockingIterable.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2016-2021 VMware Inc. or its affiliates, All Rights Reserved.
+ * Copyright (c) 2016-2023 VMware Inc. or its affiliates, All Rights Reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -181,6 +181,7 @@ final class BlockingIterable<T> implements Iterable<T>, Scannable {
 					}
 					catch (InterruptedException ex) {
 						run();
+						Thread.currentThread().interrupt();
 						throw Exceptions.propagate(ex);
 					}
 					finally {
--- a/reactor-core/src/main/java/reactor/core/publisher/BlockingOptionalMonoSubscriber.java
+++ b/reactor-core/src/main/java/reactor/core/publisher/BlockingOptionalMonoSubscriber.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2016-2021 VMware Inc. or its affiliates, All Rights Reserved.
+ * Copyright (c) 2016-2023 VMware Inc. or its affiliates, All Rights Reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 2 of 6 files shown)


OlegDokuka:
@10sa welcome on board & great work! Can you please rebase this PR on `3.4.x` branch and change the target from `main` to `3.4.x`

10sa:
Yup, I'll change the target branch first, and I'll get a new branch with the 3.4.x version soon, and then force push with my commits.

10sa:
On my PC, some unit tests report failure when running `gradle check`. However, the 3.4.x branch also fails the unit test.

I don't know the cause, I've completed the rebase work for now.

OlegDokuka (approved):

OlegDokuka:
@10sa good work! Thanks for your contribution!

PR #3484, opened 2023-06-01
ensures cancel handler is wrapped in try catch and drops potential cancellation failure

### LINKED ISSUE #3417: Mono.fromCompletionStage fails to propagate error signal when exception is thrown in `map`, `flatMap` and `handle`
A Mono created with `Mono.fromCompletionStage` fails to propagate an error signal downstream in the following cases:
- `mono.map` throws an unchecked exception
- `mono.flatMap` returns a mono containing an error signal
- `mono.handle` signals an error on the provided sink

Others like `Mono.fromFuture` work fine, please see the following repository containing simple test cases on how to reproduce: https://github.com/mcanalesmayo/reactor-core-cs-bug

## Expected Behavior
Error signal to be propagated downstream

## Actual Behavior
Error signal is lost

## Steps to Reproduce
https://github.com/mcanalesmayo/reactor-core-cs-bug

## Possible Solution
Did not look into details on how to fix, but a simple workaround is just using `Mono.fromFuture(cs.toCompletableFuture())` instead

## Your Environment

* Reactor version(s) used: 3.5.4
* JVM version (`java -version`): 11
* OS and version (eg `uname -a`): macOS Ventura 13.2.1

OlegDokuka:
closes #3417.

Also, this PR prepares the ground for potential improvement over `fromCompletionStage` method. In the following PR to that I plan to propose the following overloads: 

`Mono.fromCompletionStage(stage, () -> stage.cancel())` 
`Mono.fromCompletionStage(() -> CompletableFuture.runAsync(...), future -> future.cancel())`

`Mono.fromCompletionStage(stage, null) ` where null disables any cancellation propagation
`Mono.fromCompletionStage(() -> CompletableFuture.runAsync(...), null)` where null disables any cancellation propagation

### DIFF
--- a/reactor-core/src/main/java/reactor/core/publisher/Mono.java
+++ b/reactor-core/src/main/java/reactor/core/publisher/Mono.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2016-2022 VMware Inc. or its affiliates, All Rights Reserved.
+ * Copyright (c) 2016-2023 VMware Inc. or its affiliates, All Rights Reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


OlegDokuka:
Closing since rework adds news stuff which may cause new potential bugs, will mitigate problem simpler

Issue #3487, opened 2023-06-02
refCount gets hung : race conditions in FluxRefCount#subscribe


gowa:
## Expected Behavior
Use case 1 (real world scenario): I need to split the given INPUT into several 'paths' using different filters, process each path, and then merge paths back again. With this pattern, the INPUT flux is 'processed' just once.

The following chain should not block:
REF_COUNTED = some INPUT Flux -> publish -> refCount(2)
A = apply transformation to REF_COUNTED
B = apply transformation to REF_COUNTED
Flux.merge(A,B) -> blockLast

Use case 2 (just another way to reproduce the problem): split the input into several paths, read results for each path in another thread.
Main Thread:
REF_COUNTED = some INPUT Flux -> publish -> refCount(2)
A = apply transformation to REF_COUNTED
B = apply transformation to REF_COUNTED
Thread1: A -> blockLast
Thread2: B -> blockLast

## Actual Behavior
While the above example works fine for simple 'transformations', it blocks forever if some 'transformation' involves another thread.
E.g. A = "some Mono, finishing in another thread" -> flatMapMany(REF_COUNTED)

## Steps to Reproduce

```java
import org.junit.Assert;
import org.junit.Test;
import reactor.core.CoreSubscriber;
import reactor.core.publisher.Flux;
import reactor.core.publisher.Mono;
import reactor.test.StepVerifier;

import java.time.Duration;
import java.util.Collections;
import java.util.HashMap;
import java.util.Map;
import java.util.function.Function;

public class RefCountFluxTest {
    private static final int TEST_ATTEMPTS = 100_000;

    @Test
    public void willFail() {
        performTest(Function.identity());
        //java.lang.AssertionError: expectation "expectNext(2)" failed (expected: onNext(2); actual: onComplete())
        //java.lang.AssertionError: expectation "expectNext(1)" failed (expected value: 1; actual value: 2)
        //java.lang.AssertionError: VerifySubscriber timed out on ...
    }

    @Test
    public void willNotFail() {
        performTest(SynchronizedSubscribeFlux::new);
    }

    /**
     * Flux.just(1,2) -> refCount(2) -> Flux.merge(odd,even)
     */
    private void tryReproduce(Function<Flux<Integer>, Flux<Integer>> modifyInput) {
        Flux<Integer> input = Flux.just(1, 2).publish().refCount(2);

        Flux<Integer> modifiedInput = modifyInput.apply(input); //either original input or with fix applied

[...]

PR #3488, opened 2023-06-03
#3487 fixes FluxRefCount onSubscribe race conditions

gowa:
When many subscribers subscribe from different threads, some of them may get stuck not being able to consume the source Flux, while this does not happen if they subscribe from the single thread.


OlegDokuka:
superseded by #3517

PR #3489, opened 2023-06-05
context-propagation: use ContextSnapshotFactory

chemicL:
This change removes the custom logic for setting ThreadLocal values. context-propagation library did not allow removing values missing in the Reactor Context - a feature necessary in order to avoid leaking ThreadLocals.

https://github.com/micrometer-metrics/context-propagation/pull/105 introduced this feature into the public API and now reactor-core is able to use it.


reactor-core/src/main/java/reactor/core/publisher/ContextPropagation.java
```diff
@@ -380,90 +414,5 @@ public Iterator<T> iterator() {
 			this.body = body;
 			this.contextSnapshot = contextSnapshot;
 		}
-
-	}
-
-	static final class ContextCaptureNoPredicate implements Function<Context, Context> {
```
chemicL, line 386: As part of this change I moved this class above the Queue wrappers and closer to the context capture operator logic.

OlegDokuka (approved):

Issue #3084, opened 2022-06-17
Project Loom and Reactor


lynch19:
Now that virtual threads and structured concurrency are about to get previewed for the first time on JDK19, I'd like to raise the discussion of Loom & Reactor [once more](https://github.com/reactor/reactor-core/issues/2548).
On the referenced issue, @OlegDokuka answered well that Reactor isn't just about easily writing asynchronous code, but also giving the user a bunch of great streaming operators, much better than any stream API currently offers.
But still, Reactor has lots to do with making the software non-blocking. @simonbasle @OlegDokuka I'd like to continue the discussion of the future of Reactor with virtual threads and structured concurrency. In my opinion, early support for Loom features can do lots of good for heavy Reactor users. Some heavy questions that come to my mind are:

* Will a new Loom-based Scheduler get released? And when?
* Will Reactor release JDK20+ releases include support for using the Reactor operators without `Mono` and `Flux`? Or will users need to continue using the `.block()` operation even if they won't really block?
* Will Reactor turns its focus to operator-based features & infrastructure-based features (like better observability, and better integration with other technologies like databases, debugging, etc)?
  * There are currently projects like Spring Cloud Function and Spring Cloud Stream that uses Reactor even though they block (Spring Cloud Stream using Reactor without having a reactive binder at all). Will such use cases of Reactor as a powerful processing operators framework become the main intention of the project?
  * Maybe a better integration with the Java stream API should be considered?
* Will there be a breaking changes release once Reactor support Loom? And what else do you think needs to get changed?
  * A release that will include changing lots of Reactor's internals to be based on virtual threads?
  * How will the support of Reactor for JDK18- look like, after Loom-based JDKs will get released?

I think that Reactor should be the first to get prepared for this change in Java since it affects the need for some of its killer features. Reactor **should** have answers for these questions, and for more questions similar to this since they're going to come by the time. Webflux & Reactor are pretty popular by now, and there's no user that wants to see the project losing its popularity. Thus Reactor needs to be the first to take this challenge and support Loom, explain the relationship between the projects (and how to seemingly enable them), as well as redesign and rethinking about parts of the Reactor framework that can improve the usage with Loom-included Java.

In my opinion, shifting the Project's purpose a little and making it a functional Java API dedicated to stream processing, data pipelines, APIs and more is the right direction.

aksh1618:
[...]

PR #3491, opened 2023-06-07
enables automatic loom support 

OlegDokuka:
This PR replaces VirtualThread instead of Platform Threads for `Schedulers.boundedElastic` when VTs are available on the class path.

related #3084

Signed-off-by: OlegDokuka <odokuka@vmware.com>


chemicL:
The Virtual Thread integration was released. We can revisit the mapAsync-like operator in the future.

PR #3492, opened 2023-06-07
improves onDiscard support in sinks

### LINKED ISSUE #3359: EmitterProcessor queue not cleared on immediate cancellation
`EmitterProcessor#remove` has support to discard the buffer once the last subscriber is gone. However, it returns early if the subscribers array is already `EMPTY`. For cases where the subscription is immediately cancelled (e.g. `.take(0)`) this may leak the buffered elements. 

Here is a repro case:
```
@Test
void shouldClearBufferOnCancellation() {
  BlockingQueue<Integer> queue = new LinkedBlockingQueue<>();

  Many<Integer> sink = Sinks.many().unicast().onBackpressureBuffer(queue);

  sink.tryEmitNext(1);
  sink.tryEmitNext(2);
  sink.tryEmitNext(3);

  assertThat(queue).hasSize(3);

  sink.asFlux().take(0).blockLast();

  assertThat(queue).isEmpty();
}
```

As a workaround `.takeWhile(x -> false)` does make the test pass for example.

OlegDokuka:
closes #3359 

### DIFF
--- a/reactor-core/src/main/java/reactor/core/publisher/EmitterProcessor.java
+++ b/reactor-core/src/main/java/reactor/core/publisher/EmitterProcessor.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2016-2022 VMware Inc. or its affiliates, All Rights Reserved.
+ * Copyright (c) 2016-2023 VMware Inc. or its affiliates, All Rights Reserved.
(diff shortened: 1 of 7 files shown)


chemicL:
Closing due to lack of activity.

PR #3514, opened 2023-06-22
[doc] Document usage of Mono.zip() along with Mono<Void> (#3306)

### LINKED ISSUE #3306: Document Mono.zip subscription scheme when Mono<Void> is supplied
The test below fails when using Mono<Void>, if you slightly change the code to use Mono<Integer> it passes.
3 subscribes are expected, but only one is issued.

```
    @Test
    public void testZip() {
        AtomicInteger cnt = new AtomicInteger();
        Mono<Void> m1 = Mono.create(sink -> {
            cnt.incrementAndGet();
            sink.success();
        });
        Mono<Void> m2 = Mono.create(sink -> {
            cnt.incrementAndGet();
            sink.success();
        });
        Mono<Void> m3 = Mono.create(sink -> {
            cnt.incrementAndGet();
            sink.success();
        });
        Mono<Void> m4 = m1.zipWith(Mono.zip(List.of(m2,m3), objects -> true), (value1, value2) -> value1);
        m4.block();
        assertEquals(3, cnt.get());
    }
```

RomaKudryavtsev:
Documented that upon supply of Mono<Void> it is not guaranteed that the resulted zipped Mono will not complete immediately upon empty completion of Mono<Void>. Fixes #3306
Author: Roman Zandler <rmnzndlr@gmail.com>

### DIFF
--- a/docs/asciidoc/faq.adoc
+++ b/docs/asciidoc/faq.adoc
@@ -157,57 +157,55 @@ myMethod.emptySequenceForKey("a") // this method returns empty Mono<String>
 ----
 ====
 
-[[faq.monoZipVoid]]
-== Using `zip` along with `Mono<Void>`
+[[faq.monoZipEmptyCompletion]]
+== Using `zip` along with empty-completed publishers
+
+When using the `zip` operator along with empty-completed publishers, it is important to be aware of the following behavior.
 
 Consider the following test case:
 
 ====
 [source,java]
 ----
-Function<Object[], Void> voidCombinator = objects -> Void.TYPE.cast(Void.class);
-
     @Test
-    public void testZipVoidNonEmptyCompletion() {
+    public void testZipEmptyCompletionAllSubscribed() {
         AtomicInteger cnt = new AtomicInteger();
-        Mono<Void> mono1 = Mono.create(sink -> {
+        Mono<Integer> mono1 = Mono.create(sink -> {
             cnt.incrementAndGet();
             sink.success();
         });
-        Mono<Void> mono2 = Mono.create(sink -> {
+        Mono<Integer> mono2 = Mono.create(sink -> {
             cnt.incrementAndGet();
             sink.success();
         });
-        Mono<Void> zippedMono = Mono.zip(List.of(mono1, mono2), voidCombinator);
+        Mono<Integer> zippedMono = Mono.zip(mono1, mono2, (v1, v2) -> v1);
         zippedMono.subscribe();
         assertEquals(2, cnt.get());
     }
 ----
 ====
 
-While in this case the resulting `zippedMono` does not complete immediately upon completion of `mono1`, such behaviour is not guaranteed for all cases. For instance, consider the following test case:
+While in this case the resulting `zippedMono` does not complete immediately upon empty completion of `mono1`, such behaviour is not guaranteed for all cases. For instance, consider the following test case:
 
 ====
 [source,java]
 ----
-Function<Object[], Void> voidCombinator = objects -> Void.TYPE.cast(Void.class);
-
     @Test
-    public void testZipVoidEmptyCompletion() {
+    public void testZipEmptyCompletionOneSubscribed() {
         AtomicInteger cnt = new AtomicInteger();
-        Mono<Void> mono1 = Mono.create(sink -> {
+        Mono<Integer> mono1 = Mono.create(sink -> {
             cnt.incrementAndGet();
             sink.success();
         });
-        Mono<Void> mono2 = Mono.create(sink -> {
+        Mono<Integer> mono2 = Mono.create(sink -> {
             cnt.incrementAndGet();
             sink.success();
         });
-        Mono<Void> mono3 = Mono.create(sink -> {
+        Mono<Integer> mono3 = Mono.create(sink -> {
             cnt.incrementAndGet();
             sink.success();
         });
-        Mono<Void> zippedMono = Mono.zip(List.of(mono1, Mono.zip(List.of(mono2, mono3), voidCombinator)), voidCombinator);
+        Mono<Integer> zippedMono = Mono.zip(mono1, Mono.zip(mono2, mono3, (v1, v2) -> v1), (v1, v2) -> v1);
         zippedMono.subscribe();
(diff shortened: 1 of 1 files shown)


chemicL:
I'm a bit skeptical of the use of the combinator in the examples, and especially the return of `Void` value in the `Function`. I have a feeling the explanation can be simpler. I believe it's also my mistake to assume we're documenting behaviour in case of `Void`, but narrowing it down to this specific case makes the comprehension more difficult. The principle applies to any type.

Consider the following alternative, without using `Void`:

	@Test
	public void testZipEmptyCompletionAllSubscribed() {
		AtomicInteger cnt = new AtomicInteger();
		Mono<Integer> mono1 = Mono.create(sink -> {
			cnt.incrementAndGet();
			sink.success();
		});
		Mono<Integer> mono2 = Mono.create(sink -> {
			cnt.incrementAndGet();
			sink.success();
		});
		Mono<Integer> zippedMono = Mono.zip(mono1, mono2, (v1, v2) -> v1);
		zippedMono.subscribe();
		assertEquals(2, cnt.get());
	}

	@Test
	public void testZipEmptyCompletionOneSubscribed() {
		AtomicInteger cnt = new AtomicInteger();
		Mono<Integer> mono1 = Mono.create(sink -> {
			cnt.incrementAndGet();
			sink.success();
		});
		Mono<Integer> mono2 = Mono.create(sink -> {
			cnt.incrementAndGet();
			sink.success();
		});
		Mono<Integer> mono3 = Mono.create(sink -> {
			cnt.incrementAndGet();
			sink.success();
		});
		Mono<Integer> zippedMono = Mono.zip(mono1, Mono.zip(mono2, mono3, (v1, v2) -> v1), (v1, v2) -> v1);
		zippedMono.subscribe();
		assertEquals(1, cnt.get());
	}

	@Test
	public void testZipOptionalAllSubscribed() {
		AtomicInteger cnt = new AtomicInteger();
		Mono<Integer> mono1 = Mono.create(sink -> {
			cnt.incrementAndGet();
			sink.success();
		});
		Mono<Integer> mono2 = Mono.create(sink -> {
			cnt.incrementAndGet();
			sink.success();
		});
		Mono<Integer> mono3 = Mono.create(sink -> {
			cnt.incrementAndGet();
			sink.success();
		});
		Mono<Optional<Integer>> zippedMono =
				Mono.zip(
						mono1.singleOptional(),
						Mono.zip(mono2.singleOptional(), mono3.singleOptional(), (v1, v2) -> v1),
						(v1, v2) -> v1);
		zippedMono.subscribe();
		assertEquals(3, cnt.get());
	}

RomaKudryavtsev:
@chemicL , many thanks for your feedback! So, in fact, the problem lies not in the use of Void, but in the empty completion itself (and such behavior is pretty same with any type - e.g., Mono<Integer> as well). I will refactor then my input this weekend.

RomaKudryavtsev:
@chemicL , I have committed changes in accordance with your above suggestions. Please let me know if anything else might be improved.

docs/asciidoc/faq.adoc
```diff
@@ -157,6 +157,96 @@ myMethod.emptySequenceForKey("a") // this method returns empty Mono<String>
...
+            sink.success();
+        });
+        Mono<Integer> zippedMono = Mono.zip(mono1, mono2, (v1, v2) -> v1);
+        zippedMono.subscribe();
+        assertEquals(2, cnt.get());
+    }
+----
+====
+
+While in this case the resulting `zippedMono` does not complete immediately upon empty completion of `mono1`, such behaviour is not guaranteed for all cases. For instance, consider the following test case:
```
chemicL, line 188:
```suggestion
While in this case the resulting `zippedMono` subscribes to both `mono1` and `mono2`, such behaviour is not guaranteed for all cases. For instance, consider the following test case:
```

chemicL: I think the language mentioning completing immediately is not the essence here and can be confusing. Whether it completes before or after is an implementation detail, but the user is most probably interested in the act of subscribing and processing the sources.

docs/asciidoc/faq.adoc
```diff
@@ -157,6 +157,96 @@ myMethod.emptySequenceForKey("a") // this method returns empty Mono<String>
...
+            cnt.incrementAndGet();
+            sink.success();
+        });
+        Mono<Integer> zippedMono = Mono.zip(mono1, Mono.zip(mono2, mono3, (v1, v2) -> v1), (v1, v2) -> v1);
+        zippedMono.subscribe();
+        assertEquals(1, cnt.get());
+    }
+----
+====
+In this case upon empty completion of `mono1`, `zippedMono` completes immediately.
```
chemicL, line 214:
```suggestion
In this case upon empty completion of `mono1`, `zippedMono` completes immediately and does not subscribe to `mono2` and `mono3`.
```

docs/asciidoc/faq.adoc
```diff
@@ -157,6 +157,96 @@ myMethod.emptySequenceForKey("a") // this method returns empty Mono<String>
...
+
+In cases where `zip` operator is used to combine empty-completed publishers, it is therefore not guaranteed that the resulting publisher will complete only upon completion of all publishers to be combined. In other words, while in some cases this behaviour might be different, it is not guaranteed that the resulting publisher will not complete immediately upon empty completion of the first publisher to be zipped.
```
chemicL, line 216: Please consider rewriting this according to the above suggestions around subscription and not completion.

chemicL (changes requested):

chemicL:
I made some more comments. Thanks for revisiting once more. I think the essence here is - completion without emitting an item can influence the act of subscribing to provided sources in undefined ways. It can either subscribe and cancel / discard / never subscribe. As far as I understood the risk is not subscribing to most users and the goal here is to show how to ensure subscription to all the sources.

RomaKudryavtsev:
@chemicL , many thanks for your feedback! I will revert with revisions in the following 2 days.

RomaKudryavtsev:
@chemicL , hi! I have made some changes in accordance with your suggestions. Hopefully, now the description is clear enough.

Issue #3512, opened 2023-06-21
Enhance blockhound integration to allow blocking calls in `FutureTask.handlePossibleCancellationInterrupt`


OlegDokuka:
# Problem

The `shouldNotReportWorkerScheduledFutureTask` test is failing randomly with the following cause: 

```
13:02:12.060 [foo-768] ERROR reactor.core.scheduler.Schedulers - Scheduler worker in group main failed with an uncaught exception
reactor.blockhound.BlockingOperationError: Blocking call! java.lang.Thread.yield
	at java.lang.Thread.yield(Thread.java)
	at java.util.concurrent.FutureTask.handlePossibleCancellationInterrupt(FutureTask.java:336)
	at java.util.concurrent.FutureTask.run(FutureTask.java:284)
	at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.access$201(ScheduledThreadPoolExecutor.java:180)
	at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:293)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at java.lang.Thread.run(Thread.java:750)
13:02:12.173 [foo-1003] ERROR reactor.core.scheduler.Schedulers - Scheduler worker in group main failed with an uncaught exception
reactor.blockhound.BlockingOperationError: Blocking call! java.lang.Thread.yield
	at java.lang.Thread.yield(Thread.java)
	at java.util.concurrent.FutureTask.handlePossibleCancellationInterrupt(FutureTask.java:336)
	at java.util.concurrent.FutureTask.run(FutureTask.java:284)
13:02:12.992 [foo-2386] ERROR reactor.core.scheduler.Schedulers - Scheduler worker in group main failed with an uncaught exception
reactor.blockhound.BlockingOperationError: Blocking call! java.lang.Thread.yield
	at java.lang.Thread.yield(Thread.java)
	at java.util.concurrent.FutureTask.handlePossibleCancellationInterrupt(FutureTask.java:336)
	at java.util.concurrent.FutureTask.run(FutureTask.java:284)
	at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.access$201(ScheduledThreadPoolExecutor.java:180)
[...]

PR #3515, opened 2023-06-22
Allowing Blocking call for handlePossibleCancellationInterrupt

tejavenkatlanka:
fixs #3512

### DIFF
--- a/reactor-core/src/blockHoundTest/java/reactor/core/scheduler/ReactorBlockHoundIntegrationTest.java
+++ b/reactor-core/src/blockHoundTest/java/reactor/core/scheduler/ReactorBlockHoundIntegrationTest.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2019-2022 VMware Inc. or its affiliates, All Rights Reserved.
+ * Copyright (c) 2019-2023 VMware Inc. or its affiliates, All Rights Reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/reactor-core/src/blockHoundTest/java/reactor/core/scheduler/ReactorBlockHoundIntegrationTest.java
+++ b/reactor-core/src/blockHoundTest/java/reactor/core/scheduler/ReactorBlockHoundIntegrationTest.java
@@ -29,7 +29,7 @@ import reactor.core.Disposable;
 import reactor.core.publisher.Mono;
 import reactor.test.util.RaceTestUtils;
 
-@Timeout(5)
+@Timeout(10)
 public class ReactorBlockHoundIntegrationTest {
 
 	static {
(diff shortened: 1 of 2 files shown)


reactor-core/src/blockHoundTest/java/reactor/core/scheduler/ReactorBlockHoundIntegrationTest.java
chemicL, line 1: Consider Bumping the timeout to 10s to eliminate the build failures. I think 5s might not be enough on lower end machines which the CI might run these on:

    @Timeout(10)
    public class ReactorBlockHoundIntegrationTest {

tejavenkatlanka: Thanks @chemicL, updated timeout to 10s but test case took nearly 1hr 10 mins. 
Please let me know if we can improve.

OlegDokuka (approved):
the test should not be touched at all. `ReactorBlockHoundIntegration` it selves has to be edited with 

```java
allowBlockingCallsInside("java.util.concurrent.FutureTask",
				          "handlePossibleCancellationInterrupt")
```

OlegDokuka (changes requested):
the test should not be touched at all. ReactorBlockHoundIntegration it selves has to be edited with

allowBlockingCallsInside("java.util.concurrent.FutureTask",
				          "handlePossibleCancellationInterrupt")

tejavenkatlanka:
@OlegDokuka   I have reverted the test changes.
Please check

reactor-core/src/blockHoundTest/java/reactor/core/scheduler/ReactorBlockHoundIntegrationTest.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2019-2022 VMware Inc. or its affiliates, All Rights Reserved.
```
OlegDokuka, line 2: this should also be reverted

reactor-core/src/main/java/reactor/core/scheduler/ReactorBlockHoundIntegration.java
```diff
@@ -50,5 +50,7 @@ public void applyTo(BlockHound.Builder builder) {
         // For now, let's not add a separate integration, but rather let's define the class name manually
         // ContextRegistry reads files as part of the Service Loader aspect. If class is initialized in a non-blocking thread, BlockHound would complain
         builder.allowBlockingCallsInside("reactor.core.publisher.ContextPropagation", "<clinit>");
+        builder.allowBlockingCallsInside("java.util.concurrent.FutureTask",
```
OlegDokuka, line 53: use `FutureTask.class.getName()` instead of `"java.util.concurrent.FutureTask"`

OlegDokuka (approved):

reactor-core/src/blockHoundTest/java/reactor/core/scheduler/ReactorBlockHoundIntegrationTest.java
tejavenkatlanka: Thanks @OlegDokuka  reverted this change

reactor-core/src/main/java/reactor/core/scheduler/ReactorBlockHoundIntegration.java
tejavenkatlanka: Update with class reference

tejavenkatlanka:
I've made required changes. @OlegDokuka please check

OlegDokuka (approved):

OlegDokuka:
@tejavenkatlanka great job! Thanks you for your contribution!

PR #3517, opened 2023-06-26
enforces ordering on concurrent subscription to `FluxRefCount`

### LINKED ISSUE #3487: refCount gets hung : race conditions in FluxRefCount#subscribe

OlegDokuka:
Late subscribe may observe zero events when multiple inners subscribe to `FluxRefCount` simultaneously so the one which does not invoke the `connect` method subscribes to the `FluxPublish` too late (e.g. the one which calls the `connect` subscribe faster and invoke the `connect` method earlier than the others):

```
 subscribe 1           subscriber 2
     |                      |
count + 1 (1)               |
     |                 count + 1 (2)
     |                      |
     |                  subscribe
     |                      |
     |                   connect
     |                  onComplete
  subscribe
  (hanging)
```

to solve the problem subscribe should happen before the count is incremented so we have strong ordering guaranteed by the `synchronize` block

fixes #3487

### DIFF
--- a/reactor-core/src/jcstress/java/reactor/core/publisher/FluxPublishStressTest.java
+++ b/reactor-core/src/jcstress/java/reactor/core/publisher/FluxPublishStressTest.java
@@ -171,6 +171,70 @@ public abstract class FluxPublishStressTest {
 	}
 
 	@JCStressTest
+	@Outcome(id = {"0, 0, 0, 0, 1, 1"}, expect = ACCEPTABLE, desc = "concurrent subscription succeeded")
+	@State
+	public static class RefCntConcurrentSubscriptionErrorAsyncStressTest extends
+	                                                                     RefCntConcurrentSubscriptionBaseStressTest<Object> {
+
+		static final Throwable testError = new RuntimeException("boom");
+
+		public RefCntConcurrentSubscriptionErrorAsyncStressTest() {
+			super(Flux.error(testError).publishOn(Schedulers.immediate()), null);
+		}
+
+		@Actor
+		public void subscribe1() {
+			sharedSource.subscribe(subscriber1);
+		}
+		@Actor
+		public void subscribe2() {
+			sharedSource.subscribe(subscriber2);
+		}
+
+		@Arbiter
+		public void arbiter(IIIIII_Result r) {
+			r.r1 = subscriber1.onNextCalls.get();
+			r.r2 = subscriber2.onNextCalls.get();
+			r.r3 = subscriber1.onCompleteCalls.get();
+			r.r4 = subscriber2.onCompleteCalls.get();
+			r.r5 = subscriber1.onErrorCalls.get();
+			r.r6 = subscriber2.onErrorCalls.get();
+		}
+	}
+
+	@JCStressTest
+	@Outcome(id = {"0, 0, 0, 0, 1, 1"}, expect = ACCEPTABLE, desc = "concurrent subscription succeeded")
+	@State
+	public static class RefCntConcurrentSubscriptionErrorNoneStressTest extends
+	                                                                    RefCntConcurrentSubscriptionBaseStressTest<Object> {
+
+		static final Throwable testError = new RuntimeException("boom");
+
+		public RefCntConcurrentSubscriptionErrorNoneStressTest() {
+			super(Flux.error(testError).hide(), null);
+		}
+
+		@Actor
+		public void subscribe1() {
+			sharedSource.subscribe(subscriber1);
+		}
+		@Actor
+		public void subscribe2() {
+			sharedSource.subscribe(subscriber2);
+		}
+
+		@Arbiter
+		public void arbiter(IIIIII_Result r) {
+			r.r1 = subscriber1.onNextCalls.get();
+			r.r2 = subscriber2.onNextCalls.get();
+			r.r3 = subscriber1.onCompleteCalls.get();
+			r.r4 = subscriber2.onCompleteCalls.get();
+			r.r5 = subscriber1.onErrorCalls.get();
+			r.r6 = subscriber2.onErrorCalls.get();
+		}
+	}
+
+	@JCStressTest
 	@Outcome(id = {"10, 10, 1, 1, 0, 0"}, expect = ACCEPTABLE, desc = "concurrent subscription succeeded")
 	@State
 	public static class RefCntGraceConcurrentSubscriptionRangeAsyncFusionStressTest extends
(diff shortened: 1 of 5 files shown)


reactor-core/src/jcstress/java/reactor/core/publisher/FluxPublishStressTest.java
chemicL, line 1: Perhaps adding corresponding cases for `Flux.error(Throwable)` in the fusion/hide/delay variants to also validate how error is handled could be worthwhile?

reactor-core/src/main/java/reactor/core/publisher/FluxRefCount.java
```diff
@@ -162,26 +165,33 @@ static final class RefCountInner<T>
...
-				AtomicIntegerFieldUpdater.newUpdater(RefCountInner.class, "parentDone");
+		Throwable error;
+
+
+		static final int MONITOR_SET_FLAG = 0b0010_0000_0000_0000_0000_0000_0000_0000;
+		static final int TERMINATED_FLAG  = 0b0100_0000_0000_0000_0000_0000_0000_0000;
+		static final int CANCELLED_FLAG   = 0b1000_0000_0000_0000_0000_0000_0000_0000;
 
-		RefCountInner(CoreSubscriber<? super T> actual, RefCountMonitor<T> connection) {
+		volatile int state; //used to guard against doubly terminating subscribers (e.g. double cancel)
```
chemicL, line 180: I suppose the comment is no longer relevant, as the state represents more situations and before it was just about termination/cancellation, while now it also represents the readiness (monitor set). Perhaps we can remove it.

gowa:
@OlegDokuka , if you wish to know whether the tests I wrote pass with your fix, I can confirm they do! :)

Thanks!

reactor-core/src/main/java/reactor/core/publisher/FluxRefCount.java
OlegDokuka: fixed

reactor-core/src/jcstress/java/reactor/core/publisher/FluxPublishStressTest.java
OlegDokuka: fixed

chemicL (approved):

Issue #3539, opened 2023-07-18
takeUntil(Predicate) tests after element has been emitted


AramMessdaghi9001:
takeUntil(Predicate) executes the predicate test after the element has been emitted.
If the element is mutable and modified afterwards, this changes the behavior in an unexpected way.

## Expected Behavior
The predicate test should be executed before the element has been emitted, not after it has been emitted.
Of course it should still be emitted when the predicate test returns false, in order to match the requirement.

## Actual Behavior
The predicate test is executed after the element has been consumed and possibly been modified by following publishers.

## Steps to Reproduce
This is a simplified version
```kotlin
    @Test
    fun `takeUntil(Predicate) tests after element has been emitted`() {
        data class Test(var foo: Int)

        StepVerifier.create(Flux.just(Test(1), Test(2), Test(3))
            .takeUntil { it.foo == 0 }
            .doOnNext { it.foo = 0 }
        )
            .expectNext(Test(1))
            .expectNext(Test(2))
            .expectNext(Test(3))
            .verifyComplete()
    }
```

Here is a more complex scenario:
```kotlin
private fun AsynchronousFileChannel.readBytes(bufferSize: Int, position: Long) : Mono<ByteBuffer> =
    Mono.create { sink ->
        sink.onRequest {
            val byteBuffer = ByteBuffer.allocate(bufferSize)
            read(byteBuffer, position, null, object: CompletionHandler<Int, Nothing?>{
                override fun completed(bytesRead: Int, attachment: Nothing?) = sink.success(byteBuffer.flip())
                override fun failed(exc: Throwable, attachment: Nothing?) = sink.error(exc)
            })
        }
    }

fun Path.readFileAsFlux(bufferSize : Int = 4096) : Flux<ByteBuffer> =
    Flux.using({AsynchronousFileChannel.open(this, StandardOpenOption.READ)}, { fileChannel ->
        Flux.range(0, Int.MAX_VALUE)
            .concatMap { readIndex -> fileChannel.readBytes(bufferSize, readIndex.toLong()*bufferSize) }
            .takeUntil { it.remaining() != bufferSize }
    }, AsynchronousFileChannel::close)
```
Consuming the Flux<ByteBuffer> will not read the file to the end, as `remaining` will be 0 at time of the predicate test, due to being read already by the consuming Flux

## Possible Solution
Execute the predicate.test in FluxTakeUntil before actual.onNext

## Your Environment
* Reactor version(s) used: 3.4.29

[...]

PR #3544, opened 2023-07-24
fix reactor#3539 takeUntil Predicate test before emit

AramMessdaghi9001:
This fixes issue #3539 
The documentation does not specify when takeUntil tests the predicate.
The current implementation does this after the element has been emitted. If the emitted element is modified by downstream consumers, this affects the predicate function.
This behavior might be confusing and lead to unwanted results.
This PR moves the test up and documents the behavior in order to avoid this caveat.

### DIFF
--- a/reactor-core/src/test/java/reactor/core/publisher/FluxTakeUntilPredicateTest.java
+++ b/reactor-core/src/test/java/reactor/core/publisher/FluxTakeUntilPredicateTest.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2016-2021 VMware Inc. or its affiliates, All Rights Reserved.
+ * Copyright (c) 2016-2023 VMware Inc. or its affiliates, All Rights Reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -23,6 +23,7 @@ import reactor.core.CoreSubscriber;
 import reactor.core.Scannable;
 import reactor.test.subscriber.AssertSubscriber;
 
+import java.util.concurrent.atomic.AtomicInteger;
 import static org.assertj.core.api.Assertions.assertThatExceptionOfType;
 import static org.junit.jupiter.api.Assertions.assertThrows;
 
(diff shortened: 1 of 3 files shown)


OlegDokuka (approved):

reactor-core/src/test/java/reactor/core/publisher/FluxTakeUntilPredicateTest.java
```diff
@@ -160,8 +176,7 @@ public void predicateThrows() {
 			throw new RuntimeException("forced failure");
 		}).subscribe(ts);
 
-		ts.assertValues(1)
-		  .assertNotComplete()
+		ts.assertNotComplete()
```
OlegDokuka, line 179: seems to be unrelated change

OlegDokuka (changes requested):

reactor-core/src/test/java/reactor/core/publisher/FluxTakeUntilPredicateTest.java
AramMessdaghi9001: Actually the exception in the predicate here now happens before the element "1" is emitted, so "1" never gets out.

chemicL: I have doubts here. I understand the reasoning, but this is a potentially breaking change if anyone relied on this behaviour. The behaviour is not backwards compatible with regards to error handling. Assuming there is code out there that actually deals with the delivery of the item that then causes an exception when tested, we should preserve this behaviour, or address this change for 3.6 with a clear explanation this is breaking.

AramMessdaghi9001: The later one would be cleaner, wouldn't it? Shall I rebase? Whats the procedure on breaking changes?

OlegDokuka: I don't think there is breaking change unless there is a bug in the given test, even though predicate result is false value is delivered since return statement happens after the a.onNext. I will have a closer look to figure out what is wrong with the test

OlegDokuka: the onNext method should look like the following.

```java
@Override
		public void onNext(T t) {
			if (done) {
				return;
			}

			boolean b;

			try {
				b = predicate.test(t);
			} catch (Throwable e) {
				// ensures value is delivered regardless the failure
				actual.onNext(t);
				onError(Operators.onOperatorError(s, e, t, actual.currentContext()));
				return;
			}

			actual.onNext(t);

			if (b) {
				s.cancel();

				onComplete();
			}
		}
```

having it like that will make the test pass. We have to preserve original behaviour

OlegDokuka (approved):

chemicL (approved):

OlegDokuka:
@AramMessdaghi9001 thank you for your contribution! Keep it up!

PR #3566, opened 2023-08-23
dedicated loom oriented boundeElasticScheduler implementation

OlegDokuka:
This PR provides a dedicated version of bounded elastic which satisfies no VirtualThread reuse recommendation. This implementation creates a new Thread per dedicated task. This implementation does not use any existing ExecutorService implementations underneath so it implements own offloading mechanism. Also, this implementation does not implement any scheduling mechanics and use shared ScheduledExecutorService for tasks that requires a delay or periodic execution. Once task is called by shared ScheduledExecutorService it is back to the Worker queue and eventually offloads to a dedicated VirtualThread


reactor-core/src/main/java/reactor/core/scheduler/LoomBoundedElasticScheduler.java
```diff
@@ -0,0 +1,73 @@
...
+
+final class LoomBoundedElasticScheduler
+		implements Scheduler, SchedulerState.DisposeAwaiter<LoomBoundedElasticScheduler.BoundedServices>, Scannable {
+
+	LoomBoundedElasticScheduler(int maxThreads, int maxTaskQueuedPerThread, ThreadFactory threadFactory, long ttlMillis, Clock clock) {
+		throw new UnsupportedOperationException("Unsupported in JDK lower thank 21");
+	}
+
+	/**
+	 * Create a {@link BoundedElasticScheduler} with the given configuration. Note that backing threads
```
chemicL, line 36: I suppose we don't need the javadoc here as it's not a public class. Also, it's not accurate, mentioning executors.

reactor-core/src/main/java21/reactor/core/scheduler/LoomBoundedElasticScheduler.java
```diff
@@ -0,0 +1,1394 @@
...
+			Thread t = new Thread(r, Schedulers.BOUNDED_ELASTIC + "-evictor-" + EVICTOR_COUNTER.incrementAndGet());
+			t.setDaemon(true);
+			return t;
+		};
+
+
+		final LoomBoundedElasticScheduler parent;
+		//duplicated Clock field from parent so that SHUTDOWN can be instantiated and partially used
+		final Clock                       clock;
+		final ScheduledExecutorService    evictor;
```
chemicL, line 466: We probably don't need the evictor, as we no longer have a heavy `Thread` to release, we could just rely on garbage collection and re-create a new `BoundedState` every time it is required.

reactor-core/src/main/java21/reactor/core/scheduler/LoomBoundedElasticScheduler.java
```diff
@@ -0,0 +1,1394 @@
...
+		final Deque<BoundedState>         idleQueue;
```
chemicL, line 467: We probably don't need the idle queue, we can just rely on the currently used boundedStates if the maxThreads capacity is at limits.

reactor-core/src/main/java21/reactor/core/scheduler/LoomBoundedElasticScheduler.java
```diff
@@ -0,0 +1,1394 @@
...
+			BoundedState bs;
+			while ((bs = idleQueue.pollLast()) != null) {
+				toAwait.add(bs);
+			}
+			Collections.addAll(toAwait, arr);
+			return toAwait.toArray(new BoundedState[0]);
+		}
+	}
+
+	static class BoundedState extends CountDownLatch implements Worker, Disposable, Scannable {
```
chemicL, line 696: The `BoundedState` probably needs to be decoupled from `Worker` in order to track individual `Worker`'s tasks and their disposal, not to cancel other `Worker`'s tasks if one is disposed.

reactor-core/src/main/java21/reactor/core/scheduler/LoomBoundedElasticScheduler.java
```diff
@@ -0,0 +1,1394 @@
...
+				for (;;) {
+					if (isDisposedNow(this.markCount)) {
+						clearAllTask();
+						countDown();
+						return;
+					}
+
+					final ScheduledDisposable task = q.poll();
+
+					if (task == null) {
```
chemicL, line 977: This can potentially lead to freeing up the queue and removing the boundedState if no Worker has a reference any more.

chemicL (changes requested):
In addition to the code comments, we would need jcstress tests for the state transitions when disposing at each granular level (worker/scheduler/boundedState/task) and validating that we keep progressing.

Also, a unit test for the sequential ordering of tasks would be beneficial: e.g. schedule 1M integers to be added to a queue with 1M maxThread limit and at the end traverse the queue to validate items are in the original order.

General remark: disposing a worker should not dispose other worker's tasks when they use the same queue.

reactor-core/src/main/java/reactor/core/scheduler/Schedulers.java
```diff
@@ -1000,6 +1008,29 @@ default Scheduler newBoundedElastic(int threadCap, int queuedTaskCap, ThreadFact
 			return new BoundedElasticScheduler(threadCap, queuedTaskCap, threadFactory, ttlSeconds);
 		}
 
+		/**
+		 * {@link Scheduler} that dynamically creates a bounded number of
+		 * Workers, reusing them once the Workers have been shut down. The underlying (user or daemon)
+		 * threads can be evicted if idle for more than {@code ttlSeconds}.
+		 * <p>
+		 * The maximum number of created thread pools is bounded by the provided {@code cap}.
```
chemicL, line 1016:
```suggestion
		 * The maximum number of created thread pools is bounded by the provided {@code threadCap}.
```

reactor-core/src/main/java/reactor/core/scheduler/Schedulers.java
```diff
@@ -1000,6 +1008,29 @@ default Scheduler newBoundedElastic(int threadCap, int queuedTaskCap, ThreadFact
...
+		 * <p>
+		 * The main difference between {@link BoundedElasticScheduler} and
+		 * {@link LoomBoundedElasticScheduler} is that underlying workers allocates a
```
chemicL, line 1019:
```suggestion
		 * {@link LoomBoundedElasticScheduler} is that underlying workers allocate a
```

reactor-core/src/main/java/reactor/core/scheduler/Schedulers.java
```diff
@@ -1000,6 +1008,29 @@ default Scheduler newBoundedElastic(int threadCap, int queuedTaskCap, ThreadFact
...
+		 * new thread for every new task which is the main requirements for
```
chemicL, line 1020:
```suggestion
		 * new thread for every new task which is the main requirement for
```

reactor-core/src/main/java/reactor/core/scheduler/Schedulers.java
```diff
@@ -1000,6 +1008,29 @@ default Scheduler newBoundedElastic(int threadCap, int queuedTaskCap, ThreadFact
...
+		 * {@link VirtualThread}s
+		 *
+		 * @param threadCap maximum number of underlying threads to create
```
chemicL, line 1023: I'd go with a different wording: "maximum number of simultaneously executing VirtualThreads"

reactor-core/src/main/java/reactor/core/scheduler/Schedulers.java
```diff
@@ -1000,6 +1008,29 @@ default Scheduler newBoundedElastic(int threadCap, int queuedTaskCap, ThreadFact
...
+		 * @param queuedTaskCap maximum number of tasks to enqueue when no more threads can be created. Can be {@link Integer#MAX_VALUE} for unbounded enqueueing.
+		 * @param threadFactory a {@link ThreadFactory} to use each thread initialization
+		 *
+		 * @return a new {@link Scheduler} that dynamically creates workers with an upper bound to
+		 * the number of backing threads, reuses threads and evict idle ones
```
chemicL, line 1028:
```suggestion
		 * the number of backing threads
```

chemicL: This should not be the case what this comment says

reactor-core/src/test/java21/reactor/core/scheduler/LoomBoundedElasticSchedulerTest.java
```diff
@@ -0,0 +1,267 @@
...
+
+		Assertions.assertThat(latch.await(5, TimeUnit.SECONDS)).isTrue();
+		Assertions.assertThat(disposable.isDisposed()).isTrue();
+	}
+
+	@Test
+	public void ensuresTasksDelayedScheduling() throws InterruptedException {
+		CountDownLatch latch = new CountDownLatch(1);
+
+		Disposable disposable = scheduler.schedule(latch::countDown, 200, TimeUnit.MILLISECONDS);
```
chemicL, line 48: 200ms is actually quite a lot, can we go with, e.g. 5ms ? As this is an actual clock, not a dummy...

reactor-core/src/test/java21/reactor/core/scheduler/LoomBoundedElasticSchedulerTest.java
```diff
@@ -0,0 +1,267 @@
...
+				TimeUnit.MILLISECONDS);
+
+		Assertions.assertThat(latch.await(5, TimeUnit.SECONDS)).isTrue();
+		Assertions.assertThat(disposable.isDisposed()).isFalse();
+		disposable.dispose();
+		Assertions.assertThat(disposable.isDisposed()).isTrue();
+	}
+
+	@Test
+	public void ensuresTasksPeriodicZeroInitialDelayScheduling() throws InterruptedException {
```
chemicL, line 80: This and the similar tests don't really validate whether the time is applied correctly. Can we do something about validating the arguments have any actual impact? The assertions stay the same, they should be different if the arguments differ.

reactor-core/src/test/java21/reactor/core/scheduler/LoomBoundedElasticSchedulerTest.java
```diff
@@ -0,0 +1,267 @@
...
+				          .isEqualTo(queue.poll());
+				Assertions.assertThat(value)
+				          .isEqualTo(queue.poll());
+			}
+			worker.dispose();
+		}
+	}
+
+	@Test
+	public void ensuresConcurrentWorkerTaskDisposure() throws InterruptedException {
```
chemicL, line 244: This one is nice :) Apart from the existing cases, I think it would be great to test:
- schedule adding numbers 1..10k (one by one) to a queue and validate the queue contains numbers in the same order (validates single task is executed and the backing queue holds the tasks in proper order)
- two workers sharing the same BoundedServices instance and disposing one doesn't dispose tasks of the other.

reactor-core/src/main/java21/reactor/core/scheduler/LoomBoundedElasticScheduler.java
OlegDokuka: fixed

reactor-core/src/main/java21/reactor/core/scheduler/LoomBoundedElasticScheduler.java
OlegDokuka: fixed

reactor-core/src/main/java21/reactor/core/scheduler/LoomBoundedElasticScheduler.java
OlegDokuka: fixed

reactor-core/src/main/java21/reactor/core/scheduler/LoomBoundedElasticScheduler.java
```diff
@@ -0,0 +1,1352 @@
...
+		}
+
+		/**
+		 * Is this {@link BoundedState} still in use by workers.
+		 *
+		 * @return true if in use, false if it has been disposed enough times
+		 */
+		@Override
+		public boolean isDisposed() {
+			return isDisposed(this.markCount);
```
chemicL, line 893: the argument should be the state, not markCount, or I am missing something?

reactor-core/src/main/java21/reactor/core/scheduler/LoomBoundedElasticScheduler.java
```diff
@@ -0,0 +1,1352 @@
...
+		}
+
+		void tryForceSchedule() {
+			int previousState = forceAddWork(this);
+			if (previousState != 0) {
+				return;
+			}
+
+			drain();
+		}
```
chemicL, line 831: These two methods are used in different contexts. I think I understood when you explained it to me, but now I don't remember. I think it is worth documenting the inner behaviour properly for future maintainers.

chemicL (changes requested):

reactor-core/src/main/java21/reactor/core/scheduler/LoomBoundedElasticScheduler.java
```diff
@@ -0,0 +1,1352 @@
...
+			if (len == 1) {
+				return arr[0];
+			}
+
+			BoundedState choice = arr[0];
+			int leastBusy = Integer.MAX_VALUE;
+
+			for (int i = 0; i < arr.length; i++) {
+				BoundedState state = arr[i];
+				int busy = state.markCount;
```
chemicL, line 544: Consider using pending tasks queue size instead. Probably in a lot of cases the number of users of a boundedState will be similar, but the queue size is our current best tool to measure how busy the boundedState is.

reactor-core/src/main/java21/reactor/core/scheduler/LoomBoundedElasticScheduler.java
```diff
@@ -0,0 +1,1384 @@
...
+				// In both cases we throw an exception, as the caller of init() should
+				// expect the scheduler to be running when this method returns.
+				throw new IllegalStateException(
+						"Scheduler disposed during initialization"
+				);
+			}
+		}
+		else {
+			b.currentResource.evictor.shutdownNow();
+			// Currently, isDisposed() is true for non-initialized state, but that will
```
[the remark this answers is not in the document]
OlegDokuka: We can do that as a separate effort

reactor-core/src/main/java/reactor/core/scheduler/Schedulers.java
OlegDokuka: We don't have the direct binding to VirtualThreads here although the name suggests Loom

OlegDokuka: renamed class

reactor-core/src/main/java21/reactor/core/scheduler/LoomBoundedElasticScheduler.java
```diff
@@ -0,0 +1,1352 @@
...
+					return;
+				}
+				if (BUSY_STATES.compareAndSet(this, current, replacement)) {
+					return;
+				}
+			}
+		}
+
+		/**
+		 * Pick a {@link BoundedState}, prioritizing idle ones then spinning up a new one if enough capacity.
```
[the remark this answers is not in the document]
OlegDokuka: Same story for the original BoundedElasticScheduler

reactor-core/src/main/java21/reactor/core/scheduler/LoomBoundedElasticScheduler.java
```diff
@@ -0,0 +1,1352 @@
...
+				return null; //implicit retry in the pick() loop
+			}
+			if (len == 1) {
+				return arr[0];
+			}
+
+			BoundedState choice = arr[0];
+			int leastBusy = Integer.MAX_VALUE;
+
+			for (int i = 0; i < arr.length; i++) {
```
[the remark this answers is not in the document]
OlegDokuka: not related to the main effort. Could be a good separate improvement for both the original Bounded and this new one

reactor-core/src/main/java21/reactor/core/scheduler/LoomBoundedElasticScheduler.java
```diff
@@ -0,0 +1,1352 @@
...
+
+	static class BoundedStateWorker implements Worker {
+
+		final Composite    disposables;
+		final BoundedState state;
+
+		BoundedStateWorker(BoundedState state) {
+			this.state = state;
+			this.disposables = Disposables.composite();
+			disposables.add(state);
```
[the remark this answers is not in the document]
OlegDokuka: This is correct behaviour. This mechanism also disposes the given state along with tasks. The dispose is calling release along the way. This is standard mechanism from the original Worker implementations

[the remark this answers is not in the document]
OlegDokuka: Although it can be simplified

reactor-core/src/main/java21/reactor/core/scheduler/LoomBoundedElasticScheduler.java
OlegDokuka: lets review that separately as an effort for both BoundedElasticSchedulers

reactor-core/src/main/java21/reactor/core/scheduler/LoomBoundedElasticScheduler.java
OlegDokuka: let me revisit that

gradle/toolchains.gradle
```diff
@@ -94,6 +94,25 @@ pluginManager.withPlugin("me.champeau.jmh") {
 	}
 }
 
+// Configure the JCStress plugin to use the toolchain for generating and running JCStress bytecode
+pluginManager.withPlugin("io.github.reyerizo.gradle.jcstress") {
+  if (testToolchainConfigured()) {
+	tasks.matching { it.name.contains('jcstress') && it.hasProperty('javaLauncher') }.configureEach {
+	  javaLauncher.set(javaToolchains.launcherFor {
+		languageVersion.set(testToolchainLanguageVersion())
+	  })evrika18
```
chemicL, line 103: `evrika18` looks like a syntax error, can you explain this?

reactor-core/src/main/java/reactor/core/scheduler/Schedulers.java
```diff
@@ -1000,6 +1008,31 @@ default Scheduler newBoundedElastic(int threadCap, int queuedTaskCap, ThreadFact
 			return new BoundedElasticScheduler(threadCap, queuedTaskCap, threadFactory, ttlSeconds);
 		}
 
+		/**
+		 * {@link Scheduler} that dynamically creates a bounded number of
+		 * Workers, reusing them once the Workers have been shut down. The underlying (user or daemon)
+		 * threads can be evicted if idle for more than {@code ttlSeconds}.
```
chemicL, line 1014: This is still untrue.

reactor-core/src/main/java/reactor/core/scheduler/VirtualThreadFactory.java
```diff
@@ -0,0 +1,53 @@
...
+import java.util.concurrent.atomic.AtomicLong;
+import java.util.function.BiConsumer;
+import java.util.function.Supplier;
+
+import reactor.util.annotation.NonNull;
+import reactor.util.annotation.Nullable;
+
+/**
+ * The noop {@link VirtualThread} Reactor {@link ThreadFactory} to be
+ * used with {@link ThreadPerTaskBoundedElasticScheduler}, throws exceptions when is being created, so it indicates
```
chemicL, line 29:
```suggestion
 * used with {@link ThreadPerTaskBoundedElasticScheduler}. It throws exceptions when being created, so it indicates
```

reactor-core/src/main/java/reactor/core/scheduler/VirtualThreadFactory.java
```diff
@@ -0,0 +1,53 @@
...
+ * that current Java Runtime does not support {@link VirtualThread}s.
```
chemicL, line 30: Does the `{@link VirtualThread}` work? `{@code VirtualThread}` perhaps would be better?

reactor-core/src/test/java/reactor/core/scheduler/GenericThreadPerTaskBoundedElasticSchedulerTest.java
```diff
@@ -0,0 +1,86 @@
...
+		}
+		catch (Throwable t) {
+			SUPPORTED = false;
+		}
+	}
+
+	@Override
+	protected Scheduler scheduler() {
+		Assumptions.assumeThat(SUPPORTED).isTrue();
+		ThreadPerTaskBoundedElasticScheduler test =
```
chemicL, line 42: The `freshScheduler()` method could be called here.

reactor-core/src/main/java21/reactor/core/scheduler/ThreadPerTaskBoundedElasticScheduler.java
```diff
@@ -0,0 +1,1351 @@
...
+	 * @return a best effort total count of the spinned up executors
+	 */
+	int estimateSize() {
+		return state.currentResource.activeExecutorsState.array.length;
+	}
+
+	/**
+	 * Best effort snapshot of the remaining queue capacity for pending tasks across all the backing executors.
+	 *
+	 * @return the total task capacity, or {@literal -1} if any backing executor's task
```
chemicL, line 277: we don't return -1 any more

reactor-core/src/test/java/reactor/core/scheduler/GenericThreadPerTaskBoundedElasticSchedulerTest.java
OlegDokuka: no need to do so. Just following the overall pattern that is makes no sense

reactor-core/src/main/java/reactor/core/scheduler/VirtualThreadFactory.java
OlegDokuka: docs are generated fine

reactor-core/src/test/java/reactor/core/scheduler/GenericThreadPerTaskBoundedElasticSchedulerTest.java
OlegDokuka: fixed

reactor-core/src/main/java21/reactor/core/scheduler/ThreadPerTaskBoundedElasticScheduler.java
OlegDokuka: fixed

reactor-core/src/main/java/reactor/core/scheduler/VirtualThreadFactory.java
OlegDokuka: fixed

gradle/toolchains.gradle
OlegDokuka: fixed

chemicL (approved):

PR #3583, opened 2023-09-22
removes context related deprecations

OlegDokuka:
This PR removes/changes the following deprecated methods: 

`reactor.core.publisher.FluxSink#currentContext()` -> removed, `FluxSink#contextView()` should be used instead.
`reactor.core.publisher.FluxSink#contextView()` -> set from default to abstract.

 `reactor.core.publisher.MonoSink#currentContext()` -> removed, `MonoSink#contextView()` should be used instead.
`reactor.core.publisher.MonoSink#contextView()` -> set from default to abstract.

`reactor.core.publisher.SynchronousSink#currentContext()` -> removed, `SynchronousSink#contextView()` should be used instead.
`reactor.core.publisher.SynchronousSink#contextView()` -> set from default to abstract.

` reactor.util.context.Context#putAll(reactor.util.context.Context)` -> removed. `Context#putAll(ContextView)` should be used instead


chemicL:
This hasn't been integrated at the time it was proposed. I'm closing this PR - we can circle back and remove these deprecated methods once they become a burden or we aim for 4.x of reactor-core.

PR #3588, opened 2023-09-26
removes operators related deprecations

OlegDokuka:


chemicL:
This hasn't been integrated at the time it was proposed. We can re-consider removing these deprecated methods once they become a burden or 4.x is planned.

Issue #3545, opened 2023-07-26
Repeat specification for use with repeatWhen operator


chemicL:
Similarly to the way `Retry` is implemented with accompanying `retryWhen(RetrySpec)` operator, the `Repeat` functionality from reactor-addons project should be implemented in reactor-core and phased out from reactor-extras (module of reactor-addons). It can potentially reuse the primitives that already exist for the retry functionality.

Desislav-Petrov:
Hi @chemicL  - i'm keen to explore this one - could you please assign to me? Thanks

chemicL:
@OlegDokuka can you please respond when you find a moment? (I'm going to be away for a few days)

OlegDokuka:
@Desislav-Petrov we don't literally have to port impl from reactor-addons but rather create a shared common for both retry and repeat operators. That means the logic implemented in Retry builder can go into something like "CommonRecurring" (or whatever name we endup) and then implement Repeat builder having that base mechanism as well as Retry.

Desislav-Petrov:
Got you - thanks, will have a go

kwondh5217:
Hi @chemicL!  
I'd love to contribute to this issue. Could you please assign it to me?

chemicL:
@kwondh5217 assigned, let me know if you need anything.

kwondh5217:
Hi @chemicL,

Before proceeding with the implementation, I would like to check a few points:

- What features should RepeatSpec (and CommonRecurring for Repeat) support? (e.g., transientErrors, backoff, etc.).
  I understand that we don't necessarily have to port the implementation from reactor-addons literally.

- Is the scope of this issue to implement Repeat, RepeatSpec, and integrate them into FluxRepeatWhen and MonoRepeatWhen?

- Should we modify and reuse the existing FluxRepeatWhen and MonoRepeatWhen, or should we consider reimplementing them?

Also, if there are any additional points I should be aware of, I would appreciate it if you could let me know.

Thank you!

chemicL:
Hey, I took a break, catching up here.

> * What features should RepeatSpec (and CommonRecurring for Repeat) support? (e.g., transientErrors, backoff, etc.).
>   I understand that we don't necessarily have to port the implementation from reactor-addons literally.

I think we should port the features that are present in addons.
I think transientErrors are a concept that makes sense for retries, not for repeat functionality.
What do you mean by CommonRecurring?

[...]

PR #3593, opened 2023-10-01
Repeat specification for repeatWhen

Desislav-Petrov:
hi @chemicL and @OlegDokuka - kicked off some work for https://github.com/reactor/reactor-core/issues/3545 but wanted to verify the approach before i go any further.

So my plan is as follows.

1) Work out a base approach to get the simple repeat test pass
2) Add necessary functionality for the rest of the tests to pass
3) Refactor and pull the common bits between repeat/retry as previously suggested
4) See if anything else is missing in the repeat functionality and add it

### DIFF
--- a/reactor-core/src/main/java/reactor/util/repeat/RepeatSpec.java
+++ b/reactor-core/src/main/java/reactor/util/repeat/RepeatSpec.java
@@ -1,0 +1,39 @@
+package reactor.util.repeat;
+
+import reactor.core.publisher.Flux;
+import reactor.core.publisher.Mono;
+import reactor.util.Logger;
+import reactor.util.Loggers;
+import reactor.util.context.ContextView;
+
+import java.util.function.Predicate;
+
+public class RepeatSpec extends Repeat {
+    static final Logger log = Loggers.getLogger(RepeatSpec.class);
+    private final long repeats;
+
+    RepeatSpec(Predicate<? super ContextView> repeatPredicate, long repeats) {
+        this.repeats = repeats;
+    }
+
+    static Repeat create(Predicate<? super ContextView> predicate, long times) {
+        return new RepeatSpec(predicate, times);
+    }
+
+    @Override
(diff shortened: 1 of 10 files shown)


reactor-core/src/main/java/reactor/util/repeat/RepeatSpec.java
```diff
@@ -0,0 +1,39 @@
...
+
+    @Override
+    public Flux<Long> generateCompanion(Flux<RepeatSignal> other) {
+        return Flux.deferContextual(cv ->
+                other
+                        .contextWrite(cv)
+                        .concatMap(retryWhenState -> {
+                            RepeatSignal copy = retryWhenState.copy();
+                            long iteration = copy.getRepeatsSoFar();
+                            if (iteration > repeats) {
```
Desislav-Petrov, line 31: I don't think this is the most graceful way to do this, was trying to get some inspiration from the Retry one but the case is slightly different there

chemicL:
Hey, @Desislav-Petrov. I'm trying to go through some PRs that were not given enough attention. Please accept the team's apologies. I had a brief look and would like to know whether after not gaining much attention from us you are still willing to make progress on this.

My first concern is a public API change - we need to make sure that anything that we add does not remove existing functionality.

Also, the next(-1) signal looks suspicious - in case of differences between the implementation in reactor-extras, can you list them with a high-level explanation?

Let me know, thanks.

Desislav-Petrov:
Hey @chemicL - I'm still keen on working on that as long as I get some support/gudaince - let me take a look since it's been a while to summarise the idea.

chemicL:
@Desislav-Petrov take your time. I'll try to be helpful. Should I be occupied to the limits with other concerns, I'll be sure to let you know.

chemicL:
@Desislav-Petrov it's been a month since my last message. Let's reopen if you think you'll find the time to make some progress.

PR #3599, opened 2023-10-02
Fix "Blocking Mono Result" example

pivovarit:
The provided example did not compile for two reasons:

- no matching `Mono.zip()` signature that accepts two `Publishers` (just `Mono`s)
- no `Flux#delay` anymore

Replaced the example with a more idiomatic usage - `Mono.delay()` followed by `then()` instead of `Mono.just()` followed by argument-discarding `map()`. I allowed myself to replace `currentTimeMillis` with `Instant`s so that it's easier to see the difference when printed out

### DIFF
--- a/README.md
+++ b/README.md
@@ -133,12 +133,11 @@ Mono.fromCallable(System::currentTimeMillis)
 ```
 
 Blocking Mono result :
-```java    
-Tuple2<Long, Long> nowAndLater = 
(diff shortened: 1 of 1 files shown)


chemicL (approved):
Thanks for the PR @pivovarit :) I know it has been open for some time but I finally got to it and it is still helpful so I'll merge it now.

PR #3600, opened 2023-10-03
Fix "Flux.create" example

pivovarit:
The example used the old `timeout()` method with the number of seconds as a parameter. 

Replaced it with the new method accepting a `Duration` object.


chemicL (approved):
Thanks for the PR. just for the record: I checked the old Javadocs and I feel there never was a `timeout(Integer)` method so this was just a mistake all along.

PR #3611, opened 2023-10-05
provides extra check for contextualName presence

### LINKED ISSUE #3572: Micrometer observation contextual name is alway overriden
Hi,

According to the documentation: 
https://github.com/reactor/reactor-core/blob/main/docs/asciidoc/metrics-details.adoc

Supplying an observation factory we can configure any attributes of the observation.

However, using contextualName is always overridden by sequence name.  E.g.:
```java
Mono<String> userName(String userId) {
		return Mono.defer(() -> {
			log.info("Getting user name for user with id <{}>", userId);
			return Mono.delay(Duration.ofMillis(random.nextLong(200L)))
					.thenReturn("foo");
			})
				.tap(Micrometer.observation(registry, r -> Observation.createNotStarted("user.name", registry)
						.lowCardinalityKeyValue("userType", "admin")
						.highCardinalityKeyValue("userId", userId)
						.contextualName("getting-user-name")));
	}
```

Results in "reactor.observation" observation/span.

Note: observation can be also created convention, which can also populate name / contextual name.

https://github.com/reactor/reactor-core/blob/5553aa80137482ec26acb960f4d4f42b8a44da94/reactor-core-micrometer/src/main/java/reactor/core/observability/micrometer/MicrometerObservationListener.java#L101

Can we move contextualName setter into defaultObservation in case if there is no custom created observation?

Thanks,
Artyom

OlegDokuka:
this check allows keeping user provided configuration.

closes #3572

### DIFF
--- a/reactor-core-micrometer/src/main/java/reactor/core/observability/micrometer/MicrometerObservationListener.java
+++ b/reactor-core-micrometer/src/main/java/reactor/core/observability/micrometer/MicrometerObservationListener.java
@@ -97,11 +97,11 @@ final class MicrometerObservationListener<T> implements SignalListener<T> {
 		//while doOnSubscription matches the moment where the Publisher acknowledges said subscription
 		//NOTE: we don't use the `DocumentedObservation` features to create the Observation, even for the ANONYMOUS case,
 		//because the discovered tags could be more than the documented defaults
-		tapObservation = defaultObservation(configuration, observationSupplier)
+		tapObservation = supplyOrCreateObservation(configuration, observationSupplier)
 				.lowCardinalityKeyValues(configuration.commonKeyValues);
 	}
 
-	private static Observation defaultObservation(
+	private static Observation supplyOrCreateObservation(
 			MicrometerObservationListenerConfiguration configuration,
 			@Nullable Function<ObservationRegistry, Observation> observationSupplier) {
 		if (observationSupplier != null) {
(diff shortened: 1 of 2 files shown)


reactor-core-micrometer/src/main/java/reactor/core/observability/micrometer/MicrometerObservationListener.java
```diff
@@ -98,7 +98,6 @@ final class MicrometerObservationListener<T> implements SignalListener<T> {
 		//NOTE: we don't use the `DocumentedObservation` features to create the Observation, even for the ANONYMOUS case,
 		//because the discovered tags could be more than the documented defaults
 		tapObservation = defaultObservation(configuration, observationSupplier)
```
chemicL, line 100: The name `defaultObservation` is misleading, it should be changed because of issues like this one, e.g. `getOrCreateObservation`.

reactor-core-micrometer/src/test/java/reactor/core/observability/micrometer/MicrometerObservationListenerTest.java
```diff
@@ -74,6 +74,47 @@ public long monotonicTime() {
...
+	@ParameterizedTestWithName
+	@ValueSource(booleans =  {true, false})
+	void whenStartedFluxWithDefaultNameWithCustomObservationSupplier(boolean automatic) {
+		if (automatic) {
+			Hooks.enableAutomaticContextPropagation();
+		}
+		configuration = new MicrometerObservationListenerConfiguration(
+				MicrometerObservationListenerDocumentation.ANONYMOUS.getName(),
+				//note: "type" key is added by MicrometerObservationListenerConfiguration#fromFlux (which is tested separately)
+				KeyValues.of("testTag1", "testTagValue1","testTag2", "testTagValue2"),
```
chemicL, line 86:
```suggestion
				KeyValues.of("testTag1", "testTagValue1", "testTag2", "testTagValue2"),
```

chemicL (approved):

PR #3614, opened 2023-10-06
ensures that proper index is used during onNext check

### LINKED ISSUE #3525: flaky test `FluxBufferTimeoutTest > flushShouldNotRaceWithNext()`
flaky test. seeing it constantly

```
FluxBufferTimeoutTest > flushShouldNotRaceWithNext() FAILED
    org.opentest4j.AssertionFailedError: 
    expected: 500
     but was: 499
        at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
        at sun.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
        at sun.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
        at reactor.core.publisher.FluxBufferTimeoutTest.flushShouldNotRaceWithNext(FluxBufferTimeoutTest.java:337)
```

OlegDokuka:
closes #3525

### DIFF
--- a/reactor-core/src/test/java/reactor/core/publisher/FluxBufferTimeoutTest.java
+++ b/reactor-core/src/test/java/reactor/core/publisher/FluxBufferTimeoutTest.java
@@ -24,11 +24,13 @@ import java.util.Set;
 import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.concurrent.RejectedExecutionException;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.Consumer;
 
+import org.assertj.core.api.Assertions;
 import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.Test;
 import org.reactivestreams.Subscription;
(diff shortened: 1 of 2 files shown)


reactor-core/src/test/java/reactor/core/publisher/FluxBufferTimeoutTest.java
```diff
@@ -310,31 +312,54 @@ public void scanSubscriberCancelled() {
 
 	@Test
 	public void flushShouldNotRaceWithNext() {
-		Set<Integer> seen = new HashSet<>();
-		Consumer<List<Integer>> consumer = integers -> {
-			for (Integer i : integers) {
-				if (!seen.add(i)) {
-					throw new IllegalStateException("Duplicate! " + i);
+		for (int i = 0; i < 100; i++) {
```
chemicL, line 315: The changes in the test look like were introduced because the race testing utils are not trusted. At the same time, changing a test at the same time while introducing a change to the production code is suspicious - are these the final changes? Do we want to repeat the test 100 times?

chemicL (changes requested):

reactor-core/src/test/java/reactor/core/publisher/FluxBufferTimeoutTest.java
OlegDokuka: it is probability related thing. Just running race test once gives no guarantee of event to occur. JCStress are doing the same underneath (looping many times until a possible event is being triggered). Here, we do that manually. 

Ideally that should be moved to JCStress suit which does that in more sophisticated way + adds different VM configurations in addition which may create higher probability to a specific event occurrence.

OlegDokuka: There should be a separate effort moving all such tests to JCStress. For now just wanna eliminate the bug

chemicL (approved):

PR #3617, opened 2023-10-09
Revised documentation about context propagation

chemicL:
Improving context propagation documentation to include details about the automatic mode.

As a next step, we could also rework `C.8. What Is a Good Pattern for Contextual Logging? (MDC)` to also offer a `ThreadLocalAccessor` to get the job done.

### DIFF
--- a/docs/asciidoc/advanced-contextPropagation.adoc
+++ b/docs/asciidoc/advanced-contextPropagation.adoc
@@ -10,19 +10,21 @@ This library is intended as a means to easily adapt between various implementati
 It implements the SPI and is loaded via `java.util.ServiceLoader`.
 No user action is required, other than having a dependency on both reactor-core and `io.micrometer:context-propagation`. The `ReactorContextAccessor` class is public but shouldn't generally be accessed by user code.
 
-We support two modes of operation with `io.micrometer:context-propagation`:
+Reactor-Core supports two modes of operation with `io.micrometer:context-propagation`:
 
 - the **default** (limited) mode,
-- and the **automatic** mode, enabled via `Hooks.enableAutomaticContextPropagation()`
-when the app starts.
+- and the **automatic** mode, enabled via `Hooks.enableAutomaticContextPropagation()`.
+  Please note that this mode applies only to new subscriptions, so it is recommended to
+enable this hook when the application starts.
 
-We will discuss the differences in the context of either writing data to Reactor `Context` or accessing
-`ThreadLocal` state that reflects the contents of the `Context` of currently attached
-`Subscriber` for reading.
+Their key differences are discussed in the context of either <<context-writing, writing
+data>> to Reactor `Context`, or <<context-accessing,accessing `ThreadLocal` state>> that
+reflects the contents of the`Context` of currently attached `Subscriber` for reading.
 
 The act of discovering the mapping between `ThreadLocal` values and relevant entries in
 the Reactor `Context` is performed using the registered `ThreadLocalAccessor`(s).
 
+[[context-writing]]
 == Writing to `Context`
 
 Depending on the individual application, you might either have to store already populated
@@ -64,7 +66,7 @@ not necessary to add it.
 === `contextWrite` Operator
 
 When the `ThreadLocal` values are not and do not need to be present at the time of
-subscription, we can immediately store the value in the `Context`:
+subscription, they can immediately be stored in the `Context`:
 
 ====
 [source,java]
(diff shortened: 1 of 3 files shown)


docs/asciidoc/advanced-contextPropagation.adoc
```diff
@@ -32,25 +48,80 @@ TL.set("HELLO");
...
 ====
 
-== Operators that transparently restore a snapshot: `handle` and `tap`
-Both `Flux` and `Mono` variants of `handle` and `tap` will have their behavior slightly modified
-if the Context-Propagation library is available at runtime.
+== Accessing `ThreadLocal` state corresponding to `ContextView`
+
+As can be seen in the above examples, in the default mode, the `ThreadLocal` value is
+`null` inside the lambda provided to the `map` operator. This shows a major difference
+between the **default** (limited) mode, and the **automatic** mode.
```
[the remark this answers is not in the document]
OlegDokuka: may be we can say even simpler "In contrast to default mode which does not restore threadLocals throughout the whole chain automatic mode gives consistent access"

docs/asciidoc/advanced-contextPropagation.adoc
```diff
@@ -32,25 +48,80 @@ TL.set("HELLO");
...
+
+=== Automatic mode
```
OlegDokuka, line 93: why do we have automatic and default modes again if the previous sections was talking about default mode already?

docs/asciidoc/advanced-contextPropagation.adoc
```diff
@@ -32,25 +48,80 @@ TL.set("HELLO");
...
 
-Namely, if their downstream `ContextView` is not empty they will assume a context capture has occurred
-(either manually or via the `contextCapture()` operator) and will attempt to restore `ThreadLocal`s from
-that snapshot transparently.
+It is not an easy task to achieve, as the Reactive Streams specification makes reactive
+chains `Thread`-agnostic. However, we did our best to control sources of `Thread` switches
+and perform snapshot restoration based on the Reactor `Context`, which is treated as the
+source of truth for `ThreadLocal` state. In case you are interested, we described how
+we got here in a three-article series on
+https://spring.io/blog/2023/03/28/context-propagation-with-project-reactor-1-the-basics[the spring.io blog].
```
OlegDokuka, line 103: there was not any blog links before although there were dedicated blogs on those topics, thus maybe it is better to follow the same style as before and give brief explanation while a reader can find other sources on their own

docs/asciidoc/advanced-contextPropagation.adoc
OlegDokuka: may be not reference at the beginning about default at all and instead of automatic start with default saying that this mode is available starting from 3.5 and does minimal stuff and then mention automatic which is available in 3.5.9+ and stabilised in 3.6

OlegDokuka (changes requested):
majorly I suggest to change writing style to follow the original one (e.g. instead of we support - reactor core or the library supports). 
Sections should probably assume less connection since one can just jump to a specific section and if it starts with "above example shows" it may confuse since there was no "above examples". Hence, it could be simplified

docs/asciidoc/advanced-contextPropagation.adoc
```diff
@@ -32,25 +48,80 @@ TL.set("HELLO");
...
+.contextWrite(ctx -> ctx.put(TLKEY, "HELLO"))
+.block(); // returns "delayed ctx[TLKEY]=HELLO, TL=null" in default mode
+          // returns "delayed ctx[TLKEY]=HELLO, TL=HELLO" in automatic mode
 ----
 ====
 
-== Operators that transparently restore a snapshot: `handle` and `tap`
-Both `Flux` and `Mono` variants of `handle` and `tap` will have their behavior slightly modified
-if the Context-Propagation library is available at runtime.
+== Accessing `ThreadLocal` state corresponding to `ContextView`
```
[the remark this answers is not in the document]
chemicL: No, not really. The 'ThreadLocal' values reflect what the values stored in `ContextView`. What binds them together are common keys. There are no ThreadLocal values "available in ContextView"..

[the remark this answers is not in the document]
OlegDokuka: it does not read well to me. I would name it differently. E.g. "restoring ThreadLocal values using ContextRegistry and ContextView as value carrier" it is longer but it is simpler to understand

docs/asciidoc/advanced-contextPropagation.adoc
```diff
@@ -5,20 +5,38 @@ Since 3.5.0, Reactor-Core embeds support for the `io.micrometer:context-propagat
...
+
+Depending on the individual application, you might either have to store already populated
+`ThreadLocal` state as entries in the `Context`, or might only need to directly populate
+the `Context`.
+
+=== `contextCapture` Operator
 This operator can be used when one needs to capture `ThreadLocal` value(s) at subscription time and reflect these values in the Reactor `Context` for the benefit of upstream operators.
-It relies on the `context-propagation` library and notably the registered `ThreadLocalAccessor`(s) to discover relevant ThreadLocal values.
 
 This is a convenient alternative to `contextWrite` which uses the `context-propagation` API to obtain a `ContextSnapshot` and then uses that snapshot to populate the Reactor `Context`.
```
OlegDokuka, line 37: btw. Maybe it is me but I reread multiple times until I got that it is about contextCapture and not contextWrite which uses contextRegistry. may be worth rephrasing as well to something like: 

"In opposite to manual `contextWrite` `contextCapture` uses the `context-propagation` API ...`

docs/asciidoc/advanced-contextPropagation.adoc
```diff
@@ -5,20 +5,38 @@ Since 3.5.0, Reactor-Core embeds support for the `io.micrometer:context-propagat
...
+- and the **automatic** mode, enabled via `Hooks.enableAutomaticContextPropagation()`.
+  Please note that this mode applies only to new subscriptions, so it is recommended to
+enable this hook when the application starts.
+
+Their key differences are discussed in the context of either <<context-writing, writing
+data>> to Reactor `Context`, or <<context-accessing,accessing `ThreadLocal` state>> that
+reflects the contents of the`Context` of currently attached `Subscriber` for reading.
+
+The act of discovering the mapping between `ThreadLocal` values and relevant entries in
+the Reactor `Context` is performed using the registered `ThreadLocalAccessor`(s).
```
OlegDokuka, line 25: I would move that section to context-accessing paragraph so it does not duplicate

chemicL:
@OlegDokuka thanks for the feedback! I reordered contextWrite and contextCapture sections and made some readability improvements in the ThreadLocal restoration section - please have a look.

docs/asciidoc/advanced-contextPropagation.adoc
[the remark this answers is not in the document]
chemicL: I removed "corresponding to `ContextView`" and clarified the logistics in the text below.

OlegDokuka (approved):

PR #3623, opened 2023-10-24
ensures SchedulerTask uses isShutdown instead of isDisposed

OlegDokuka:
the SchedulerTask was using the isDisposed which is showing false until both the isShutdown is true and the count of active tasks is zero. That means that any recurring task could not be interrupted since this flag never returns true while this task is active. We need to check only isShutdown which is set to true when gracefulShutdown is triggered so any recurring task can stop eventually


chemicL (approved):
LGTM. Please update the PR with a meaningful title.

OlegDokuka:
@chemicL thanks for the review. title updated

PR #3635, opened 2023-11-10
Improves documentation for VirtualThreads boundedElastic behavior

OlegDokuka:


docs/asciidoc/advancedFeatures.adoc
```diff
@@ -624,6 +624,18 @@ invoked whenever a `Runnable` task submitted to a `Scheduler` throws an `Excepti
 that if there is an `UncaughtExceptionHandler` set for the `Thread` that ran the task,
 both the handler and the hook are invoked).
 
+[[scheduler-virtual-thread]]
```
chemicL, line 627: Please review section 4.5 and consider updating the boundedElastic description instead of adding a new one here. The remark about newBoundedElastic can become part of the above section for factories, so it offers a guidance to the user how to replace the old behaviour with the new one in places that they need a non-global instance.

reactor-core/src/main/java/reactor/core/scheduler/Schedulers.java
```diff
@@ -63,6 +63,8 @@
  * Factories prefixed with {@code new} (eg. {@link #newBoundedElastic(int, int, String)} return a new instance of their flavor of {@link Scheduler},
  * while other factories like {@link #boundedElastic()} return a shared instance - which is the one used by operators requiring that flavor as their default Scheduler.
  * All instances are returned in a {@link Scheduler#init() initialized} state.
+ * <p>
+ * Please note, starting from 3.6.0
```
chemicL, line 67: Consider just replacing with `@since 3.6.0`.

reactor-core/src/main/java/reactor/core/scheduler/Schedulers.java
```diff
@@ -215,9 +217,21 @@ public static Scheduler fromExecutorService(ExecutorService executorService, Str
 	 * between callers. They can however be all {@link #shutdownNow() shut down} together, or replaced by a
 	 * {@link #setFactory(Factory) change in Factory}.
 	 *
+	 * <p>
```
chemicL, line 220: IMO the entire javadoc of this method needs to be rearranged. Adding these notes is confusing after having read the above specification. Please mention at the beginning that there are two implementations for the boundedElastic Scheduler - one for reusing a pool of platform threads with idle eviction and another for virtual threads that spawns a new VT for every submitted task while keeping each workers' tasks pipelined to avoid parallel execution. - And that the latter implementation requires JDK21 and a specific property to be set.

reactor-core/src/main/java/reactor/core/scheduler/Schedulers.java
```diff
@@ -215,9 +217,21 @@ public static Scheduler fromExecutorService(ExecutorService executorService, Str
...
+	 * {@link #DEFAULT_BOUNDED_ELASTIC_ON_VIRTUAL_THREADS} set to {@code true} the
+	 * returned scheduler will run on {@link VirtualThread}. Consider increasing
+	 * {@link #DEFAULT_BOUNDED_ELASTIC_SIZE} to run more concurrent
+	 * {@link VirtualThread} underneath.
+	 *
+	 *
 	 * @return the common <em>boundedElastic</em> instance, a {@link Scheduler} that dynamically creates workers with
 	 * an upper bound to the number of backing threads and after that on the number of enqueued tasks, that reuses
-	 * threads and evict idle ones
+	 * threads and evict idle ones. If runs on Java 21 with
```
chemicL, line 230: The `@return` section needn't be so detailed. It can just mention that a shared global boundedElastic instance is created and returned instead of repeating notes from the above.

reactor-core/src/main/java/reactor/core/scheduler/Schedulers.java
```diff
@@ -392,6 +406,12 @@ public static Scheduler newBoundedElastic(int threadCap, int queuedTaskCap, Stri
 	 * will prevent the JVM from exiting until their worker has been disposed AND they've been evicted by TTL,
 	 * or the whole scheduler has been {@link Scheduler#dispose() disposed}.
 	 *
+	 * <p>
```
chemicL, line 409: This comment needs to be also applied to the overloads (there are 3 above).

chemicL (changes requested):

reactor-core/src/main/java/reactor/core/scheduler/Schedulers.java
```diff
@@ -187,26 +191,69 @@ public static Scheduler fromExecutorService(ExecutorService executorService, Str
...
+	 * <li> Since 3.6.0 there is an thread-per-task implementation tailored to run on
+	 * {@link VirtualThread} instances. This implementation is enabled if the
+	 * application runs within Java 21 runtime and the
+	 * {@link #DEFAULT_BOUNDED_ELASTIC_ON_VIRTUAL_THREADS} system property is set to
+	 * {@code true}. Every Worker is based on the custom implementation of the execution
+	 * mechanism which ensures every submitted task runs on a new instance of the
+	 * {@link VirtualThread}. This implementation has a shared instance of
+	 * {@link ScheduledExecutorService} used to schedule delayed and periodic tasks
+	 * which upon the trigger is being offloaded to dedicated new instance of
+	 * {@link VirtualThread}.
```
rstoyanchev, line 217:
```suggestion
	 * <li> As of 3.6.0 there is a thread-per-task implementation tailored for use
	 * with virtual threads. This implementation is enabled if the
	 * application runs on a JDK 21+ runtime and the system property
	 * {@link #DEFAULT_BOUNDED_ELASTIC_ON_VIRTUAL_THREADS} is set to
	 * {@code true}. Every Worker is based on the custom implementation of the execution
	 * mechanism which ensures every submitted task runs on a new
	 * {@link VirtualThread} instance. This implementation has a shared instance of
	 * {@link ScheduledExecutorService} used to schedule delayed and periodic tasks
	 * such that when triggered they are offloaded to a dedicated new
	 * {@link VirtualThread} instance.
```

reactor-core/src/main/java/reactor/core/scheduler/VirtualThreadFactory.java
```diff
@@ -17,16 +17,14 @@
...
 
 import reactor.util.annotation.NonNull;
 import reactor.util.annotation.Nullable;
 
 /**
  * The noop {@link VirtualThread} Reactor {@link ThreadFactory} to be
- * used with {@link ThreadPerTaskBoundedElasticScheduler}. It throws exceptions when is
+ * used with {@link BoundedElasticPerThreadScheduler}. It throws exceptions when is
  * being created, so it indicates that current Java Runtime does not support
  * {@link VirtualThread}s.
```
rstoyanchev, line 29: It may be worth saying something like "This VirtualThreadFactory variant is included when Reactor is used with JDK versions lower than 21, and all methods raise an UnsupportedOperationException. An alternative variant is available for use on JDK 21+ where virtual threads are supported".

reactor-core/src/main/java/reactor/core/scheduler/BoundedElasticPerThreadScheduler.java
```diff
@@ -22,10 +22,10 @@
 import reactor.core.Disposable;
 import reactor.core.Scannable;
 
-final class ThreadPerTaskBoundedElasticScheduler
-		implements Scheduler, SchedulerState.DisposeAwaiter<ThreadPerTaskBoundedElasticScheduler.BoundedServices>, Scannable {
+final class BoundedElasticPerThreadScheduler
+		implements Scheduler, SchedulerState.DisposeAwaiter<BoundedElasticPerThreadScheduler.BoundedServices>, Scannable {
 
-	ThreadPerTaskBoundedElasticScheduler(int maxThreads, int maxTaskQueuedPerThread, ThreadFactory factory) {
+	BoundedElasticPerThreadScheduler(int maxThreads, int maxTaskQueuedPerThread, ThreadFactory factory) {
```
rstoyanchev, line 28: Similarly here, something like:

"This `BoundedElasticPerThreadScheduler` variant is included when Reactor is used with JDK versions lower than 21, and all methods raise an UnsupportedOperationException. An alternative variant is available for use on JDK 21+ where virtual threads are supported".

rstoyanchev (commented):
Looks good overall. Just some minor polishing suggestions for consideration.

reactor-core/src/main/java21/reactor/core/scheduler/BoundedElasticSchedulerSupplier.java
```diff
@@ -24,10 +24,10 @@
 import static reactor.core.scheduler.Schedulers.DEFAULT_BOUNDED_ELASTIC_SIZE;
 import static reactor.core.scheduler.Schedulers.LOOM_BOUNDED_ELASTIC;
 import static reactor.core.scheduler.Schedulers.newBoundedElastic;
-import static reactor.core.scheduler.Schedulers.newThreadPerTaskBoundedElastic;
+import static reactor.core.scheduler.Schedulers.factory;
 
 /**
- * JDK 8 Specific implementation of BoundedElasticScheduler supplier which uses
+ * JDK 21 Specific implementation of BoundedElasticScheduler supplier which uses
```
chemicL, line 30:
```suggestion
 * JDK 21+ Specific implementation of BoundedElasticScheduler supplier which uses
```

reactor-core/src/main/java/reactor/core/scheduler/Schedulers.java
```diff
@@ -1024,11 +1090,44 @@ default Scheduler newBoundedElastic(int threadCap, int queuedTaskCap, ThreadFact
...
+		 * @since 3.6.0
+		 *
+		 * @deprecated in favor of
+		 * {@link #newBoundedElasticThreadPerTask(int, int, ThreadFactory)}.
+		 * Should be safely removed in 3.8.0
+		 *
 		 * @return a new {@link Scheduler} that dynamically creates workers with an upper bound to
 		 * the number of backing threads
 		 */
 		default Scheduler newThreadPerTaskBoundedElastic(int threadCap, int queuedTaskCap,	ThreadFactory threadFactory) {
```
chemicL, line 1102: Now we have both threadPerTaskBoundedElastic and boundedElasticPerTask... Even though the latter is more favourable, with the fact that we already released threadPerTaskBoundedElastic, I think it's just better to leave one variant.

chemicL (changes requested):
With the release of 3.6.0 without this PR I suggest we keep the name that is already in the public API - threadPerTaskBoundedElastic, instead of having both variants.

docs/asciidoc/coreFeatures.adoc
```diff
@@ -209,13 +209,23 @@ dedicated thread, use `Schedulers.newSingle()` for each call.
...
+implementation is like its predecessor `elastic()` creates new worker pools as needed
+and reuses idle ones. Worker pools that stay idle for too long (the default is 60s) are
 also disposed. Unlike its `elastic()` predecessor, it has a cap on the number of backing threads it can create (default is number of CPU cores x 10).
 Up to 100 000 tasks submitted after the cap has been reached are enqueued and will be re-scheduled when a thread becomes available
-(when scheduling with a delay, the delay starts when the thread becomes available). This is a better choice for I/O blocking work.
-`Schedulers.boundedElastic()` is a handy way to give a blocking process its own thread so that
-it does not tie up other resources. See <<faq.wrap-blocking>>, but doesn't pressure the system too much with new threads.
+(when scheduling with a delay, the delay starts when the thread becomes available).
+ - Thread-per-task-based, designed to run on `VirtualThread` instances.
+To embrace that functionality, the application should run in Java 21 environment and set the `reactor.schedulers.defaultBoundedElasticOnVirtualThreads` system property to `true`.
```
chemicL, line 223:
```suggestion
To embrace that functionality, the application should run in Java 21+ environment and set the `reactor.schedulers.defaultBoundedElasticOnVirtualThreads` system property to `true`.
```

reactor-core/src/main/java/reactor/core/scheduler/Schedulers.java
```diff
@@ -63,6 +63,10 @@
  * Factories prefixed with {@code new} (eg. {@link #newBoundedElastic(int, int, String)} return a new instance of their flavor of {@link Scheduler},
  * while other factories like {@link #boundedElastic()} return a shared instance - which is the one used by operators requiring that flavor as their default Scheduler.
  * All instances are returned in a {@link Scheduler#init() initialized} state.
+ * <p>
+ * Since 3.6.0 {@link #boundedElastic()} can run tasks on {@link VirtualThread}s if the application
+ * runs on a Java 21 runtime and the {@link #DEFAULT_BOUNDED_ELASTIC_ON_VIRTUAL_THREADS}
```
chemicL, line 68: Please revisit all the places where we mention JDK 21 and consider if 21+ is more appropriate

chemicL (approved):

PR #3641, opened 2023-11-17
SinkManyUnicast discard support during subscription cancel

chemicL:
When subscription to `SinkManyUnicast` is cancelled at the time of being delivered, the discarding fails, as the actual `Subscriber` is not yet assigned. This change reorders the operations so that the actual is stored properly before delivering the `Subscription`.

### DIFF
--- a/reactor-core/src/main/java/reactor/core/publisher/SinkManyUnicast.java
+++ b/reactor-core/src/main/java/reactor/core/publisher/SinkManyUnicast.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2016-2022 VMware Inc. or its affiliates, All Rights Reserved.
+ * Copyright (c) 2016-2023 VMware Inc. or its affiliates, All Rights Reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -411,8 +411,8 @@ final class SinkManyUnicast<T> extends Flux<T> implements InternalManySink<T>, D
 		if (once == 0 && ONCE.compareAndSet(this, 0, 1)) {
 
 			this.hasDownstream = true;
-			actual.onSubscribe(this);
 			this.actual = actual;
+			actual.onSubscribe(this);
 			if (cancelled) {
 				this.hasDownstream = false;
 			} else {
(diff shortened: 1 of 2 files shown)


chemicL:
Analysing the test failures.

OlegDokuka (changes requested):
This may cause issues of concurrent onNext with onSubscribe or even onNext hapenning before onSubscribe

reactor-core/src/test/java/reactor/test/publisher/BaseOperatorTest.java
```diff
@@ -595,20 +595,14 @@ final void inputFusedAsyncOutputFusedAsyncCancel(OperatorScenario<I, PI, O, PO>
...
 					            qs.requestFusion(ASYNC);
-					            //UnicastProcessor#actual
-					            if (up.scan(Attr.ACTUAL) != qs || scenario.prefetch() == -1) {
-						            qs.size(); //touch undeterministic
-					            }
-					            else {
-						            assertThat(qs.size()).isEqualTo(up.scan(Attr.BUFFERED)); //UnicastProcessor#size
-					            }
 					            try {
+									// Assuming scenario.producing >= 3
```
chemicL, line 605: This seems to have been a mistake as on line 598 the demand was 0. So it was actually illegal to poll, which causes onNext signals to be triggered. In cases such as filtering out all the values, this would trigger upstream demand to continuously increase forcing termination. Therefore, the initial demand is set to scenario.producing and this code assumes it's not less than 3 to meet the three polls that happen here.

chemicL:
> This may cause issues of concurrent onNext with onSubscribe or even onNext hapenning before onSubscribe

Thanks, I updated the code to address this as discussed and added a few notes in this PR to clarify the changes.

OlegDokuka (approved):

reactor-core/src/test/java/reactor/test/publisher/BaseOperatorTest.java
chemicL: Actually, as discussed with @OlegDokuka `poll`-ing expresses demand on its own, so can be an optimization technique in these fused scenarios and not really a misuse. The issue observed was with not protecting against simultaneous `onNext` that can happen as a result of calling `poll` during `onSubscribe`. Now with proper protection in place, the problem does not occur.

PR #3660, opened 2023-12-11
Fix pendingTask count not tracking task rejection

### LINKED ISSUE #3642: PendingTasks leak in TimedScheduler
In `TimedScheduler`, micrometer's counter `pendingTasks` is never stop if the underlying scheduler reject the task.

## Expected Behavior

Micrometer's counter `pendingTasks` must be stopped if the underlying scheduler throws a `RejectedExecutionException` 

## Actual Behavior

Micrometer's counter `pendingTasks` is not be stopped when the underlying scheduler throws a `RejectedExecutionException`, resulting in never ending LongRunningTask 

## Steps to Reproduce

```java
    @Test
    void test() {
        CountDownLatch cdl = new CountDownLatch(1);
        SimpleMeterRegistry smr = new SimpleMeterRegistry();
        ExecutorService executorService =
            new ThreadPoolExecutor(1, 1, 0L, TimeUnit.MILLISECONDS,
                new SynchronousQueue<>()
            );

        Scheduler originalScheduler = Schedulers.fromExecutorService(executorService); // 1 thread - SynchronousQueue
        Scheduler timedScheduler = Micrometer.timedScheduler(originalScheduler, smr, "timed_scheduler");
        var meterIdStr = "timed_scheduler.scheduler.tasks.pending";
        RequiredSearch requiredSearch = smr.get(meterIdStr);
        LongTaskTimer longTaskTimer = requiredSearch.longTaskTimer();

        Runnable supp = () -> {
            try {
                cdl.await();
            }
        };

        // 1 - Runnable is successfully submitted, but will be blocked by CDL
        Assertions.assertDoesNotThrow(() -> timedScheduler.schedule(supp));

        // 2nd is rejected because scheduler has only 1 thread and is currently busy with exec 1
        Assertions.assertThrows(RejectedExecutionException.class, () -> timedScheduler.schedule(supp));
        cdl.countDown(); // release 1st

        assertEquals(0, longTaskTimer.activeTasks(), "Expected no active task since 2nd submission has been rejected");
    }
```

## Possible Solution
Catch any RejectedExecutionException to stop any pending tasks

## Your Environment

* Reactor version(s) used: 
  * reactor-core 3.6.0 
  * reactor core-micrometer 1.1.0 
* JVM version (`java -version`): 17.0.6
* Micrometer: 1.11.4

Nicolas125841:
Fixes issue with `TimedScheduler` where tasks rejected by underlying scheduler would not be removed from `pendingTasks` count. Idea is to catch any `RejectedExecutionException` thrown when scheduling the task, stop its pending sample, and rethrow the error so caller knows that the task was rejected.

Fixes #3642

### DIFF
--- a/reactor-core-micrometer/src/main/java/reactor/core/observability/micrometer/TimedScheduler.java
+++ b/reactor-core-micrometer/src/main/java/reactor/core/observability/micrometer/TimedScheduler.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2022 VMware Inc. or its affiliates, All Rights Reserved.
+ * Copyright (c) 2022-2023 VMware Inc. or its affiliates, All Rights Reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -16,6 +16,7 @@
 
 package reactor.core.observability.micrometer;
 
+import java.util.concurrent.RejectedExecutionException;
 import java.util.concurrent.TimeUnit;
 
 import io.micrometer.core.instrument.Counter;
@@ -87,13 +88,30 @@ final class TimedScheduler implements Scheduler {
 	@Override
 	public Disposable schedule(Runnable task) {
 		this.submittedDirect.increment();
-		return delegate.schedule(wrap(task));
+		task = wrap(task);
+
+		try {
+			return delegate.schedule(task);
+		}
+		catch (RejectedExecutionException exception) {
+			((TimedRunnable) task).pendingSample.stop();
+
+			throw exception;
+		}
 	}
 
 	@Override
 	public Disposable schedule(Runnable task, long delay, TimeUnit unit) {
 		this.submittedDelayed.increment();
-		return delegate.schedule(wrap(task), delay, unit);
+		task = wrap(task);
+
+		try {
+			return delegate.schedule(task, delay, unit);
+		}
+		catch (RejectedExecutionException exception) {
+			((TimedRunnable) task).pendingSample.stop();
+			throw exception;
+		}
 	}
 
 	@Override
@@ -150,13 +168,30 @@ final class TimedScheduler implements Scheduler {
 		@Override
 		public Disposable schedule(Runnable task) {
 			parent.submittedDirect.increment();
-			return delegate.schedule(parent.wrap(task));
+			task = parent.wrap(task);
+
+			try {
+				return delegate.schedule(task);
+			}
+			catch (RejectedExecutionException exception) {
+				((TimedRunnable) task).pendingSample.stop();
+
+				throw exception;
+			}
 		}
 
 		@Override
 		public Disposable schedule(Runnable task, long delay, TimeUnit unit) {
 			parent.submittedDelayed.increment();
-			return delegate.schedule(parent.wrap(task), delay, unit);
+			task = parent.wrap(task);
+
+			try {
+				return delegate.schedule(task, delay, unit);
+			}
+			catch (RejectedExecutionException exception) {
+				((TimedRunnable) task).pendingSample.stop();
+				throw exception;
+			}
 		}
 
 		@Override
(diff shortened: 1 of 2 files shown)


reactor-core-micrometer/src/main/java/reactor/core/observability/micrometer/TimedScheduler.java
```diff
@@ -87,13 +88,30 @@ Runnable wrapPeriodic(Runnable task) {
...
 	public Disposable schedule(Runnable task) {
 		this.submittedDirect.increment();
-		return delegate.schedule(wrap(task));
+		task = wrap(task);
+
+		try {
+			return delegate.schedule(task);
+		}
+		catch (RejectedExecutionException exception) {
+			((TimedRunnable) task).pendingSample.stop();
```
chemicL, line 97: Consider changing the signature of `wrap` and `wrapPeriodic` to return `TimedRunnable`, create a variable of that type instead of assigning to `task` and then you can avoid the cast here. I think it would be more elegant. This is justified in my opinion, as the `TimedScheduler` is a final class and the mentioned methods are just used here in the implementations and inside `TimedWorker`.

reactor-core-micrometer/src/test/java/reactor/core/observability/micrometer/TimedSchedulerTest.java
```diff
@@ -324,4 +329,65 @@ void workerSchedulePeriodicallyIsCorrectlyMetered() throws InterruptedException
 				+ testScheduler.submittedPeriodicInitial.count()
 				+ testScheduler.submittedPeriodicIteration.count(), "completed tasks == sum of all timer counts");
 	}
+
+	@Test
+	void pendingScheduleRemovedOnScheduleRejection() {
+		CountDownLatch cdl = new CountDownLatch(1);
```
chemicL, line 335: Instead of using a latch, consider creating a disposed `Scheduler` (e.g. `Schedulers.newSingle("testScheduler"`).

reactor-core-micrometer/src/test/java/reactor/core/observability/micrometer/TimedSchedulerTest.java
```diff
@@ -324,4 +329,65 @@ void workerSchedulePeriodicallyIsCorrectlyMetered() throws InterruptedException
...
+
+		assertThatNoException().isThrownBy(() -> testScheduler.schedule(supp));
+		assertThatExceptionOfType(RejectedExecutionException.class).isThrownBy(() -> testScheduler.schedule(supp));
+		assertThatExceptionOfType(RejectedExecutionException.class)
+				.isThrownBy(() -> testScheduler.schedule(supp, 0, TimeUnit.SECONDS));
+
+		cdl.countDown();
+
+		assertThat(longTaskTimer.activeTasks())
+				.as("longTaskTimer.activeTasks()")
```
chemicL, line 359: It seems you validate it gets zeroed, however the precondition should be to first validate a condition under which the timer is non-zero.

reactor-core-micrometer/src/test/java/reactor/core/observability/micrometer/TimedSchedulerTest.java
chemicL: As noticed in the comment below, feel free to ignore this suggestion once a check is added for the metric being non-zero.

chemicL (changes requested):
Thanks for the contribution! A few easy-to-incorporate suggestions from my side.

Nicolas125841:
Thank you for the feedback! I committed your suggestions.

chemicL (approved):

chemicL:
@Nicolas125841 thank you for the contribution, great job! 🚀

chemicL:
@Nicolas125841 just FYI the tests turned out to be flaky. Please have a look at my attempt at improving the predictability of the outcomes in #3669

Nicolas125841:
@chemicL Thanks for the update, yours looks much more stable. I will keep this in mind!

chemicL:
👍 
BTW: what helped me to reproduce the issues when debugging is the [repeat functionality in IDEA](https://www.jetbrains.com/help/idea/run-debug-configuration-junit.html#tests), which allowed me to also notice the need for shutting down the executors at the end, which unreleased threw OOM errors.

Looking forward to more contributions from you, @Nicolas125841 🚀

PR #3673, opened 2023-12-20
Restore where onLastOperatorHook is applied

chemicL:
Recent improvements in the automatic context propagation (#3549) resulted in a regression – some sites where the "last operator" hook was previously applied no longer saw that behaviour. This change restores it.

Implementation wise, it's worth noting that the "last operator" functionality relies on executing the subscribe(Subscriber) method from the base reactive-streams Publisher instead of the overloads that come from CorePublisher. The implementations of the reactive-streams base method in reactor-core apply this hook and that is when something is considered a "last operator". The wrapping of the Publisher when a non-internal producer is encountered to restore ThreadLocal values has changed the compiler's inference of the signature to use the CoreSubscriber argument variant, breaking the behaviour.

This commit does not bring any tests as the functionality was not extensively tested before. The issue was discovered in spring-security in https://github.com/spring-projects/spring-security/issues/14207 and the change has been validated against the actual use case.

### DIFF
--- a/reactor-core/src/main/java/reactor/core/publisher/FluxBufferWhen.java
+++ b/reactor-core/src/main/java/reactor/core/publisher/FluxBufferWhen.java
@@ -360,7 +360,8 @@ final class FluxBufferWhen<T, OPEN, CLOSE, BUFFER extends Collection<? super T>>
 
 			BufferWhenCloseSubscriber<T, BUFFER> bc = new BufferWhenCloseSubscriber<>(this, idx);
 			subscribers.add(bc);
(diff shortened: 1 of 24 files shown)


chemicL:
For the record, the execution sites are briefly tested in https://github.com/reactor/reactor-core/blob/main/reactor-core/src/test/java/reactor/core/publisher/HooksTest.java#L716-L921 but do not cover all the existing operators. As this functionality should most probably be deprecated in favour of the new context propagation mechanisms (default/limited vs auto), I didn't add extensive tests for this at this moment.

OlegDokuka (approved):

PR #3696, opened 2024-01-16
Flux/Mono.usingWhen late resource memory leak (#3695)

### LINKED ISSUE #3695: Flux/Mono.usingWhen late resource memory leak
Depending on the resource acquisition mechanism usingWhen method may lead to a memory or resourse leak. Latest related fixes  asured that provided resource will be closed or subscription will be cancelled. The problem is that cancellation process most often is not deterministic and either not intuitive or impossible to implement (side effects during resource creation). Yes, sometimes cancelling the resource subscription may save some resources if we cancel it but in most cases it leads to resource leaks that are very hard to find and fix (the amount of existing issues for this method is a good proof of this point).

This issue is a result of investigation of previously created issues (#1233, #2661, #124, #2836) and should meet all the requirements.

### Expected Behavior
asyncCancel (respective asyncCleanup) should be always called.

### Actual Behavior
Resource creation process was started but pipeline was cancelled so cleanup never happened.

### Steps to Reproduce
Ensure some delay in the resource allocation
Cancel the main pipeline before the resource is emitted.

Below is the modified version of an existing test that shows that problem still exists if resource allocation is long. Existing test passes because resource allocation is delayed (so can be cancelled) but not long.
Random is not needed since problem is 100% reproducible.

```java
	@Test
	void cancelEarlyDoesNotLeak() {
		Releaseable releaseable = new Releaseable();
		Mono<Long> mono = Mono.usingWhen(Mono.fromSupplier(() -> {
			LockSupport.parkNanos(Duration.ofMillis(150).toNanos());
			releaseable.allocate();
			return releaseable;
		}), it -> Mono.just(1L), it -> Mono.fromRunnable(releaseable::release));

		StepVerifier.create(mono, 1)
		            .thenAwait(Duration.ofMillis(50))
		            .thenCancel()
		            .verify();

		Awaitility.await()
		          .atMost(Duration.ofSeconds(10))
		          .until(releaseable::wasAcquired);
		Awaitility.await()
		          .atMost(Duration.ofSeconds(10))
		          .until(releaseable::wasReleased);
	}

	static class Releaseable {

		private final AtomicBoolean allocated = new AtomicBoolean();
		private final AtomicBoolean released = new AtomicBoolean();

		void allocate() {
			System.out.println("alloc");
			allocated.set(true);
		}

		void release() {
			System.out.println("release");
			released.set(true);
		}

		boolean wasReleased() {
			return released.get();
		}

		boolean wasAcquired() {
			return released.get();
		}
	}
```

### Possible Solution
Remove this cancel method override from ResourceSubscriber that forces early cancel for resource pipeline. That will allow the process to gracefully finish, emit the resource and do a cleanup.
```java
		@Override
		public void cancel() {
			if (!resourceProvided) {
				resourceSubscription.cancel();
			}

			super.cancel();
		}
```

### Your Environment
Reactor version(s) used: 3.6.3
Other relevant libraries versions (eg. netty, ...): -
JVM version (java -version): Java 17
OS and version (eg uname -a): Windows 11

mlozbin:
UsingWhen method may lead to a memory or resource leak of allocation was finished after the main pipeline was cancelled. This change assures that asyncCancel is always called even if deferred.

Fixes #3695.

### DIFF
--- a/reactor-core/src/main/java/reactor/core/publisher/FluxUsingWhen.java
+++ b/reactor-core/src/main/java/reactor/core/publisher/FluxUsingWhen.java
@@ -25,7 +25,6 @@ import java.util.function.Function;
 import org.reactivestreams.Publisher;
 import org.reactivestreams.Subscriber;
 import org.reactivestreams.Subscription;
-
 import reactor.core.CoreSubscriber;
 import reactor.core.Exceptions;
 import reactor.core.Fuseable.ConditionalSubscriber;
@@ -240,15 +239,6 @@ final class FluxUsingWhen<T, S> extends Flux<T> implements SourceProducer<T> {
 				actual.onSubscribe(this);
 				s.request(Long.MAX_VALUE);
 			}
-		}
-
-		@Override
-		public void cancel() {
-			if (!resourceProvided) {
-				resourceSubscription.cancel();
-			}
-
-			super.cancel();
 		}
 
 		@Override
--- a/reactor-core/src/main/java/reactor/core/publisher/MonoUsingWhen.java
+++ b/reactor-core/src/main/java/reactor/core/publisher/MonoUsingWhen.java
@@ -24,7 +24,6 @@ import java.util.function.Function;
 import org.reactivestreams.Publisher;
 import org.reactivestreams.Subscriber;
 import org.reactivestreams.Subscription;
-
 import reactor.core.CoreSubscriber;
 import reactor.core.Exceptions;
 import reactor.core.publisher.FluxUsingWhen.UsingWhenSubscriber;
@@ -220,15 +219,6 @@ final class MonoUsingWhen<T, S> extends Mono<T> implements SourceProducer<T> {
 				actual.onSubscribe(this);
 				s.request(Long.MAX_VALUE);
 			}
-		}
-
-		@Override
-		public void cancel() {
-			if (!resourceProvided) {
-				resourceSubscription.cancel();
-			}
-
-			super.cancel();
 		}
 
 		@Override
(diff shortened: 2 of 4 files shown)


chemicL:
Thank you @mlozbin for the effort. Unfortunately, as highlighted [in my comment in the original issue](https://github.com/reactor/reactor-core/issues/3695#issuecomment-1954224578), this PR breaks existing functionality and can't be accepted in the current form. Let's work through the design first and decide if a change in the code is necessary.

chemicL:
I reopened the PR hoping to run the CI and prove a point raised in the original issue, but there are two problems:

* The PR fails the preliminary tests due to not having the license headers updated (`./gradlew spotlessApply` fixes that)
* There are failing tests and removed tests which should remain in the codebase.

Most importantly, this PR removes tests that validate functionality that currently is supported. Perhaps not in a documented way, but this behaviour can be something users out there rely on. So this change can't go in a patch release and definitely not without a documented migration path. Luckily, while the removal happened in `MonoUsingWhenTest` class, they are still present in `FluxUsingWhenTest` class and fail with this PR: 

* `FluxUsingWhenTest > lateFluxResourcePublisherIsCancelledOnCancel()`
* `FluxUsingWhenTest > lateMonoResourcePublisherIsCancelledOnCancel()`

In summary, in the current form, I'm discarding this PR, but I remain open to discussing the plan for both solving the issue you describe while giving the existing usage patterns either a migration path or keeping the current behaviour supported (my preferred option). Thank you @mlozbin for the work you did so far and I look forward to constructive discussion in the original issue.

PR #3704, opened 2024-01-25
Add `AutoCloseable` shourtcut on `Flux#using`, `Mono#using`

### LINKED ISSUE #3333: AutoCloseable and using shortcuts
## Motivation
I noticed that most times I was using Flux.using and Mono.using that I was using instances of AutoCloseable. So I added the following methods so I didn't have to pass in a closer each time.

Would it be possible to add the desired solution into the core. If so I can create a pull request.

## Desired solution

```
Flux.using(
  ()-> Files.newInputStream(),
  (in)-> ...
);
```

 ```java
class Flux {
   :
// This could go in a shared class. Cached closer so it doesn't need to be created on each call
 private static Consumer<AutoCloseable> AUTO_CLOSE = (resource) -> {
    try {
      resource.close();
    } catch (final Exception e) {
    }
  }; 

  public static <R extends AutoCloseable, V> Flux<V> using(final Callable<R> resourceSupplier,
    final Function<R, Publisher<V>> action) {
    return Flux.using(resourceSupplier, action, AUTO_CLOSE);
  }
}
class Mono {
   :
 private static Consumer<AutoCloseable> AUTO_CLOSE = (resource) -> {
    try {
      resource.close();
    } catch (final Exception e) {
    }
  };

  public static <R extends AutoCloseable, V> Mono<V> using(final Callable<R> resourceSupplier,
    final Function<R, Mono<V>> action) {
    return Mono.using(resourceSupplier, action, AUTO_CLOSE);
  }
}
```

### PR DESCRIPTION
**Pull request by injae-kim**
Make `Flux, Mono#using` to close `AutoCloseable` resource,
so that user don't have to pass resourceCleanup consumer.

Fixes #3333.

### CODE REVIEW

File: reactor-core/src/main/java/reactor/core/publisher/Flux.java [hunk 1]
```diff
@@ -2100,6 +2101,32 @@ public static <T, D> Flux<T> using(Callable<? extends D> resourceSupplier, Funct
...
+	 * @param resourceSupplier a {@link Callable} that is called on subscribe to generate the resource
+	 * @param sourceSupplier a factory to derive a {@link Publisher} from the supplied resource
+	 * @param <T> emitted type
+	 *
+	 * @return a new {@link Flux} built around a disposable resource
+	 * @see #usingWhen(Publisher, Function, Function, BiFunction, Function)
+	 * @see #usingWhen(Publisher, Function, Function)
+	 */
+	public static <T> Flux<T> using(Callable<? extends AutoCloseable> resourceSupplier,
+									Function<? super AutoCloseable, ? extends Publisher<? extends T>> sourceSupplier) {
```
Sage-Pierce, line 2126: This feature caught my eye. For better ease of use, you probably want this signature to be:
```java
public static <T, D extends AutoCloseable> using(Callable<D> resourceSupplier, Function<? super D, ? extends Publisher<? extends T>> sourceSupplier)
```

He-Pin, line 2126: Need format.

injae-kim: ✅ updated, good suggestion!

injae-kim: ✅ updated

reactor-core/src/main/java/reactor/core/publisher/Flux.java
```diff
@@ -2100,6 +2101,33 @@ public static <T, D> Flux<T> using(Callable<? extends D> resourceSupplier, Funct
...
+	 * @param resourceSupplier a {@link Callable} that is called on subscribe to generate the resource
+	 * @param sourceSupplier a factory to derive a {@link Publisher} from the supplied resource
+	 * @param <T> emitted type
+	 * @param <D> resource type
+	 *
+	 * @return a new {@link Flux} built around a disposable resource
+	 * @see #usingWhen(Publisher, Function, Function, BiFunction, Function)
+	 * @see #usingWhen(Publisher, Function, Function)
+	 */
+	public static <T, D extends AutoCloseable> Flux<T> using(Callable<D> resourceSupplier,
```
chemicL, line 2126:
```suggestion
	public static <T, D extends AutoCloseable> Flux<T> using(Callable<? extends D> resourceSupplier,
```

reactor-core/src/main/java/reactor/core/publisher/Mono.java
```diff
@@ -912,6 +913,28 @@ public static <T, D> Mono<T> using(Callable<? extends D> resourceSupplier,
...
+	 * <img class="marble" src="doc-files/marbles/usingForMono.svg" alt="">
+	 *
+	 * @param resourceSupplier a {@link Callable} that is called on subscribe to create the resource
+	 * @param sourceSupplier a {@link Mono} factory to create the Mono depending on the created resource
+	 * @param <T> emitted type
+	 * @param <D> resource type
+	 *
+	 * @return new {@link Mono}
+	 */
+	public static <T, D extends AutoCloseable> Mono<T> using(Callable<D> resourceSupplier,
```
chemicL, line 934:
```suggestion
	public static <T, D extends AutoCloseable> Mono<T> using(Callable<? extends D> resourceSupplier,
```

reactor-core/src/main/java/reactor/core/publisher/Mono.java
```diff
@@ -912,6 +913,28 @@ public static <T, D> Mono<T> using(Callable<? extends D> resourceSupplier,
...
+			Function<? super D, ? extends Mono<? extends T>> sourceSupplier) {
+		return using(resourceSupplier, sourceSupplier, Flux.AUTO_CLOSE, true);
```
chemicL, line 936: Instead of using `Flux.AUTO_CLOSE` here, perhaps it makes sense to move `AUTO_CLOSE` to the `reactor.core.Exceptions` class and use that from `Mono` and `Flux`.

reactor-core/src/main/java/reactor/core/publisher/Flux.java
chemicL, line 2126: I do feel we also need the counter-part that allows to delay the resource cleanup until after the sequence has terminated (using the `eager` argument).

reactor-core/src/test/java/reactor/core/publisher/FluxUsingTest.java
```diff
@@ -146,6 +146,24 @@ public void normalEager() {
...
 
+	@Test
+	public void normalAutoCloseable() {
+		AssertSubscriber<AutoCloseable> ts = AssertSubscriber.create();
+
+		AtomicInteger cleanup = new AtomicInteger();
+
+		AutoCloseable resource = cleanup::incrementAndGet;
+
+		Flux.using(() -> resource, r -> Flux.just(resource))
```
chemicL, line 157: As a matter of fact, this test can also check `doOnTerminate` condition just like `Mono` does. As a side quest, the existing `normal` and `normalEager` tests in `FluxUsingTest` could leverage this distinction as currently they are the same test basically.

reactor-core/src/main/java/reactor/core/publisher/Flux.java
```diff
@@ -2100,6 +2101,33 @@ public static <T, D> Flux<T> using(Callable<? extends D> resourceSupplier, Funct
 		return using(resourceSupplier, sourceSupplier, resourceCleanup, true);
 	}
 
+	/**
+	 * Uses a {@link AutoCloseable} resource, generated by a supplier for each individual Subscriber,
```
chemicL, line 2105:
```suggestion
	 * Uses an {@link AutoCloseable} resource, generated by a supplier for each individual Subscriber,
```

reactor-core/src/main/java/reactor/core/publisher/Mono.java
```diff
@@ -912,6 +913,28 @@ public static <T, D> Mono<T> using(Callable<? extends D> resourceSupplier,
 		return using(resourceSupplier, sourceSupplier, resourceCleanup, true);
 	}
 
+	/**
+	 * Uses a {@link AutoCloseable} resource, generated by a supplier for each individual Subscriber,
```
chemicL, line 917:
```suggestion
	 * Uses an {@link AutoCloseable} resource, generated by a supplier for each individual Subscriber,
```

chemicL (changes requested):
Thanks for the contribution 🎉 I left a few suggestions which will make the contribution complete in my view, hope you can apply them and we can ship this nice feature.

reactor-core/src/main/java/reactor/core/publisher/Mono.java
chemicL: I wonder whether `Exceptions.propagate(t)` would be a good candidate here. It feels like a natural choice.

reactor-core/src/main/java/reactor/core/publisher/Flux.java
```diff
@@ -11110,6 +11138,14 @@ static <I> Flux<I> wrap(Publisher<? extends I> source) {
 		return target;
 	}
 
+	static final Consumer<? super AutoCloseable> AUTO_CLOSE = resource -> {
+		try {
+			resource.close();
+		} catch (Exception e) {
+			throw new RuntimeException(e);
```
chemicL, line 11145: I wonder whether `Exceptions.propagate(t)` would be a good candidate here. It feels like a natural choice.

reactor-core/src/test/java/reactor/core/publisher/FluxUsingTest.java
```diff
@@ -146,6 +146,24 @@ public void normalEager() {
 		assertThat(cleanup).hasValue(1);
 	}
 
+	@Test
```
chemicL, line 149: I do understand we have tests for the regular cleanup, but it needs to be tested in case the implementations diverge at any point -> validating both eager and lazy cleanup behaviour in the scenario that the `AutoCloseable` throws.

reactor-core/src/test/java/reactor/core/publisher/FluxUsingTest.java
chemicL: For this you can consider parameterized methods not to duplicate the codebase in e.g. `sourceFactoryAndResourceCleanupThrow`

reactor-core/src/main/java/reactor/core/publisher/Mono.java
He-Pin: I thik it would be better with a  `R` than `D`.

injae-kim: https://github.com/reactor/reactor-core/blob/70f512672cfc2e0f5ba1ddce1f3ee25b3a9e7f9b/reactor-core/src/main/java/reactor/core/publisher/Flux.java#L2098

`<T, D> Flux<T> using` already uses `D` and I just copy it 😅
If we decide to use `R instead D`, we need to fix other methods..

reactor-core/src/main/java/reactor/core/publisher/Flux.java
injae-kim: ✅ updated, thanks!

reactor-core/src/main/java/reactor/core/publisher/Mono.java
injae-kim: ✅ updated, nice suggestion!

reactor-core/src/main/java/reactor/core/publisher/Flux.java
injae-kim: ✅ updated, add method with `eager` argument!

reactor-core/src/test/java/reactor/core/publisher/FluxUsingTest.java
injae-kim: ✅ updated, I use `ParameterizedTest` to de-dup test codes and add&enhance test cases.

I'm not sure it's what you want, so feel free to share your opnion~!

reactor-core/src/main/java/reactor/core/publisher/Flux.java
injae-kim: ✅ updated

reactor-core/src/main/java/reactor/core/publisher/Mono.java
injae-kim: ✅ updated

reactor-core/src/main/java/reactor/core/publisher/Flux.java
injae-kim: ✅ updated, `Exceptions.propagate(t)` is what I'm looking for!

reactor-core/src/main/java/reactor/core/publisher/Flux.java
```diff
@@ -2100,6 +2101,64 @@ public static <T, D> Flux<T> using(Callable<? extends D> resourceSupplier, Funct
 		return using(resourceSupplier, sourceSupplier, resourceCleanup, true);
 	}
 
+	/**
```
chemicL, line 2104: For alignment purposes, I'd suggest moving the two new methods below the existing `using` methods but before `usingWhen`.

reactor-core/src/test/java/reactor/core/publisher/MonoUsingTest.java
```diff
@@ -36,6 +41,74 @@
...
+	public static List<Mono<Integer>> sourcesThrowEager() {
+		return Arrays.asList(
+				Mono.using(() -> 1, r -> { throw new RuntimeException("forced failure"); }, cleanup::set),
+				Mono.using(() -> 1, r -> { throw new RuntimeException("forced failure"); }, cleanup::set, true),
+				Mono.using(() -> cleanup::incrementAndGet, r -> { throw new RuntimeException("forced failure"); }),
+				Mono.using(() -> cleanup::incrementAndGet, r -> { throw new RuntimeException("forced failure"); }, true)
+		);
+	}
+
+	private static final AtomicInteger cleanup = new AtomicInteger();
```
chemicL, line 105: Do you think we can avoid the static field? E.g. have the test cases contain the cleanup counter?

    static class CleanupCase {
        final Mono<Integer> resourceMono;
        final AtomicInteger cleanup;
        // ...
    }
    
And then have the test cases like

    public static List<CleanupCase> sourcesNonEager() {
        // ...
    }

I just want to prevent a situation where at some point we decide to use parallel tests and this would become a case to solve at that point.

reactor-core/src/test/java/reactor/core/publisher/MonoUsingTest.java
```diff
@@ -59,15 +132,12 @@ public void resourceCleanupNull() {
 		});
 	}
 
-	@Test
-	public void normal() {
+	@ParameterizedTestWithName
```
chemicL, line 135: Currently different test cases are indistinguishable when they fail. The suggestion for `CleanupCase` wrapper class can be used to add a `name` field and then implement a `toString` method that uses the `name`. This way we can see exactly which scenario failed. Please consult e.g. `ContextLossDetectionTest` to see how that works.

chemicL (changes requested):
Thanks for applying the changes! I added a few more suggestions. Also, I don't see a test case for when the `close` action fails and how the exception is handled/dropped, is it there somewhere? If not, I believe it would be worthwhile to add such validation.

reactor-core/src/main/java/reactor/core/publisher/Mono.java
chemicL: Agreed, I believe consistency is preferred.

chemicL:
And one more thing. Please notice the CI is failing due to license headers being non-compliant. Please run `./gradlew spotlessApply` before committing.

chemicL:
Please also update the `methodExcludes` section in `japicmp` entry in `reactor-core/build.gradle` file as this is a source incompatible change we need to account for. That's why the CI build fails.

reactor-core/src/main/java/reactor/core/publisher/Flux.java
injae-kim: ✅ updated!

reactor-core/src/test/java/reactor/core/publisher/MonoUsingTest.java
injae-kim: > we decide to use parallel tests and this would become a case to solve at that point.

oh it's really good point. ✅ updated by `CleanupCase`!

reactor-core/src/test/java/reactor/core/publisher/MonoUsingTest.java
injae-kim: image

✅ updated! override `toString`, now I can distinguish each test by `name` 👍

injae-kim: > Please consult e.g. ContextLossDetectionTest to see how that works.

Thanks for helpful guide! it saved my time a lot.

injae-kim:
> I don't see a test case for when the close action fails and how the exception is handled/dropped

✅ updated, I added `resourcesCleanupThrow` test cases~!

> Please also update the methodExcludes section in japicmp entry in reactor-core/build.gradle file

✅ updated! this is the reason why CI build fails. thanks!

chemicL (approved):
Excellent work! Thank you for the time and applying all the suggestions. It was a pleasure to review your work and hope to see more contributions in the future 🚀

He-Pin:
Thanks Again, I'm  using reactor core at $work too.

injae-kim:
Thank you to all reviewers! your code reviews are really great and kind, I've learn a lot from review.

I'm big fan of reactive-streams and spring-framework, so I'll keep contributing on reactor-core. 🚀

PR #3705, opened 2024-01-26
Fix return a value that will be nullalbe on mapNotNull

KSH-code:
https://github.com/reactor/reactor-kotlin-extensions/issues/52

The above issue explains `mapNotNull` does not change values into non-null values.

Thus, below code occurs an error such as `Operator call corresponds to a dot-qualified call 'it.times(3)' which is not allowed on a nullable receiver 'it'.` 
```
    Flux.just(1, 2, 3, null).mapNotNull {
        it
    }.map {
        println(it * 3)
    }
```

image

Therefore, `mapNotNull` method must return `@Nonnull` value only.

### DIFF
--- a/reactor-core/src/main/java/reactor/core/publisher/Flux.java
+++ b/reactor-core/src/main/java/reactor/core/publisher/Flux.java
@@ -70,6 +70,7 @@ import reactor.core.scheduler.Scheduler.Worker;
 import reactor.core.scheduler.Schedulers;
 import reactor.util.Logger;
 import reactor.util.Metrics;
+import reactor.util.annotation.NonNull;
 import reactor.util.annotation.Nullable;
 import reactor.util.concurrent.Queues;
 import reactor.util.context.Context;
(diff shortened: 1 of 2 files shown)


chemicL:
I'm not very knowledgeable in Kotlin, but the proposed change seems to be redundant. The [`reactor.core.publisher` package has](https://github.com/reactor/reactor-core/blob/main/reactor-core/src/main/java/reactor/core/publisher/package-info.java#L33) a [`NonNullApi` annotation](https://github.com/reactor/reactor-core/blob/main/reactor-core/src/main/java/reactor/util/annotation/NonNullApi.java#L28-L49) which should be equivalent to what this PR introduces but for all the method's return values. Can you please explain why this change would be required?

KSH-code:
@chemicL In my understanding, `@NonNullApi` is applied to only `Flux` or `Mono`. not for a generic class `T`.
And also I'm lack of understanding about generic types and annotations. I will check out those then I will fix this pr.

PR #3714, opened 2024-02-06
[issues/3340] bugfix concurrency issues FluxReplay.Size/SizeAndTime

MikkelHJuul:
This change adds a bunch of tests and improves concurrency of the Size(AndTime)BoundReplayBuffers (I believe they will be able to run in unsafe-mode after this change). The user will not see any changes to the API.

This PR fixes the current replay buffers' issue with holding onto one too many items, as the head is reused (it is final) and will only point at the current Atomic linked list's head. In stead of it being the volatile reference that is being reassigned to roll over the linked list (but always only replaying `head.get()`; which leaked `head.value`). 
This PR further fix an issue in the #add methods of both implementations that caused writes to be missed. The tail value was non-atomically read and written to, losing updates along the way.

See [#3340](https://github.com/reactor/reactor-core/issues/3340).

### DIFF
--- a/reactor-core/src/jcstress/java/reactor/core/publisher/SinksManyReplayLatestStressTest.java
+++ b/reactor-core/src/jcstress/java/reactor/core/publisher/SinksManyReplayLatestStressTest.java
@@ -1,0 +1,239 @@
+package reactor.core.publisher;
+
+import org.openjdk.jcstress.annotations.Actor;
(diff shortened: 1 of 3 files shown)


chemicL:
Closing as explained in https://github.com/reactor/reactor-core/issues/3340#issuecomment-2242821598.

PR #3716, opened 2024-02-06
Nightly build to run jcstress weekly, snapshots daily

chemicL:


.github/workflows/snapshots.yml
```diff
@@ -0,0 +1,64 @@
...
+        branch: [ 3.5.x, main ]
+    steps:
+      - uses: actions/checkout@8ade135a41bc03ea155e62e844d188df1ea18608 # tag=v4
+        with:
+          ref: ${{ matrix.branch }}
+      - uses: actions/setup-java@0ab4596768b603586c0de567f2430c30f5b0d2b0 # tag=v3
+        with:
+          distribution: 'temurin'
+          java-version: 8
+      - uses: gradle/gradle-build-action@842c587ad8aa4c68eeba24c396e15af4c2e9f30a # tag=v2
```
pderop, line 23:
```suggestion
      - uses: gradle/actions/setup-gradle@v3
```

While testing the workflow, I saw that gradle-build-action@v2 produces this following deprecation warning:

```

 
--
Node.js 16 actions are deprecated. Please update the following actions to use Node.js 20: gradle/gradle-build-action@842c587ad8aa4c68eeba24c396e15af4c2e9f30a. For more information see: https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/.
```

according to https://github.com/gradle/gradle-build-action, you can use `gradle/actions/setup-gradle@v3` as a replacement. (you might prefer to use commit id instead of @v3).

pderop (commented):
In addition to the two comments, what do you think about trying to use a matrix in order to reduce the size of the working, something like this:

```
name: Snapshots Check

on:
  schedule:
    - cron: "0 20 * * *"
permissions: read-all
jobs:
  run-tests:
    runs-on: ubuntu-latest
    strategy:
      fail-fast: false
      matrix:
        branch: [ 3.5.x, main ]
        test-type:
          - type: fast
            arguments: ":reactor-core:test --no-daemon -Pjunit-tags=!slow -DuseSnapshotMicrometerVersion=true"
          - type: slow
            arguments: ":reactor-core:test --no-daemon -Pjunit-tags=slow -DuseSnapshotMicrometerVersion=true"
          - type: other
            arguments: "check -x :reactor-core:test -x spotlessCheck -x :reactor-core:jcstress --no-daemon -DuseSnapshotMicrometerVersion=true -Pjcstress.mode=default"
    name: Test on ${{ matrix.branch }} - ${{ matrix.test-type.type }} tests
    steps:
      - name: Checkout Repository
        uses: actions/checkout@8ade135a41bc03ea155e62e844d188df1ea18608 # tag=v4
        with:
          ref: ${{ matrix.branch }}

      - name: Setup Java
        uses: actions/setup-java@v4
        with:
          distribution: 'temurin'
          java-version: 8

      - name: Run Gradle Tests
        uses: gradle/actions/setup-gradle@v3
        with:
          arguments: ${{ matrix.test-type.arguments }}
```

violetagg:
IMO from security point of view it is better to pin the GitHub actions https://blog.rafaelgss.dev/why-you-should-pin-actions-by-commit-hash

chemicL:
Thanks @pderop and @violetagg :)

PR #3678, opened 2024-01-03
Improve RaceTestUtils to yield when CPUs don't suffice

chemicL:
When `RaceTestUtils`-based concurrency tests involving more tasks than available CPUs were run, they'd time-out when run in the Java 21 setup with loom-based boundedElastic `Scheduler`. This change should eliminate such situations.

Addresses related issues below and potentially more.
Fixes #3629
Fixes #3628

PR #3721, opened 2024-02-08
Restore java21Test back into workflows

chemicL:
As a follow-up after temporarily disabling java21 reactor-core tests and then fixing the `RaceTestUtils` to accommodate virtual threads limitations with regards to number of CPUs and busy spinning, the java21 tests can be restored.

Related #3590
Follow-up to #3678


violetagg (approved):

chemicL:
@violetagg thank you!

PR #3725, opened 2024-02-12
Fix auto-cancelled Sinks.many() accepting emissions

### LINKED ISSUE #3715: Auto-cancelled Sink still accepts emissions
Sinks created via `Sinks.many().multicast().onBackpressureBuffer()` will keep accepting emissions even after they're cancelled by `autoCancel` (which is implicitly enabled), until the internal buffer is filled.

## Expected Behavior
After the sink is cancelled, `sink.tryEmitNext(...)` should fail.

## Actual Behavior
`sink.tryEmitNext(...)` succeeds and keeps filling the buffer.

## Steps to Reproduce
```java
@Test
public void cancelledSinkStillAcceptsEmissions() {
    Sinks.Many<String> sink = Sinks.many().multicast().onBackpressureBuffer();
    Disposable subscription1 = sink.asFlux().subscribe(s -> System.out.println("1: " + s));
    assertEquals(1, sink.currentSubscriberCount());
    sink.tryEmitNext("Test1");
    subscription1.dispose();
    assertEquals(0, sink.currentSubscriberCount()); //autoCancel kicked in
    Disposable subscription2 = sink.asFlux().subscribe(s -> System.out.println("2: " + s));
    assertTrue(subscription2.isDisposed()); //auto-cancelled
    assertEquals(0, sink.currentSubscriberCount());
    assertTrue(sink.tryEmitNext("Test2").isFailure(), "Emissions on a cancelled sink should fail");
}
```

## Your Environment

* JVM version (`java -version`): 17, 21

kaqqao:
Fixes #3715 

I don't know if this is indeed the correct thing to do, but it does help demonstrate the situation and a possible solution.

### DIFF
--- a/reactor-core/src/test/java/reactor/core/publisher/SinksTest.java
+++ b/reactor-core/src/test/java/reactor/core/publisher/SinksTest.java
@@ -24,6 +24,7 @@ import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
+import java.util.function.Consumer;
 import java.util.function.Supplier;
 import java.util.stream.Stream;
 
@@ -34,6 +35,7 @@ import org.junit.jupiter.api.DynamicTest;
 import org.junit.jupiter.api.Nested;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.TestFactory;
+import reactor.core.Disposable;
 import reactor.core.scheduler.Schedulers;
 import reactor.test.StepVerifier;
 import reactor.test.StepVerifierOptions;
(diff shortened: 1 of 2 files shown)


reactor-core/src/test/java/reactor/core/publisher/SinksTest.java
chemicL, line 1: Please also run `spotlessApply` task for this file.

chemicL:
Thanks so much for the effort so far! 🎉 

I think the proposed implementation which checks for cancellation upon emission attempts looks simple and elegant. Unfortunately, I'm not sure it's correct. It would definitely improve the situation. However, I am worried about potential races. As I understand this improvement would prevent unnoticed leaks of values that end up in the queue despite having the sink already unusable. In concurrent scenarios, with the proposed implementation I think there is a risk that the cancellation is about to be triggered while an item is offered.

`T1`: cancelling thread
`T2`: emitting thread

Timeline:

```
T1: ---------------------------- L607:terminate() -- L610:q.clear() ------------------->
T2: L253:isCancelled()-> false -------------------------------------- L279:q.offer(t) ->
```

As far as I understand, the access to this `Sink` implementation is serialized however only from the producer perspective, so this race is possible. Probably you can confirm that using a JCStress test (it would be neat to add one for this case).

Do you agree about the risk? Would you be willing to verify and improve the implementation? Thanks in advance for your effort!

chemicL (changes requested):
Thanks so much for the effort so far! 🎉 

I think the proposed implementation which checks for cancellation upon emission attempts looks simple and elegant. Unfortunately, I'm not sure it's correct. It would definitely improve the situation. However, I am worried about potential races. As I understand this improvement would prevent unnoticed leaks of values that end up in the queue despite having the sink already unusable. In concurrent scenarios, with the proposed implementation I think there is a risk that the cancellation is about to be triggered while an item is offered.

`T1`: cancelling thread
`T2`: emitting thread

Timeline:

```
T1: ---------------------------- L607:terminate() -- L610:q.clear() ------------------->
T2: L253:isCancelled()-> false -------------------------------------- L279:q.offer(t) ->
```

As far as I understand, the access to this `Sink` implementation is serialized however only from the producer perspective, so this race is possible. Probably you can confirm that using a JCStress test (it would be neat to add one for this case).

Do you agree about the risk? Would you be willing to verify and improve the implementation? Thanks in advance for your effort!

chemicL:
@kaqqao are you still interested in this?

chemicL:
Closing due to lack of activity. Please reopen if you'd like to follow-up.

bajibalu:
Hi @chemicL I am interested to work on this issue if this issue is still open for pickup. Ideally, the queue should not accept any more items once the sink is canceled. Wondering would adding a function to the `SpscArrayQueue` / `SpscArrayQueueCold` to freeze the queue will help?

PR #3732, opened 2024-02-22
Cleanup changelog and dependency upgrades

chemicL:
Excluded chores, dependency upgrades and test changes from the release notes. Also, labeling transitive or shaded dependencies upgraded by Dependabot as enhancements to be included in the changelog.

### DIFF
--- a/.github/release.yml
+++ b/.github/release.yml
@@ -2,6 +2,9 @@ changelog:
   exclude:
     labels:
       - ignore-changelog
+      - type/test
+      - type/chore
+      - type/dependency-upgrade
   categories:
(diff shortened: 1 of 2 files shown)


.github/release.yml
```diff
@@ -2,6 +2,9 @@ changelog:
   exclude:
     labels:
       - ignore-changelog
+      - type/test
+      - type/chore
```
violetagg, line 6: below it was `type/chores` while here it is missing the `s`

chemicL: nice catch, thanks

violetagg (approved):

chemicL:
Thanks @violetagg !

PR #3733, opened 2024-02-22
Throw wrapped `TimeoutException` on `Mono.block*` and `Flux.block*`

### LINKED ISSUE #3709: Timeout behavior in Mono.block(timeout)
As already posted in https://github.com/reactor/reactor-core/issues/3519 , it would be very useful to catch a more specific exception when Mono/Flux.block fails with an IllegalStateException, because this error could be likely retried later.
Catching and parsing the String message "Timeout on blocking read for ..." is a very ugly "fallback" solution for this use case.

Throwing an IllegalStateException with the same current message and nesting a TimeoutException exception (or other exceptions) could be a good option. This would not break other users experience who are already manually matching the exception message and would let other users to catch a specific root cause.
Of course, throwing a new RuntimeException specific subclass for timeout will be even better, if no other users are matching message and/or the current exception type

injae-kim:
Throw wrapped `TimeoutException` on `Mono.block*` and `Flux.block*` so that user can use this when timeout instead of digging into IllegalStateException message.

Fixes #3709.

### DIFF
--- a/reactor-core/src/main/java/reactor/core/publisher/BlockingOptionalMonoSubscriber.java
+++ b/reactor-core/src/main/java/reactor/core/publisher/BlockingOptionalMonoSubscriber.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2016-2023 VMware Inc. or its affiliates, All Rights Reserved.
+ * Copyright (c) 2016-2024 VMware Inc. or its affiliates, All Rights Reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -19,6 +19,7 @@ package reactor.core.publisher;
 import java.util.Optional;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
 
 import org.reactivestreams.Subscription;
 import reactor.core.Disposable;
@@ -148,7 +149,8 @@ final class BlockingOptionalMonoSubscriber<T> extends CountDownLatch
 			try {
 				if (!await(timeout, unit)) {
 					dispose();
-					throw new IllegalStateException("Timeout on blocking read for " + timeout + " " + unit);
+					String errorMessage = "Timeout on blocking read for " + timeout + " " + unit;
+					throw new IllegalStateException(errorMessage, new TimeoutException(errorMessage));
 				}
 			}
 			catch (InterruptedException ex) {
--- a/reactor-core/src/main/java/reactor/core/publisher/BlockingSingleSubscriber.java
+++ b/reactor-core/src/main/java/reactor/core/publisher/BlockingSingleSubscriber.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2016-2023 VMware Inc. or its affiliates, All Rights Reserved.
+ * Copyright (c) 2016-2024 VMware Inc. or its affiliates, All Rights Reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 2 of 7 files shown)


reactor-core/src/main/java/reactor/core/publisher/BlockingSingleSubscriber.java
```diff
@@ -124,7 +125,8 @@ final T blockingGet(long timeout, TimeUnit unit) {
 			try {
 				if (!await(timeout, unit)) {
 					dispose();
-					throw new IllegalStateException("Timeout on blocking read for " + timeout + " " + unit);
+					String errorMessage = "Timeout on blocking read for " + timeout + " " + unit;
+					throw new IllegalStateException(errorMessage, new TimeoutException(errorMessage));
```
injae-kim, line 129: https://github.com/reactor/reactor-core/blob/bbc9bd42392dd4908eefe601432db2e7fc5a5613/reactor-core/src/main/java/reactor/core/publisher/Mono.java#L1800

https://github.com/reactor/reactor-core/blob/bbc9bd42392dd4908eefe601432db2e7fc5a5613/reactor-core/src/main/java/reactor/core/publisher/Flux.java#L2786

https://github.com/reactor/reactor-core/blob/bbc9bd42392dd4908eefe601432db2e7fc5a5613/reactor-core/src/main/java/reactor/core/publisher/Flux.java#L2836

I checked that `Mono#block(timeout), Flux#blockFirst(timeout), Flux#blockLast(timeout)` use this code~!

injae-kim: https://github.com/reactor/reactor-core/blob/bbc9bd42392dd4908eefe601432db2e7fc5a5613/reactor-core/src/main/java/reactor/core/publisher/Mono.java#L1787

```
// AS-IS
If the provided timeout expires, a {@link RuntimeException} is thrown.

// TO-BE
If the provided timeout expires, a {@link RuntimeException} is thrown with {@link TimeoutException} as the cause.
```

(just curiosity) should we enhance document on `Mono#block*, Flux#block*` like above?

chemicL: Yes, please.

chemicL (changes requested):
Thanks @injae-kim ! Great job. Do you mind adding the bit about the cause to the javadocs as you suggested?

reactor-core/src/test/java/reactor/core/publisher/BlockingOptionalMonoSubscriberTest.java
```diff
@@ -103,7 +104,8 @@ public void timeoutOptionalTimingOut() {
 		// Using sub-millis timeouts after gh-1734
 		assertThatExceptionOfType(IllegalStateException.class)
 				.isThrownBy(() -> source.blockOptional(Duration.ofNanos(100)))
-				.withMessage("Timeout on blocking read for 100 NANOSECONDS");
+				.withMessage("Timeout on blocking read for 100 NANOSECONDS")
+				.withCause(new TimeoutException("Timeout on blocking read for 100 NANOSECONDS"));
```
He-Pin, line 108: Update the Javadoc too

reactor-core/src/main/java/reactor/core/publisher/BlockingOptionalMonoSubscriber.java
```diff
@@ -148,7 +149,8 @@ final Optional<T> blockingGet(long timeout, TimeUnit unit) {
 			try {
 				if (!await(timeout, unit)) {
 					dispose();
-					throw new IllegalStateException("Timeout on blocking read for " + timeout + " " + unit);
+					String errorMessage = "Timeout on blocking read for " + timeout + " " + unit;
+					throw new IllegalStateException(errorMessage, new TimeoutException(errorMessage));
```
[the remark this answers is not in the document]
injae-kim: ✅ update javadoc :)

reactor-core/src/main/java/reactor/core/publisher/BlockingSingleSubscriber.java
injae-kim: ✅ update javadoc :)

chemicL (approved):
Thank you, @injae-kim 👏

PR #3736, opened 2024-03-01
Add `doAfterSubscirbe` method on Flux and Mono

### LINKED ISSUE #3286: `doAfterSubscribe` operator similar to `doFirst`
@simonbasle
 > alternative idea: add it to SignalListener with a default NO-OP implementation

injae-kim:
Add `doAfterSubscirbe` method on Flux and Mono.

Fixes #3286.

### DIFF
--- a/reactor-core/src/main/java/reactor/core/publisher/Flux.java
+++ b/reactor-core/src/main/java/reactor/core/publisher/Flux.java
@@ -4800,7 +4800,7 @@ public abstract class Flux<T> implements CorePublisher<T> {
 	 */
 	public final Flux<T> doAfterTerminate(Runnable afterTerminate) {
 		Objects.requireNonNull(afterTerminate, "afterTerminate");
-		return doOnSignal(this, null, null, null, null, afterTerminate, null, null);
+		return doOnSignal(this, null, null, null, null, null, afterTerminate, null, null);
 	}
 
 	/**
@@ -4817,7 +4817,7 @@ public abstract class Flux<T> implements CorePublisher<T> {
 	 */
 	public final Flux<T> doOnCancel(Runnable onCancel) {
 		Objects.requireNonNull(onCancel, "onCancel");
-		return doOnSignal(this, null, null, null, null, null, null, onCancel);
+		return doOnSignal(this, null, null, null, null, null, null, null, onCancel);
 	}
 
 	/**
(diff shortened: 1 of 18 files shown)


chemicL (changes requested):
Hi, @injae-kim 👋 Thanks for the attempt and the work you put into this. I'll comment in the issue in a moment to start the discussion around the need for this change.

Regarding the PR, please consider the following test that I came up with based on one that you created:

	@Test
	public void syncdoAfterSubscribeCalledInOrder() {

		AtomicInteger onSubscribe = new AtomicInteger();
		AtomicInteger afterSubscribe = new AtomicInteger();
		AtomicInteger onRequest = new AtomicInteger();
		AtomicInteger onNext1 = new AtomicInteger();
		AtomicInteger onNext2 = new AtomicInteger();
		AtomicInteger onComplete = new AtomicInteger();

		AssertSubscriber<Object> ts = AssertSubscriber.create();

		Flux.range(1, 2)
		    .hide()
		    .doOnSubscribe(s -> onSubscribe.set(counter.incrementAndGet()))
		    .doAfterSubscribe(s -> afterSubscribe.set(counter.incrementAndGet()))
			.doOnRequest(r -> onRequest.set(counter.incrementAndGet()))
			.doOnNext(i -> (i == 1 ? onNext1 : onNext2).set(counter.incrementAndGet()))
			.doOnComplete(() -> onComplete.set(counter.incrementAndGet()))
		    .subscribe(ts);

		ts.assertNoError()
		  .assertValues(1, 2)
		  .assertComplete();

		String[] inOrder = new String[6];
		inOrder[onSubscribe.get() - 1] = "onSubscribe";
		inOrder[afterSubscribe.get() - 1] = "afterSubscribe";
		inOrder[onRequest.get() - 1] = "onRequest";
		inOrder[onNext1.get() - 1] = "onNext1";
		inOrder[onNext2.get() - 1] = "onNext2";
		inOrder[onComplete.get() - 1] = "onComplete";

		System.out.println("Actions in order: " + Arrays.toString(inOrder));

		assertThat(onSubscribe.get()).isEqualTo(1);
		assertThat(afterSubscribe.get()).isEqualTo(2);
		assertThat(onRequest.get()).isEqualTo(3);
		assertThat(onNext1.get()).isEqualTo(4);
		assertThat(onNext2.get()).isEqualTo(5);
		assertThat(onComplete.get()).isEqualTo(6);
	}

And the output:

```
Actions in order: [onSubscribe, onRequest, onNext1, onNext2, onComplete, afterSubscribe]


expected: 2
 but was: 6
```

I believe the current setup makes this implementation not really useful, as in my view, the callback should be invoked immediately after the subscription is delivered, but before any request is propagated. That's making it harder to implement I suppose. On the other hand, perhaps it works as intended, but then it seems confusing that all the other actions happen before this callback is invoked. Then if the real subscription consumer didn't immediately drain the source, but hold off for more demand, the callback would be invoked somewhere in the middle. So this would really be a very advanced operator for power users.

injae-kim:
Close this PR cause `doAfterSubscribe` is not actually needed now, see https://github.com/reactor/reactor-core/issues/3286#issuecomment-1986844034

Issue #3772, opened 2024-03-21
BoundedElasticThreadPerTaskSchedulerTest > ensuresTasksScheduling() FAILED


chemicL:
```
BoundedElasticThreadPerTaskSchedulerTest > ensuresTasksScheduling() FAILED
    org.opentest4j.AssertionFailedError: 
    Expecting value to be true but was false
        at jdk.internal.reflect.DirectConstructorHandleAccessor.newInstance(DirectConstructorHandleAccessor.java:62)
        at java.lang.reflect.Constructor.newInstanceWithCaller(Constructor.java:502)
        at reactor.core.scheduler.BoundedElasticThreadPerTaskSchedulerTest.ensuresTasksScheduling(BoundedElasticThreadPerTaskSchedulerTest.java:78)
```

* Reactor version(s) used: 3.6.4
* JVM version (`java -version`): Java_Temurin-Hotspot_jdk/21.0.2-13/x64
* OS and version (eg `uname -a`): Ubuntu 22.0 LTS

injae-kim:
```java
	@Test
	public void ensuresTasksScheduling() throws InterruptedException {
		CountDownLatch latch = new CountDownLatch(1);

		Disposable disposable = scheduler.schedule(latch::countDown);

		Assertions.assertThat(latch.await(5, TimeUnit.SECONDS)).isTrue();

                // AS-IS
		Assertions.assertThat(disposable.isDisposed()).isTrue();
                // TO-BE
		await().untilAsserted(() -> disposable.isDisposed()).isTrue());
	}
```
Hmm how about using `await().untilAsserted(..)` to check `disposable.isDisposed()).isTrue()`?

Cause `disposable.dispose()` is not finished yet right after latch reaches zero!

chemicL:
That sounds correct. Are you interested in a contribution?

injae-kim:
https://github.com/reactor/reactor-core/pull/3779 my friend who really loves opensource create fix PR! PTAL 🙇

PR #3779, opened 2024-03-30
#3772: wrapped with `await().untilAsserted`

chickenchickenlove:
Wrapped `Assertions.assertThat(disposable.isDisposed()).isTrue()` with `Awaitility.await()`. 
Because `disposable.dispose()` is not finished yet right after latch reaches zero.

Fixes https://github.com/reactor/reactor-core/issues/3772

### DIFF
--- a/reactor-core/src/test/java21/reactor/core/scheduler/BoundedElasticThreadPerTaskSchedulerTest.java
+++ b/reactor-core/src/test/java21/reactor/core/scheduler/BoundedElasticThreadPerTaskSchedulerTest.java
@@ -75,7 +75,7 @@ class BoundedElasticThreadPerTaskSchedulerTest {
 		Disposable disposable = scheduler.schedule(latch::countDown);
 
 		Assertions.assertThat(latch.await(5, TimeUnit.SECONDS)).isTrue();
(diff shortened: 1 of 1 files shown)


chickenchickenlove:
@chemicL Thanks for your comment 🙇‍♂️ 
I executed this command to rebase to 3.6.x.
Please take a look when you have free time! 
```shell
$ git remote add upstream https://github.com/reactor/reactor-core.git
$ git checkout -b 3.6.x upstrea/3.6.x
$ git checkout 240330-test-fix
$ git rebase 3.6.x # conflict occurs at gradle.properties (i resolved it)
$ git push --force
```

chemicL:
Thanks! I changed the base and I think there won't be any issue, but next time consider using `--onto` as an argument to `rebase`. [Check here for justifications](https://stackoverflow.com/questions/33942588/difference-between-rebase-master-and-rebase-onto-master-from-a-branch-deri). When the tests pass we can merge it :)

chickenchickenlove:
I see. thanks for your kindful comments 🙇‍♂️

chemicL (approved):

chemicL:
Thank you for the contribution @chickenchickenlove 🚀

PR #3783, opened 2024-04-10
Remove Flux#thenMany optimization breaking concat

### LINKED ISSUE #3777: thenMany does not ignore all emissions of a concat due to incorrect optimization
## Expected Behavior

`Flux.thenMany` should ignore all emissions of a `Flux.concat`

## Actual Behavior

It only ignores emissions of the last concatted publisher due to an incorrect optimization.

## Steps to Reproduce

```java
@Test
void reproCase() {
  Flux.concat(Flux.just("a"), Flux.just("b"))
            // adding a .hide() here fixes the test case
            .thenMany(Flux.just("c"))
            .as(StepVerifier::create)
            .expectNext("c")
            .verifyComplete();
}
```

Result:
> java.lang.AssertionError: expectation "expectNext(c)" failed (expected value: c; actual value: a)

Adding the `.hide()` or enabling `Hooks.onOperatorDebug()` fixes the test case.

## Possible Solution

`Flux.thenMany` uses an incorrect optimization:

```java
    if (this instanceof FluxConcatArray) {
	FluxConcatArray<T> fluxConcatArray = (FluxConcatArray<T>) this;
	return fluxConcatArray.concatAdditionalIgnoredLast(other); // <- incorrect!
    }
```

## Your Environment
* Reactor version(s) used: 3.6.1
* JVM version (`java -version`): openjdk version "21.0.1" 2023-10-17 LTS

chemicL:
The optimization incorrectly ignored only the last Publisher instead of all currently concatenated Publishers causing undesired values to be emitted.

Fixes #3777

### DIFF
--- a/reactor-core/src/main/java/reactor/core/publisher/Flux.java
+++ b/reactor-core/src/main/java/reactor/core/publisher/Flux.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2016-2022 VMware Inc. or its affiliates, All Rights Reserved.
+ * Copyright (c) 2016-2024 VMware Inc. or its affiliates, All Rights Reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


chemicL:
@violetagg thank you for the review :)

PR #3789, opened 2024-04-18
Drain SinkManyEmitterProcessor buffer after cancel

### LINKED ISSUE #3715: Auto-cancelled Sink still accepts emissions

bajibalu:
The internal buffer/queue in SinkManyEmitterProcessor will be drained after all the subscriptions are canceled.

As explained [here](https://github.com/reactor/reactor-core/issues/3715#issuecomment-2060902010) the queue/buffer in SinkManyEmitterProcessor is not drained properly after the last subscriber canceled the subscription. This was happening due to the WIP marker is left in an unclean state. This PR fixes the issue by updating the WIP marker. I am not sure if this is the ideal approach though.

Fixes #3715

### DIFF
--- a/reactor-core/src/main/java/reactor/core/publisher/SinkManyEmitterProcessor.java
+++ b/reactor-core/src/main/java/reactor/core/publisher/SinkManyEmitterProcessor.java
@@ -384,6 +384,7 @@ final class SinkManyEmitterProcessor<T> extends Flux<T> implements InternalManyS
 
 	final void drain() {
 		if (WIP.getAndIncrement(this) != 0) {
+			WIP.decrementAndGet(this);
 			return;
 		}
 
@@ -398,6 +399,7 @@ final class SinkManyEmitterProcessor<T> extends Flux<T> implements InternalManyS
 			boolean empty = q == null || q.isEmpty();
 
 			if (checkTerminated(d, empty)) {
+				WIP.addAndGet(this, -missed);
 				return;
 			}
 
@@ -432,6 +434,7 @@ final class SinkManyEmitterProcessor<T> extends Flux<T> implements InternalManyS
 						v = null;
 					}
 					if (checkTerminated(d, v == null)) {
+						WIP.addAndGet(this, -missed);
 						return;
 					}
 					if (sourceMode != Fuseable.SYNC) {
@@ -459,6 +462,7 @@ final class SinkManyEmitterProcessor<T> extends Flux<T> implements InternalManyS
 					empty = v == null;
 
 					if (checkTerminated(d, empty)) {
+						WIP.addAndGet(this, -missed);
 						return;
 					}
 
@@ -593,6 +597,7 @@ final class SinkManyEmitterProcessor<T> extends Flux<T> implements InternalManyS
 				//happens when the removed inner makes the subscribers array EMPTY
 				if (autoCancel && b == EMPTY && Operators.terminate(S, this)) {
 					if (WIP.getAndIncrement(this) != 0) {
+						WIP.decrementAndGet(this);
 						return;
 					}
 					terminate();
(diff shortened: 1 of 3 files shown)


reactor-core/src/main/java/reactor/core/publisher/SinkManyEmitterProcessor.java
```diff
@@ -383,10 +383,12 @@ public Object scanUnsafe(Attr key) {
 	}
 
 	final void drain() {
-		if (WIP.getAndIncrement(this) != 0) {
+		if (WIP.get(this) != 0) {
```
bajibalu, line 386: In the cases where more than 1 thread calling the drain function, all the threads are incrementing the WIP counter while only one of them proceeds to clear the queue and decrements the counter. This leaves the WIP in unclean state.

reactor-core/src/main/java/reactor/core/publisher/SinkManyEmitterProcessor.java
```diff
@@ -398,6 +400,7 @@ final void drain() {
 			boolean empty = q == null || q.isEmpty();
 
 			if (checkTerminated(d, empty)) {
+				WIP.addAndGet(this, -missed);
```
bajibalu, line 403: Similarly, these returns also leave the WIP in an unclean state.

reactor-core/src/main/java/reactor/core/publisher/SinkManyEmitterProcessor.java
chemicL: By separating the atomic operation into two operations you break the exclusive access to the critical section below. Please start with creating JCStress tests if you're considering to spend more time on this. These drain methods are quite a critical piece and ensuring proper lock-free coordination between different actors is essential. Another aspect is performance so that we make as little volatile accesses as we can.

chemicL (changes requested):

bajibalu:
@chemicL I addressed your comments and also added a JCStress test. But I am not sure whether the test case is correct. #

chemicL:
Hey. Thank you for the effort. Unfortunately, I'm not able to accept this in the current form. There are multiple issues here and it would require quite an effort on my part to help correct these. The issues I can see are:
* WIP doesn't seem to be correctly understood -> this is a mechanism to signal a Thread that is currently draining that more work needs to be considered and also as means to guard access to the critical section in a lock-free fashion
* in case of cancellation, the return value can't be "OK", but rather "FAIL_CANCELLED"
* the JCStress test should not consider an item stuck in the queue forever as "acceptable, interesting" but rather fail the evaluation

In general, I think the `SinkManyEmitterProcessor` is a super critical piece in the `Sinks` API and should be handled with care. The attempt to fix it should consider a broad set of circumstances and would require means to atomically represent the cancelled state and a way to reject new additions and ensure proper cleanup in the face of high concurrency. With that, I'm sorry to reject the PR.

PR #3790, opened 2024-04-19
Add `@Incubating` annotation

### LINKED ISSUE #3647: add `@Incubating` annotation
# Rationale 

Reactor is exploring new horizons so we need to have safe harbour for some of the enhancements we add. Functionality marked with `@Incubating` annotation would let us make breaking changes in functionality between minor incrmenets

violetagg:
Fixes #3647

### DIFF
--- a/reactor-core/src/main/java/reactor/util/annotation/Incubating.java
+++ b/reactor-core/src/main/java/reactor/util/annotation/Incubating.java
@@ -1,0 +1,36 @@
+/*
+ * Copyright 2024 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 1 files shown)


reactor-core/src/main/java/reactor/util/annotation/Incubating.java
```diff
@@ -0,0 +1,36 @@
...
+package reactor.util.annotation;
+
+import java.lang.annotation.Documented;
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * This signifies that the feature is currently in a developmental phase.
```
chemicL, line 26:
```suggestion
 * Signifies that the annotated functionality is currently in a developmental phase.
```

reactor-core/src/main/java/reactor/util/annotation/Incubating.java
```diff
@@ -0,0 +1,36 @@
...
+ * An API marked as {@code @Incubating} is subject to potential modifications without prior notice.
```
chemicL, line 27:
```suggestion
 * An API marked as {@code @Incubating} is subject to potential modifications (including removal) without prior notice.
```

chemicL:
I left two suggestions about the javadoc content, I think they can help set expectations better.

chemicL (approved):

PR #3797, opened 2024-04-27
Update to put a more clear message about the requirement of Java 8

otaviojava:
It is a minor enhancement at the readme about the minimum requirement of Java 8.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ Since `3.3.x`, this repository also contains `reactor-tools`, a java agent aimed
 
 ## Getting it
    
-**Reactor 3 requires Java 8 or + to run**.
+**The Minimum requirement for Reactor 3 is Java 8 or higher**.
 
 With Gradle from repo.spring.io or Maven Central repositories (stable releases only):
 


README.md
```diff
@@ -15,7 +15,7 @@ Since `3.3.x`, this repository also contains `reactor-tools`, a java agent aimed
 
 ## Getting it
    
-**Reactor 3 requires Java 8 or + to run**.
+**The Minimum requirement for Reactor 3 is Java 8 or higher**.
```
chemicL, line 18: Thanks for the suggestion. I'm not a native speaker so I might not know best, but it feels like the proposed sentence is not as concise and the caps/non-caps are not consistent. Why would "Minimum" be capitalized? To be honest, the current flow of the sentence makes more sense to me. Would "Reactor 3 requires Java 8 or higher to run" be acceptable? This way, only the "+" sign is replaced with "higher" and maintains the conciseness.

chemicL (changes requested):
Thanks. Please see my comment referring to the proposed change.

PR #3798, opened 2024-04-28
Feature/3340 Add best-effort replay sink

MikkelHJuul:
This PR adds the ArraySizeBoundReplayBuffer, and links it for the user via `Sink.many().replay().bestEffort()`.

This new spec offer support for the purely size-bound buffer implementations. It adds `#limit(n)`, `#latest()` and `#latestOrDefault(T)`, reflecting the same methods from its parent, `Sink.many().replay()`.
The `ArraySizeBoundReplayBuffer` in contrast to the `SizeBoundReplayBuffer` will never hold more than the stated items referenced. The `SizeBoundReplayBuffer`, through the `Subscription`, will retain old references, such that they can be replayed to the subscriber. This is very correct, but can also make the buffer grow beyond its bounds, if a subscriber lacks request they will still be able to replay all elements, but these elements remain in memory. The `ArraySizeBoundReplayBuffer` hold and replay only what is kept in the array.

Fixes issues/3340

### DIFF
--- a/reactor-core/src/jcstress/java/reactor/core/publisher/SinksManyReplayLatestStressTest.java
+++ b/reactor-core/src/jcstress/java/reactor/core/publisher/SinksManyReplayLatestStressTest.java
@@ -1,0 +1,68 @@
+package reactor.core.publisher;
+
+import org.openjdk.jcstress.annotations.Actor;
+import org.openjdk.jcstress.annotations.Arbiter;
+import org.openjdk.jcstress.annotations.JCStressTest;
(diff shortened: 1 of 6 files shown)


MikkelHJuul:
I still need a few tests to document this further

MikkelHJuul:
I'm having pipeline issues with gradle on M1 Mac, so I'm stuck

chemicL:
Closing as explained in https://github.com/reactor/reactor-core/issues/3340#issuecomment-2242821598.

PR #3818, opened 2024-05-28
Don't cleanup docs zip in case of build errors

pderop:
The publication process triggered by the CI (controlled by publish.yml) may occasionally fail (like when having some flaky tests). In such cases, the cleanup job always runs and deletes any previously uploaded docs zip files created by the `build-docs-zip` job. However, it's inconvenient to remove the docs-zip when certain jobs fail because if we attempt to rerun the failed jobs later, we lose access to the docs-zip attached to the workflow run.

Ideally, the docs-zip should only be removed when the publish process succeeds. If errors occur, such as flaky tests, the docs-zip should not be removed. This ensures that we can rerun the publish jobs successfully.

It's important to note that the docs-zip is set to be retained for 1 day. If a failed jobs is not rerun within this timeframe, the docs-zip will then be automatically removed from the workflow run after one day.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -95,7 +95,7 @@ jobs:
         with:
           name: ${{ env.DOCS_ZIP }}
           path: ${{ env.DOCS_ZIP_PATH }}
-          retention-days: 1
+          retention-days: 3
           if-no-files-found: error
 
   #deploy the snapshot artifacts to Artifactory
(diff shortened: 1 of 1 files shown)


pderop:
in commit https://github.com/reactor/reactor-core/pull/3818/commits/0a911a64d8a3f61b3c9da2f53ce97ac289b7e096:

- the cleanup job is now run only if either deploySnapshot, or tagRelease, or tagMilestone job status is in success. It means that if prepare job fails and needs to be re-run, then it will work because the docs-zip artifact won't be deleted from the workflow run.
- Also, increased the docs zip artifact retention day to 3, in case one is having trouble while doing a release, and needs to wait one more day, before re-running any deployX job

@chemicL , can you please take a look ?

pderop:
notice that I tried to shorten the cleanup condition like this:

```
cleanup:
  needs: [ deploySnapshot, tagRelease, tagMilestone ]
  if: always() && contains(needs.*.result, 'success')
```

but this simply just does not work, see https://github.com/actions/runner/issues/1540

chemicL (approved):

pderop:
Thanks @chemicL for the review and for the merge.

PR #3819, opened 2024-05-29
Fail on missing asciidoctor pdf command

pderop:
When you want to build the documentation, you can using the following command:

```
 ./gradlew docs -PforcePdf
```

Now, PDF generation requires the `asciidoctor-pdf` command to be available in the PATH.
This PR simply makes the build failing in case the `asciidoctor-pdf` command is not available from the PATH.
Also, updated README.md to indicate that this command needs to be available when generating PDF from Antora documentation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,7 +99,18 @@ Then you can build the antora documentation like this:
 ./gradlew docs
 ```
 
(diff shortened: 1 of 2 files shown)


chemicL (approved):
Thanks!

Issue #3840, opened 2024-07-04
Huge Performance Overhead From ContextPropagation With Mircrometer Tracing


rady66:
## Expected Behavior
We would not expect more than 5% overhead from logging capabilities as industry standard. (Not sure if the problem is with the reactor project or micrometer capabilities)

## Actual Behavior
Huge overhead from micrometer - more than 25-30%

## Steps to Reproduce
Simple request in spring webflux (same jvm process) where we want to have trace-id and correlate logs. See the profiler results showing huge overhead in the context propagation for setting thread locals:
![image](https://github.com/reactor/reactor-core/assets/17241575/d7ab0d1e-cba7-4650-88e9-7f567cf61caf)

## Your Environment

* Reactor version(s) used:
reactor-core:3.6.7
* Other relevant libraries versions (eg. `netty`, ...):
spring-webflux:3.3.1
io.micrometer:1.13.1
* JVM version (`java -version`): java17
* OS and version (eg `uname -a`): windows 11

PeterPaul-Perfana:
We also noticed large increase of CPU usage when upgrading from reactor 3.5 to 3.6 in Spring Boot with micrometer. See this post for the details: https://stackoverflow.com/questions/78677203/moving-from-reactor-3-5-to-3-6-in-spring-flux-shows-large-increase-in-cpu-due-to. There is also a test project that shows the behaviour: https://github.com/stokpop/reactive-tracing-issue. 

We notice a lot more calls from `FluxContextWriteRestoringThreadLocals$ContextWriteRestoringThreadLocalsSubscriber.onNext()` and even more calls to `decorateScope`. This seems similar to the profiler screenshot 👆.

chemicL:
Hi @rady66, and @PeterPaul-Perfana !

Let me start by clarifying some aspects.

> We would not expect more than 5% overhead from logging capabilities as industry standard.

I appreciate that point of view. However, please bear in mind that what you are demonstrating with the screenshot (somehow it's now not able to open, results in 404 :/) involves:

* `MDC` using SLF4J and Logback as the implementation - obviously for **logging** purposes.
* `Observation` - a Micrometer concept, which provides capabilities for **logging**, **metrics**, and also **tracing** with extension points for more capabilities.
* `TracingObservationHandler` using Brave -> that means you are also expecting **tracing** capabilities.

[...]

PR #3845, opened 2024-07-11
Skip Automatic Context Propagation in special operators

chemicL:
This change should improve performance of Automatic Context Propagation in certain cases when `doOnDiscard`, `onErrorContinue`, and `onErrorStop` are used.

The context-propagation integration requires `contextWrite` and `tap` operators to be barriers for restoring `ThreadLocal` values. Some internal usage of `contextWrite` does not require us to treat the operators the same way and we can skip the ceremony of restoring `ThreadLocal` state as we know that no `ThreadLocalAccessor` can be registered for them. Therefore, a private variant is introduced to avoid unnecessary overhead when not required.

Related #3840


violetagg (approved):

chemicL:
@violetagg thank you for the review!

PR #3848, opened 2024-07-16
Skip automatic context-propagation in Flux.generate

chemicL:
As `Flux.generate` uses `SynchronousSink`, which should not be used asynchronously, we can eliminate the unnecessary `ThreadLocal` restoration from this operator.

Related to #3840.


violetagg (approved):

chemicL:
Thanks for the review @violetagg :)

PR #3854, opened 2024-07-22
Allow registering a custom `Predicate` for determining non-blocking threads

### LINKED ISSUE #3833: Provide more ways to mark a thread as non-blocking
## Motivation

We maintain a reactive web framework called Armeria (https://armeria.dev). It allows a user to implement a reactive web application using the Reactive Streams implementation of their choice. A user can choose from Armeria's own Reactive Streams implementation, Reactor and RxJava. Therefore, we don't have any mandatory dependency on Reactor and RxJava to keep our dependencies lean and stay unopinionated, and we want to keep it this way.

When a user writes their application using Armeria and Reactor, Armeria needs to tell Reactor that its event loop threads are non-blocking. Reactor currently determines whether a thread is non-blocking or not using `Schedulers.isInNonBlockingThread()` and `Schedulers.isNonBlockingThread()` ([here](https://github.com/reactor/reactor-core/blob/85959a5d2874d8d8deb0ca8b010d744611c0a403/reactor-core/src/main/java/reactor/core/scheduler/Schedulers.java#L660-L680)).

It means, for Armeria to mark a thread as non-blocking, Armeria needs to make its event loop thread class implement the `NonBlocking` interface, e.g.:

```java
public class ArmeriaEventLoopThread extends Thread implements NonBlocking { ... }
```

However, because Armeria has no mandatory dependency on Reactor, JVM will fail to load `ArmeriaEventLoopThread` when Reactor is not available in the classpath, if we introduce `NonBlocking` into our class hierarchy.

## Desired solution

Provide an alternative way to mark a thread (or thread class) as non-blocking, so that Armeria team can define a non-blocking thread without introducing mandatory runtime dependency on `reactor-core`. There are a few options to solve this:

- Provide a small separate JAR that contains `NonBlocking` (and potentially other tagging interfaces) and nothing more.
  - Straightforward change with no additional overhead and a minimal dependency size increase.
  - However, it requires reorganization in your codebase.
  - `reactor-core` needs to depend on the new JAR, which may make things a little bit more complicated?
- Java SPI
  - Let a user define an SPI implementation that's loaded by Reactor in runtime.
  - This will work because: 1) our threads will not depend on any Reactor classes and 2) the SPI implementation will be loaded only when Reactor is initialized.
  - However, it will perhaps increase the overhead.
- Provide API to register a subclass of `Thread` as non-blocking.
  - Let a user call a new Reactor API that registers the given `Thread` subclass to the list of non-blocking thread types.
  - This is similar to Java SPI option, but Armeria team will need to call this API using reflection. (We're OK doing that though.)
  - Same overhead concern.
- Check the class name
  - Just check if the FQCN contains `NonBlocking`, as well as the `instanceof NonBlocking` check.
  - A little bit of overhead concern.

## Considered alternatives

- As an alternative, we defined the identical duplicate `NonBlocking` interface in our codebase. This works because `NonBlocking` is just a tag interface and thus having a duplicate class in the classpath won't hurt. However, this doesn't work with Java Modules, which never allows having the same classes in more than one module.
- We could simply add `reactor-core` as a mandatory dependency, but our users will not be happy with it because it'd make their dependency tree bigger even if they don't use Reactor.

trustin:
Related issue: #3833
Motivation:

It is currently not possible to create a non-blocking threads without implementing the `reactor.core.scheduler.NonBlocking` interface. Some third-party libraries and frameworks don't directly depend on `reactor-core`, yet they want to mark the threads they manage as non-blocking.

Modifications:

- Added a new method `Schedulers.registerNonBlockingThreadPredicate()` so that a user can register their own `Predicate` that determines whether a given thread is non-blocking or not
  - Also added `Schedulers.resetNonBlockingThreadPredicate()` so that a user can unregister all previous `Predicate`s
- Fixed an incorrectly implemented test that doesn't really test anything:
  - `SchedulersTest.isInNonBlockingThreadTrue()`

Result:

- A user can now mark their own `Thread` classes as non-blocking without depending on `reactor-core` or implementing the `NonBlocking` marker interface.

### DIFF
--- a/reactor-core/src/main/java/reactor/core/scheduler/Schedulers.java
+++ b/reactor-core/src/main/java/reactor/core/scheduler/Schedulers.java
@@ -34,6 +34,7 @@ import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.BiConsumer;
 import java.util.function.BiFunction;
 import java.util.function.Function;
+import java.util.function.Predicate;
 import java.util.function.Supplier;
 
 import io.micrometer.core.instrument.MeterRegistry;
@@ -120,6 +121,10 @@ public abstract class Schedulers {
 			Optional.ofNullable(System.getProperty("reactor.schedulers.defaultBoundedElasticOnVirtualThreads"))
 			        .map(Boolean::parseBoolean)
 			        .orElse(false);
+
+	static final Predicate<Thread> DEFAULT_NON_BLOCKING_THREAD_PREDICATE = thread -> false;
+
+	static Predicate<Thread> nonBlockingThreadPredicate = DEFAULT_NON_BLOCKING_THREAD_PREDICATE;
 
 	/**
 	 * Create a {@link Scheduler} which uses a backing {@link Executor} to schedule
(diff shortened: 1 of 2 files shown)


chemicL:
Please run `./gradlew spotlessApply` and push the license header updates.

trustin:
@chemicL It looks like 3.6.x doesn't have Spotless. I can update the license headers of the modified files, but could you point to the right license header?

trustin:
By the way, I ran `./gradlew spotlessApply` against the `main` branch and it gives me this error:

```
> Task :spotlessJava FAILED

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':spotlessJava'.
> org.eclipse.jgit.errors.CorruptObjectException: DIRC checksum mismatch
```

Environment:

```
❯ ./gradlew --version     
------------------------------------------------------------
Gradle 8.5
------------------------------------------------------------

Build time:   2023-11-29 14:08:57 UTC
Revision:     28aca86a7180baa17117e0e5ba01d8ea9feca598

Kotlin:       1.9.20
Groovy:       3.0.17
Ant:          Apache Ant(TM) version 1.10.13 compiled on January 4 2023
JVM:          21.0.3 (Eclipse Adoptium 21.0.3+9-LTS)
OS:           Linux 6.9.10-arch1-1 amd64
```

chemicL:
I'm not sure what's up with your local setup, but there's nothing fishy on our side I think. I will make comments which you can incorporate into the PR and we can merge.

Here's what I did:

```
➜  reactor-core git:(3.6.x) gh pr checkout 3854
[...]
Switched to branch 'custom_nonblocking_thread_predicate_3.6'
➜  reactor-core git:(custom_nonblocking_thread_predicate_3.6) ./gradlew spotlessApply
[...]
BUILD SUCCESSFUL in 8s
7 actionable tasks: 2 executed, 5 up-to-date
```

And running `git diff` gives:

```
diff --git a/reactor-core/src/main/java/reactor/core/scheduler/Schedulers.java b/reactor-core/src/main/java/reactor/core/scheduler/Schedulers.java
index 02dfa31c5..7617cb180 100644
--- a/reactor-core/src/main/java/reactor/core/scheduler/Schedulers.java
+++ b/reactor-core/src/main/java/reactor/core/scheduler/Schedulers.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2016-2023 VMware Inc. or its affiliates, All Rights Reserved.
+ * Copyright (c) 2016-2024 VMware Inc. or its affiliates, All Rights Reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
diff --git a/reactor-core/src/test/java/reactor/core/scheduler/SchedulersTest.java b/reactor-core/src/test/java/reactor/core/scheduler/SchedulersTest.java
index 821fbdb5a..7232a0248 100644
--- a/reactor-core/src/test/java/reactor/core/scheduler/SchedulersTest.java
+++ b/reactor-core/src/test/java/reactor/core/scheduler/SchedulersTest.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2016-2022 VMware Inc. or its affiliates, All Rights Reserved.
+ * Copyright (c) 2016-2024 VMware Inc. or its affiliates, All Rights Reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
```

chemicL (approved):
Thank you for the contribution!

PR #3858, opened 2024-08-01
Try common interface for context

mikrethor:


mikrethor:
Wanted to try CI because I can't make the project compile on my laptop

Issue #3443, opened 2023-04-20
Document when groupBy can deadlock because of beckpressure


OlegDokuka:
recent activity related to groupBy deadlock (#3442 #3427) adds extra motivation to clarify how groupBy has to be used: 

## Documentation Issue
* Document that the inner group MUST be subscribed
* Document that operators such as flatMap/concatMap may cause deadlock because of backpressure

MikkelHJuul:
Just so as to not flood the issues-section I found that the groupBy operator blocks/deadlocks in a case where the flatMap retries. Using backoff on the retry fixes this situation.

e.g.
```java
Flux.just(
    Map.entry(1,1),
    Map.entry(2,1), //error from this blocks everything
    Map.entry(3,3),
    Map.entry(2,2)
).groupBy(Map.Entry::getKey)
.flatMap(grp -> 
    grp.switchMap(e -> Mono.just(e).flatMapMany(this::factory).retry())
)

<T> Flux<T> factory(Map.Entry<Integer, Integer> entry) {
    if (!Objects.equals(entry.getKey(), entry.getValue()))
        return Flux.error(...);
    return Flux.just(...);
}
```

NamrataGuptaRoy:
@OlegDokuka I want to get started with contributing to this. Could you point out where the documentation needs to be added?
I see a good documentation here https://projectreactor.io/docs/core/release/api/reactor/core/publisher/Flux.html#groupBy-java.util.function.Function-

chemicL:
Hey, @NamrataGuptaRoy. There seem to be 4 groupBy overloads. We usually have the relevant content repeated for each overload. I think a dedicated note in each will be useful. Also, as @He-Pin noted, `concatMap` sounds like a good operator to take note of this combination. The suggestion from @MikkelHJuul is also worth exploring and considering how exactly that circumstance can be triggered (what kind of backoff? how does backoff help and regular retry runs into problems?) and making a relevant note in the appropriate operator.

Also, we have a mention of the problems in [the reference documentation](https://projectreactor.io/docs/core/milestone/reference/advancedFeatures/advanced-three-sorts-batching.html#grouping-with-flux-groupedflux). Perhaps adding a bit more examples to the warning section would be helpful.

Perhaps best would be to start with the list of considered additions here in the issue and once we consider the list is cohesive you could open a PR? Thanks, I look forward to your ideas!

NamrataGuptaRoy:
[...]

PR #3872, opened 2024-08-15
[doc] Document when groupBy can deadlock because of backpressure

NamrataGuptaRoy:
Adds documentation for the 4 groupBy overloads about the scenario where backpressure can cause a deadlock with groupBy and suggesting how it could be avoided.
Adds example of deadlock with groupBy and concatMap in [the reference documentation](https://projectreactor.io/docs/core/milestone/reference/advancedFeatures/advanced-three-sorts-batching.html#grouping-with-flux-groupedflux).

See #3443


chemicL (approved):
Thank you. I have some formatting and style remarks but I'll add them as a separate commit.

chemicL:
@NamrataGuptaRoy ef152df80d57713f9d33cb6080d8f2ee2b90d093 contains my alterations - mostly formatting and styling. One thing I wanted to clarify was the notion of consumption rate. I believe that term would have been misleading, as the rate didn't have any significance in the scenarios provided - it was the consumption pattern (concurrency too low) that prevented progress. The rate could have been enormous and nothing would have changed :)

PR #3876, opened 2024-08-20
Call init() on delegate from TimedScheduler

luukveenis:
This fixes an issue where wrapping a `BoundedElasticThreadPerTaskScheduler` with `Micrometer#timedScheduler()` causes it to error out immediately.

The [TimedScheduler](https://github.com/reactor/reactor-core/blob/ef152df80d57713f9d33cb6080d8f2ee2b90d093/reactor-core-micrometer/src/main/java/reactor/core/observability/micrometer/TimedScheduler.java) class doesn't currently override the `init()` method, so it calls the [default](https://github.com/reactor/reactor-core/blob/ef152df80d57713f9d33cb6080d8f2ee2b90d093/reactor-core/src/main/java/reactor/core/scheduler/Scheduler.java#L187-L189) from the interface, which delegates to `start()`. This works fine for other scheduler implementations because they still implement `start()`, but for the virtual thread scheduler, it simply [throws an error](https://github.com/reactor/reactor-core/blob/ef152df80d57713f9d33cb6080d8f2ee2b90d093/reactor-core/src/main/java21/reactor/core/scheduler/BoundedElasticThreadPerTaskScheduler.java#L127-L129). When we wrap it in a timed scheduler, it crashes on this error as soon as `init()` gets called.

### DIFF
--- a/reactor-core-micrometer/src/main/java/reactor/core/observability/micrometer/TimedScheduler.java
+++ b/reactor-core-micrometer/src/main/java/reactor/core/observability/micrometer/TimedScheduler.java
@@ -135,6 +135,11 @@ final class TimedScheduler implements Scheduler {
 		delegate.start();
 	}
 
+	@Override
+	public void init() {
+		delegate.init();
+	}
+
 	static final class TimedWorker implements Worker {
 
 		final TimedScheduler parent;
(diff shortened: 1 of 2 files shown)


chemicL:
@luukveenis thanks! I took the liberty to add a test to your PR.

luukveenis:
> @luukveenis thanks! I took the liberty to add a test to your PR.

Sorry I meant to add one, thank you!

chemicL (approved):

chemicL:
Thank you for the contribution, @luukveenis 🚀

PR #3915, opened 2024-11-04
Add MonoFirstWithValue missing tests

subbarao:
Adding missing tests for MonoFirstWithValue.

MonoFirstWithValue tests which verifies cancellation of in flight publishers when first value is emitted is missing. 
If you comment https://github.com/reactor/reactor-core/blob/main/reactor-core/src/main/java/reactor/core/publisher/FluxFirstWithValue.java#L283 current tests will pass.
In this test case we start three monos with 3 different sleep. Once Mono with lowest delay completes we verify remaining
monos are cancelled.

### DIFF
--- a/reactor-core/src/test/java/reactor/core/publisher/MonoFirstWithValueTest.java
+++ b/reactor-core/src/test/java/reactor/core/publisher/MonoFirstWithValueTest.java
@@ -176,28 +176,28 @@ class MonoFirstWithValueTest {
 	}
 	@Test
 	void cancelInflightMono() {
-		AtomicLong inflightMonos = new AtomicLong(0);
+		AtomicLong cancelledInflightMonos = new AtomicLong(0);
 		Mono<Integer> inflightMono1 = Mono.fromCallable(() -> {
 			Thread.sleep(300);
 			return 1;
-		}).doOnCancel(inflightMonos::getAndIncrement).subscribeOn(Schedulers.boundedElastic());
+		}).doOnCancel(cancelledInflightMonos::getAndIncrement).subscribeOn(Schedulers.boundedElastic());
 
 		Mono<Integer> inflightMono2 = Mono.fromCallable(() -> {
 			Thread.sleep(400);
 			return 2;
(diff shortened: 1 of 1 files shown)


violetagg:
@subbarao We definitely will make a review so bear with us.

chemicL (changes requested):
@subbarao thanks for the PR. Please rewrite your test in a way that avoids `Thread.sleep()` but relies on more reliable and timely synchronization mechanisms, e.g. `CountDownLatch`. Using `Thread.sleep` in tests and relying on actual time passing contributes negatively to the overall CI duration and can also be prone to flakiness due to different scheduling, delays, etc.

subbarao:
> @subbarao thanks for the PR. Please rewrite your test in a way that avoids `Thread.sleep()` but relies on more reliable and timely synchronization mechanisms, e.g. `CountDownLatch`. Using `Thread.sleep` in tests and relying on actual time passing contributes negatively to the overall CI duration and can also be prone to flakiness due to different scheduling, delays, etc.

updated pr

chemicL (approved):
Thank you for applying the review comment and your contribution! 🚢

chemicL:
This test seems to be flaky though. I just had it fail after a PR merge in https://github.com/reactor/reactor-core/actions/runs/11793828102

```
Test class reactor.core.publisher.MonoFirstWithValueTest


MonoFirstWithValueTest > cancelInflightMono() FAILED
    java.lang.AssertionError: 
    Expecting AtomicLong(1) to have value:
      2L
    but did not.
        at reactor.core.publisher.MonoFirstWithValueTest.cancelInflightMono(MonoFirstWithValueTest.java:205)
```

I will revert. Please try evaluating it by running the test e.g. 1k times (Intellij IDEA allows you to do that using non-gradle test run) and make sure there's some synchronization.

chemicL:
@subbarao are you interested in revisiting this topic?

subbarao:
> @subbarao are you interested in revisiting this topic?

will submit pr for this by end of the week.

PR #3921, opened 2024-11-07
Add Flux.unfold

akuleshov7:
### Summary

This PR introduces a new unfold operator to Flux, providing a functional approach for generating a sequence from an initial seed value. The unfold method offers a declarative alternative to the existing generate operator, inspired by similar functionality in Scala and Haskell.

### Motivation

The unfold operator enhances the API by allowing users to construct sequences in a functional style, which may be more familiar to developers coming from FP-centric languages. While functionally similar to generate, unfold focuses on a clean, declarative approach to defining recursive sequences.

### Functional approach

The current SynchronousSink<T> API, while effective, operates at a low level and encourages side effects. In contrast, using a pure function S -> (T, S) eliminates these risks and offers a more reliable way to model state transitions.

The proposed unfold operator introduces a declarative approach to codata definition, where each step produces the next value and state. This leads to clearer, more maintainable code:

```
// evens [0, 2, 4, 6, ...]
Flux.unfold(0, s -> Optional.of(Tuples.of(s, s + 2)));

// fib [0, 1, 1, 2, 3, 5, 8, ...]
Flux.unfold(Tuples.of(0, 1), s -> {
    var a = s.getT1();
    var b = s.getT2();
    return Optional.of(Tuples.of(a, Tuples.of(b, a + b)));
});

// countdown [3, 2, 1]
Flux.unfold(3, s -> s == 0 ? Optional.empty() : Optional.of(Tuples.of(s, s - 1))); Asynchronous variant
```

State-of-the-art libraries like fs2 and zio-streams have set the pattern with unfoldEval and unfoldZIO. In our case, an API like this would be a logical extension:

abstract <A, S> Flux<A> unfoldMono(S init, Function<S, Mono<Optional<Tuple2<A, S>>>> f); However, this is beyond the scope of this PR, as Flux.generate requires immediate state calculation, which doesn't align with asynchronous Mono.

### Next steps

If the proposal is interesting, I’ll add documentation and tests. For a deeper theoretical foundation, I recommend Conal Elliott’s talk on folds and unfolds.


akuleshov7:
# PLEASE NOTE THAT I AM NOT FROM RUSSIA!

EgorBu:
Please measure your head, it's important to follow all european strict requirements and regulations
image

bissquit:
LGTM

elmosmokinweed:
To be sure, you need to identify your pronouns as `not/russian`.

Issue #3936, opened 2024-11-13
switchOnMap with single returned item cancels input


yawkat:
## Expected Behavior

When a switchOnFirst function returns a single item that contains the input flux but is not derived through operators, the input flux should remain available without restriction.

## Actual Behavior

The input flux is cancelled when the outer flux is cancelled (e.g. by `Mono.from`), making it unusable.

## Steps to Reproduce

```java
    public static void main(String[] args) {
        Flux<String> input = Flux.just("foo", "bar");
        Mono<Flux<String>> nested = Mono.from(input.switchOnFirst((first, all) -> Mono.just(all)));
        Flux<String> item = nested.block();
        System.out.println(item); // prints SwitchOnFirstMain as expected
        System.out.println(item.collectList().block()); // never terminates as the item has been cancelled
    }
```

## Possible Solution

In the above example, Mono.from sees the onNext immediately, and cancels the FluxSwitchOnFirst. Cancelling FluxSwitchOnFirst however also cancels the SwitchOnFirstMain, which it should not. Cancellation should only be sent to the publisher returned by the switchOnFirst lambda (in this case the Mono.just)

## Your Environment

* Reactor version(s) used: 3.6.9
* Other relevant libraries versions (eg. `netty`, ...): N/A

chemicL:
Hey, @yawkat !

Thanks for the detailed report. The thing is that you are attempting to use the internal product of `switchOnFirst` outside of its scope.

After you've obtained the `Flux<String> item` from `block()` it means that the `nested` chain has completed. Subscribing to its' inner representation is disallowed outside of the scope of the lambdas that you can define. As stated in the javadoc:

> The whole source (including the first signal) is passed as second argument to the [BiFunction](https://docs.oracle.com/javase/8/docs/api/java/util/function/BiFunction.html?is-external=true) and it is very strongly advised to always build upon with operators (see below).

What you do instead is not derive anything but rather capture this intermediate result and terminate the chain with it. That's not supported. However, perhaps the javadoc can be more specific here.

There's also the following note:

> It is advised to return a [Publisher](https://www.reactive-streams.org/reactive-streams-1.0.3-javadoc/org/reactivestreams/Publisher.html?is-external=true) derived from the original [Flux](https://projectreactor.io/docs/core/release/api/reactor/core/publisher/Flux.html) in all cases

But it follows with another risk that you are not actually running into since you complete the execution. That's when the source is actually cancelled!

I propose two follow up tasks:
[...]

PR #3939, opened 2024-11-21
Fix `SwitchOnFirst` hanging subscription

chemicL:
In a situation where `SwitchOnFirstMain` is subscribed to after the operator itself has terminated, the expectation is that either an error is propagated due to cancellation of the source or the first actual subscription succeeds. Due to a regression introduced in #2794 the subscription would hang with no termination. This change addresses the issue and the subscrption properly terminates.

Related to #3936.


OlegDokuka:
thanks for fixing that!

Issue #3371, opened 2023-03-01
BlockingSingleSubscriber causes memory leak


manoj-mathivanan:
BlockingSingleSubscriber.java accumulates exception in suppressed exception list in the same error object which gets accumulated over time leading to memory leak.

## Expected Behavior
New error object should be created and accumulation on old error object should be avoided.

## Actual Behavior
Error object is cached internally in the kafka sender which is used as a parent error. Future errors while performing send operations  which results in error are accumulated as suppressed exceptions inside the parent error.

## Steps to Reproduce
Create a normal kafka sender. Introduce a config exception like below:
```
delivery.timeout.ms=1200
max.block.ms=1000
request.timeout.ms=2000
linger.ms=10
```
This will throw an error while creating a kafka sender - `org.apache.kafka.common.config.ConfigException: delivery.timeout.ms should be equal to or larger than linger.ms + request.timeout.ms`
Next, send some messages using the kafka sender.
```java
kafkaSender.send(Mono.just(SenderRecord.create(
                        new ProducerRecord<>("topic", "test"),1)))
                .doOnError((Throwable throwable) -> {
              
                })
               .blockFirst();
```

## Your Environment
reactor-core:3.4.16
200 messages/sec creates 200 exceptions in the suppressed exception list. This keeps growing resulting in memory leak.

* Reactor version(s) used: 3.4.16
* Other relevant libraries versions (eg. `netty`, ...): reactor-kafka:1.3.11
* JVM version (`java -version`): 1.8
* OS and version (eg `uname -a`): ubuntu20

OlegDokuka:
@manoj-mathivanan do you have a pointer to an exception which is accumulating errors? 

if it is out of reactor-core than we can do nothing about it

manoj-mathivanan:
@OlegDokuka This is the place where the accumulation happens: `https://github.com/reactor/reactor-core/blob/main/reactor-core/src/main/java/reactor/core/publisher/BlockingSingleSubscriber.java#L99` which is in reactor core.

OlegDokuka:
Right, but that should be okay unless propagated exception is a static one. That say it could be a problem of a different library

manoj-mathivanan:
@OlegDokuka, you are right. The propagated exception is always the same and the new errors are getting added to the same exception inside suppressed exception list. 
Let me try to explain with a scenario.

I create a kafka sender in my bean with wrong configuration 
```
[...]

PR #3941, opened 2024-11-22
Fix memoryleak in FluxOnAssembly

travishaagen:
This PR contains one possible fix for a memory leak discovered in `FluxOnAssembly`.

My colleague, @foo4u discovered a memory leak in one of our applications, and I believe that it is caused by some code in `FluxOnAssembly`. The `nodesPerId` `HashMap` expects the key to be derived from calling `hashCode()` on an object, but inside `FluxOnAssembly.OnAssemblyException#add(Publisher<?>,  Publisher<?>, String, String)` the debugger is showing that an `AssemblyOp` implementation is the object type. None of these implementations implement `hashCode()`, so we're getting the integer returned by `Object.hashCode()` instead.

What I'm observing in a Kotlin Spring Webflux project is that every time we make a HTTP request to a `Controller`, this code path is being hit three times, which results in three objects being added to the `HashMap`. Indirectly, this code path is being triggered by some code in https://github.com/DataDog/dd-trace-java

`FluxOnAssembly` hasn't been changed in three years, and to be honest, the `nodesPerId` and how it is used in this class is odd. It might only be used to increment a counter, but it might also be used as a single point of concurrency control for a couple blocks of code.

All the `AssemblyOp` implementations override `toString()` with something that looks unique, and was probably what the original author intended to base the `nodesPerId` key on.

Here is an example of three keys, from this PR, attributed to a single HTTP request:

```
checkpoint("Handler com.xyz.MyController#myFunc(String, String, String, String, String, ServerWebExchange, Continuation) [DispatcherHandler]")

checkpoint("com.xyz.ShutdownWebFilter [DefaultWebFilterChain]")

checkpoint("HTTP POST "/endpoint" [ExceptionHandlingWebHandler]")
```

Willing to discuss!

Also! There are some other uses of `System.identityHashCode` in the codebase, but I don't have insights into them at the moment.

### DIFF
--- a/reactor-core/src/main/java/reactor/core/publisher/FluxOnAssembly.java
+++ b/reactor-core/src/main/java/reactor/core/publisher/FluxOnAssembly.java
@@ -384,8 +384,7 @@ final class FluxOnAssembly<T> extends InternalFluxOperator<T, T> implements Fuse
 				.findFirst()
 				.orElse(null);
 
-			int thisId = System.identityHashCode(currentAssembly);
-			int parentId = System.identityHashCode(parentAssembly);
+			int thisId = currentAssembly.toString().hashCode();
 
 			ObservedAtInformationNode thisNode;
 			synchronized (nodesPerId) {
@@ -402,6 +401,7 @@ final class FluxOnAssembly<T> extends InternalFluxOperator<T, T> implements Fuse
 					root.addNode(thisNode);
 				}
 				else {
+					int parentId = parentAssembly.toString().hashCode();
 					ObservedAtInformationNode parentNode = nodesPerId.get(parentId);
 					if (parentNode != null) {
 						parentNode.addNode(thisNode);


travishaagen:
The leak only happens when exceptions are thrown. The reasons why one of our applications are so heavily affected are that:

1. It gets a lot of traffic
2. It uses exceptions for flow control in some places

The exceptions are `static` with pre-allocated stack traces as described in this [blog](http://normanmaurer.me/blog/2013/11/09/The-hidden-performance-costs-of-instantiating-Throwables/) (creator of Netty). This technique is used by Netty as well, and is fully supported by the JVM without any hacks.

reactor-core/src/main/java/reactor/core/publisher/FluxOnAssembly.java
```diff
@@ -441,6 +441,15 @@ public String getMessage() {
...
 
+				if (Hooks.GLOBAL_TRACE) {
+					// attach dangling node, without a parent, to root
+					nodesPerId.forEach((id, node) -> {
+						if (node.parent == null && !node.children.isEmpty()) {
+							root.children.add(node);
+						}
+					});
+				}
+
```
travishaagen, line 452: The above change was added, because the following "parallel" test cases were failing, but only with debug mode enabled. Debugging led me to this fix.

image

chemicL:
Hey, thanks for the description of your issue. I did look into the implementation to map out the things you raise and I don't see an indication why the mentioned implementation details would lead to a leak in any way. I am not saying there is no leaks, just the guess that the `nodesPerId`,`root`, and `System.identityHashCode` combination you describe does not seem to be the problem in my view.

The intention of using the `System.identityHashCode` is to identify objects regardless of their actual type by their reference. That makes sense for operators which can be attached multiple times, yet their distinct position in the chain can be pointed to by being a separate object. In your example, three `checkpoint()` operators are defined in the chain, however each one of them has a different identity.

Unless you are able to provide a reproducer for the leak caused by the above I don't think we will be able to apply the proposed change, as it goes against the original design of this functionality as described above.

However, I think the source of the problem is described in your comment about reusing static instances of exceptions and is triggered by:

```java
		final Throwable fail(Throwable t) {
			boolean lightCheckpoint = snapshotStack.isLight();

			OnAssemblyException onAssemblyException = null;
			for (Throwable e : t.getSuppressed()) {
			// ...
			}

			// ...
			t = Exceptions.addSuppressed(t, onAssemblyException);
			/...
		}
```

This causes your static exception to accumulate all the checkpoint data.

This is a similar issue to #3371.

I do not have any solution in mind at this time rather than the suggestion to avoid the shared pre-instantiated and static exceptions or avoid using the checkpoint operator in case you're not willing to give that up.
A middle ground is also possible. Norman's article doesn't mention something that can be found in the comments - you can disable stack trace creation for an exception but don't have to share it. That can be achieved using a constructor available since JDK 1.7 or via overriding the `fillInStackTrace()` method to immediately return `this`.

Regarding the `ParallelFlux*` tests, if you are able to enhance the suite to parameterize tests to run with either enabled or disabled debug mode, feel free to submit a separate PR for that change.

travispeloton:
@chemicL , thank you for your response. I found out that I can use AspectJ on the `FluxOnAssembly.OnAssemblySubscriber.fail(t)` method, and if my `static` exception type is encountered, it simply returns `t`, and my program is behaving as expected. In that case `OnAssemblyException` is never created.

What would you say to a System Property that would enable this behavior via configuration?

`OnAssemblyException` does some potentially expensive things every time an exception is handled:

1. Creates a `HashMap` and `HashSet` (inside `ObservedAtInformationNode`) under all circumstances
2. Uses `synchronized` against `HashMap` multiple times (in my case 3 to 4) in a single operation
3. As I have explained above, they can grow uncontrollably when exceptions are `static`

Some possible options on uncontrolled growth,

- `WeakHashMap` and similar solution for the `Set`
- Limit the number of items that can be added to the `Map` and `Set`. Under normal circumstances these should not grow too high should they? This limit could be configurable via a System Property too.

travispeloton:
Setting `enableSuppression` to `false` will make it so that my `static` exception does not trigger the memory leak condition.

```
RuntimeException(String message, Throwable cause,
                               boolean enableSuppression,
                               boolean writableStackTrace)
```

travispeloton:
@chemicL what about limiting the number of objects that can be added to the Map/Set?

Is there any real use case where you would expect many items in them? 25? 50? 100?

The problem with the current code is that its extremely difficult for someone impacted by this to figure out what is wrong. My project has been impacted for 3 years and we finally figured it out, only because someone took Java Flight Recorder snapshots in production over the course of a week and luckily stumbled upon the right clues.

travispeloton:
Yep, a mention about setting `enableSuppression` to `false` for `static` exceptions would be good to document.

chemicL:
@travispeloton I opened a PR for the documentation note.

chemicL:
Please also take note of https://github.com/reactor/reactor-netty/pull/3529

PR #3944, opened 2024-12-02
[docs] Add warning about static exceptions

chemicL:
In case of enabled debugging or checkpoints, usage of static exceptions in user code can cause memory leaks. A warning note has been added to the debugging section.

Related to #3941


violetagg (approved):

chemicL:
Thanks @violetagg ! Combined with https://github.com/reactor/reactor-core/pull/3901 this should improve the user experience at not much additional cost.

chemicL:
Thanks for the review @violetagg :)

PR #3949, opened 2024-12-04
Polish #3901

chemicL:
@Stephan202 this is a follow-up to #3901 with the most recent feedback and the String-view utility tests and fixes.


Stephan202:
Oh nice! Sorry for not yet replying in the other PR; I'm currently packing boxes as I'll move apartments on Friday, and the deadline is looming :sweat_smile:. I might find some time for a closer look EOD, but please don't feel the need to wait for me. :+1:

Stephan202:
Only just now noticed that #3901 was merged. Thanks! :rocket:

PR #3951, opened 2024-12-12
Update schedulers.adoc, Supplementary semicolon

ShuiMu-peng:
Supplementary semicolon

The last part of the code is missing a semicolon, which will compile an error


chemicL (approved):
Thank you :)

PR #3960, opened 2025-01-02
Support logging exception stack traces when the last argument is Throwable

raccoonback:
## Motivation
Hello,

Reactor Core supports SLF4J, Console Logger, and JDK Logger.
In SLF4J implementations (e.g., Logback), when the last argument is of type Throwable, the exception stack trace is automatically logged, as shown below.
(ref. [SLF4J FAQ](https://www.slf4j.org/faq.html#paramException))
(ref: [Logback implementation](https://github.com/qos-ch/logback/blob/1da2f171dc108a2576eb33918198db7f755b1224/logback-classic/src/main/java/ch/qos/logback/classic/spi/LoggingEvent.java#L140-L142))

```java
log.info("An error occurred: {}", "details", new Exception("Sample exception"));
```

However, the Console Logger and JDK Logger supported by Reactor Core do not provide this functionality, and users must manually check for `Throwable` and handle it accordingly in their code.

```java
	@Test
	void useVerboseConsoleLoggers() throws Exception {
		try {
			Loggers.useVerboseConsoleLoggers();
			Logger l = Loggers.getLogger("test");

			l.debug("Processing data for user: {}", "user123", new Exception("Sample exception"));
		}
		finally {
			Loggers.resetLoggerFactory();
		        ReactorLauncherSessionListener.resetLoggersFactory();
		}
	}
```

```
[DEBUG] (Test worker) Processing data for user: user123
```

This results in unnecessary branching in user code and decreases consistency across logging mechanisms.

Therefore, in this change, we have modified the `ConsoleLogger` and `JdkLogger` to detect if the last argument is of type `Throwable` and automatically log the exception stack trace.

## Description

This PR enhances the logging functionality in Reactor Core to check if the last argument in logger methods is of type `Throwable`. If a `Throwable` is detected, the exception stack trace is included in the log output.

## Key Changes
1. Added logic to inspect the last argument in logger method calls.
2. Automatically handles `Throwable` instances for improved debugging.
3. Ensures compatibility with existing SLF4J-style message formatting for consistency across logging frameworks.

## Example Usage
```java
log.info("Processing data for user: {}", "user123", new Exception("Sample exception"));
```

### Before
```java
INFO: Processing data for user: user123
```
### After
```java
INFO: Processing data for user: user123
java.lang.Exception: Sample exception
    at ...

```

## Benefits
- Aligns Reactor's logging capabilities with familiar SLF4J-style formatting.
- Simplifies exception handling for developers using JDK or Console Loggers.
- Improves traceability and debugging in complex systems.

## Testing
```java

	@Test
	void useVerboseConsoleLoggers() throws Exception {
		try {
			Loggers.useVerboseConsoleLoggers();
			Logger l = Loggers.getLogger("test");

			l.debug("Processing data for user: {}", "user123", new Exception("Sample exception"));
		}
		finally {
			Loggers.resetLoggerFactory();
		        ReactorLauncherSessionListener.resetLoggersFactory();
		}
	}

	@Test
	void useJdkLoggers() throws Exception {
		try {
			Loggers.useJdkLoggers();
			Logger l = Loggers.getLogger("test");

			l.debug("Processing data for user: {}", "user123", new Exception("Sample exception"));
		}
		finally {
			Loggers.resetLoggerFactory();
		        ReactorLauncherSessionListener.resetLoggersFactory();
		}
	}
```

### DIFF
--- a/reactor-core/src/main/java/reactor/util/Loggers.java
+++ b/reactor-core/src/main/java/reactor/util/Loggers.java
@@ -442,12 +442,12 @@ public abstract class Loggers {
 		}
 
 		@Nullable
-		String format(@Nullable String from, @Nullable Object... arguments) {
-			return format(from, false, arguments);
+		private String format(@Nullable String from, @Nullable Object[] arguments) {
+			return format(from, arguments, false);
 		}
 
 		@Nullable
-		private String format(@Nullable String from, boolean skipLast, @Nullable Object... arguments) {
+		private String format(@Nullable String from, @Nullable Object[] arguments, boolean skipLast) {
 			if (from != null) {
 				String computed = from;
 				if (arguments != null && arguments.length != 0) {
@@ -465,7 +465,7 @@ public abstract class Loggers {
 		private void logWithPotentialThrowable(Level level, String format, Object... arguments) {
 			Throwable t = getPotentialThrowable(arguments);
 			if (t != null) {
-				logger.log(level, format(format, true, arguments), t);
+				logger.log(level, format(format, arguments, true), t);
 				return;
 			}
 
@@ -521,12 +521,12 @@ public abstract class Loggers {
 		}
 
 		@Nullable
-		String format(@Nullable String from, @Nullable Object... arguments) {
-			return format(from, false, arguments);
+		private String format(@Nullable String from, @Nullable Object[] arguments) {
+			return format(from, arguments, false);
 		}
 
 		@Nullable
-		private String format(@Nullable String from, boolean skipLast, @Nullable Object... arguments) {
+		private String format(@Nullable String from, @Nullable Object[] arguments, boolean skipLast) {
 			if (from != null) {
 				String computed = from;
 				if (arguments != null && arguments.length != 0) {
@@ -548,7 +548,7 @@ public abstract class Loggers {
 						"[%s] (%s) %s\n",
 						level.toUpperCase(),
 						Thread.currentThread().getName(),
-						format(format, true, arguments)
+						format(format, arguments, true)
 				);
 				t.printStackTrace(logger);
 				return;
--- a/reactor-core/src/main/java/reactor/util/Loggers.java
+++ b/reactor-core/src/main/java/reactor/util/Loggers.java
@@ -625,17 +625,17 @@ public abstract class Loggers {
 
 		@Override
 		public synchronized void info(String msg) {
-			this.log.format("[INFO] (%s) %s\n", Thread.currentThread().getName(), msg);
+			this.log.format("[ INFO] (%s) %s\n", Thread.currentThread().getName(), msg);
 		}
 
 		@Override
 		public void info(String format, Object... arguments) {
-			logWithOptionalThrowable("INFO", format, arguments);
+			logWithOptionalThrowable(" INFO", format, arguments);
 		}
 
 		@Override
 		public synchronized void info(String msg, Throwable t) {
-			this.log.format("[INFO] (%s) %s - %s\n", Thread.currentThread().getName(), msg, t);
+			this.log.format("[ INFO] (%s) %s - %s\n", Thread.currentThread().getName(), msg, t);
 			t.printStackTrace(this.log);
 		}
 
@@ -646,17 +646,17 @@ public abstract class Loggers {
 
 		@Override
 		public synchronized void warn(String msg) {
-			this.err.format("[WARN] (%s) %s\n", Thread.currentThread().getName(), msg);
+			this.err.format("[ WARN] (%s) %s\n", Thread.currentThread().getName(), msg);
 		}
 
 		@Override
 		public void warn(String format, Object... arguments) {
-			logErrorWithOptionalThrowable("WARN", format, arguments);
+			logErrorWithOptionalThrowable(" WARN", format, arguments);
 		}
 
 		@Override
 		public synchronized void warn(String msg, Throwable t) {
-			this.err.format("[WARN] (%s) %s - %s\n", Thread.currentThread().getName(), msg, t);
+			this.err.format("[ WARN] (%s) %s - %s\n", Thread.currentThread().getName(), msg, t);
 			t.printStackTrace(this.err);
 		}
 
--- a/reactor-core/src/main/java/reactor/util/Loggers.java
+++ b/reactor-core/src/main/java/reactor/util/Loggers.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2016-2022 VMware Inc. or its affiliates, All Rights Reserved.
+ * Copyright (c) 2016-2025 VMware Inc. or its affiliates, All Rights Reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -18,6 +18,7 @@ package reactor.util;
 
 import java.io.PrintStream;
 import java.lang.ref.WeakReference;
+import java.util.Arrays;
 import java.util.Map;
 import java.util.Objects;
 import java.util.WeakHashMap;
@@ -353,7 +354,7 @@ public abstract class Loggers {
 
 		@Override
 		public void trace(String format, Object... arguments) {
-			logger.log(Level.FINEST, format(format, arguments));
+			logWithOptionalThrowable(Level.FINEST, format, arguments);
 		}
 
 		@Override
@@ -373,7 +374,7 @@ public abstract class Loggers {
 
 		@Override
 		public void debug(String format, Object... arguments) {
-			logger.log(Level.FINE, format(format, arguments));
+			logWithOptionalThrowable(Level.FINE, format, arguments);
 		}
 
 		@Override
@@ -393,7 +394,7 @@ public abstract class Loggers {
 
 		@Override
 		public void info(String format, Object... arguments) {
-			logger.log(Level.INFO, format(format, arguments));
+			logWithOptionalThrowable(Level.INFO, format, arguments);
 		}
 
 		@Override
@@ -413,7 +414,7 @@ public abstract class Loggers {
 
 		@Override
 		public void warn(String format, Object... arguments) {
-			logger.log(Level.WARNING, format(format, arguments));
+			logWithOptionalThrowable(Level.WARNING, format, arguments);
 		}
 
 		@Override
@@ -433,7 +434,7 @@ public abstract class Loggers {
 
 		@Override
 		public void error(String format, Object... arguments) {
-			logger.log(Level.SEVERE, format(format, arguments));
+			logWithOptionalThrowable(Level.SEVERE, format, arguments);
 		}
 
 		@Override
@@ -453,6 +454,24 @@ public abstract class Loggers {
 				return computed;
 			}
 			return null;
+		}
+
+		private void logWithOptionalThrowable(Level level, String format, Object... arguments) {
+			if(isLastElementThrowable(arguments)) {
+				int lastIndex = arguments.length - 1;
+				Object[] args = Arrays.copyOfRange(arguments, 0, lastIndex);
+				Throwable t = (Throwable) arguments[lastIndex];
+
+				logger.log(level, format(format, args), t);
+				return;
+			}
+
+			logger.log(level, format(format, arguments));
+		}
+
+		private boolean isLastElementThrowable(Object... arguments) {
+			int length = arguments.length;
+			return length > 0 && arguments[length - 1] instanceof Throwable;
 		}
 	}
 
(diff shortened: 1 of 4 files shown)


reactor-core/src/main/java/reactor/util/Loggers.java
```diff
@@ -454,6 +455,24 @@ final String format(@Nullable String from, @Nullable Object... arguments){
...
+				return;
+			}
+
+			logger.log(level, format(format, arguments));
+		}
+
+		private boolean isLastElementThrowable(Object... arguments) {
+			int length = arguments.length;
+			return length > 0 && arguments[length - 1] instanceof Throwable;
+		}
```
raccoonback, line 475: Checks whether the last argements are of type `Throwable`.

reactor-core/src/main/java/reactor/util/Loggers.java
```diff
@@ -454,6 +455,24 @@ final String format(@Nullable String from, @Nullable Object... arguments){
...
+				int lastIndex = arguments.length - 1;
+				Object[] args = Arrays.copyOfRange(arguments, 0, lastIndex);
+				Throwable t = (Throwable) arguments[lastIndex];
+
+				logger.log(level, format(format, args), t);
+				return;
+			}
+
+			logger.log(level, format(format, arguments));
+		}
```
raccoonback, line 470: In `JdkLogger`, if the last arguments are `Throwable` type, stacktrace is also logged.

reactor-core/src/main/java/reactor/util/Loggers.java
```diff
@@ -507,6 +526,39 @@ final String format(@Nullable String from, @Nullable Object... arguments){
...
+				return;
+			}
+
+			this.err.format("[%s] (%s) %s\n", level.toUpperCase(), Thread.currentThread().getName(), format(format, arguments));
+		}
+
+		private boolean isLastElementThrowable(Object... arguments) {
+			int length = arguments.length;
+			return length > 0 && arguments[length - 1] instanceof Throwable;
+		}
```
raccoonback, line 560: Checks whether the last argements are of type `Throwable`.

reactor-core/src/main/java/reactor/util/Loggers.java
```diff
@@ -507,6 +526,39 @@ final String format(@Nullable String from, @Nullable Object... arguments){
...
+				Object[] args = Arrays.copyOfRange(arguments, 0, lastIndex);
+				Throwable t = (Throwable) arguments[lastIndex];
+
+				this.log.format("[%s] (%s) %s\n", level.toUpperCase(), Thread.currentThread().getName(), format(format, args));
+				t.printStackTrace(this.log);
+				return;
+			}
+
+			this.log.format("[%s] (%s) %s\n", level.toUpperCase(), Thread.currentThread().getName(), format(format, arguments));
+		}
```
raccoonback, line 541: In `ConsoleLogger`, if the last argument is of type `Throwable`, it also prints a stack trace.
(only use for TRACE, DEBUG, INFO level)

reactor-core/src/main/java/reactor/util/Loggers.java
```diff
@@ -507,6 +526,39 @@ final String format(@Nullable String from, @Nullable Object... arguments){
...
+				Object[] args = Arrays.copyOfRange(arguments, 0, lastIndex);
+				Throwable t = (Throwable) arguments[lastIndex];
+
+				this.err.format("[%s] (%s) %s\n", level.toUpperCase(), Thread.currentThread().getName(), format(format, args));
+				t.printStackTrace(this.err);
+				return;
+			}
+
+			this.err.format("[%s] (%s) %s\n", level.toUpperCase(), Thread.currentThread().getName(), format(format, arguments));
+		}
```
raccoonback, line 555: In `ConsoleLogger`, if the last arguments are `Throwable` type, it also prints a stack trace with the `err` PrintStream.
(only use for WARN, ERROR level)

raccoonback:
@chemicL 
Hello!
Could you please take a look at this PR?
The review seems to be delayed, so I would greatly appreciate your confirmation.
Thank you!

reactor-core/src/main/java/reactor/util/Loggers.java
chemicL: I'd love to see a test for this.

reactor-core/src/main/java/reactor/util/Loggers.java
chemicL: Can we have tests to validate this?

reactor-core/src/main/java/reactor/util/Loggers.java
```diff
@@ -507,6 +526,39 @@ final String format(@Nullable String from, @Nullable Object... arguments){
 			return null;
 		}
 
+		private synchronized void logWithOptionalThrowable(String level, String format, Object... arguments) {
```
chemicL, line 529: Same comments as with the JdkLogger.

reactor-core/src/main/java/reactor/util/Loggers.java
```diff
@@ -507,6 +526,39 @@ final String format(@Nullable String from, @Nullable Object... arguments){
...
+
+				this.log.format("[%s] (%s) %s\n", level.toUpperCase(), Thread.currentThread().getName(), format(format, args));
+				t.printStackTrace(this.log);
+				return;
+			}
+
+			this.log.format("[%s] (%s) %s\n", level.toUpperCase(), Thread.currentThread().getName(), format(format, arguments));
+		}
+
+		private synchronized void logErrorWithOptionalThrowable(String level, String format, Object... arguments) {
```
chemicL, line 543: This method is almost exactly the same as the one above, can you please unify them with a `PrintStream` as an argument?

reactor-core/src/main/java/reactor/util/Loggers.java
```diff
@@ -573,17 +625,17 @@ public boolean isInfoEnabled() {
...
 		public synchronized void info(String msg) {
-			this.log.format("[ INFO] (%s) %s\n", Thread.currentThread().getName(), msg);
+			this.log.format("[INFO] (%s) %s\n", Thread.currentThread().getName(), msg);
 		}
 
 		@Override
-		public synchronized void info(String format, Object... arguments) {
-			this.log.format("[ INFO] (%s) %s\n", Thread.currentThread().getName(), format(format, arguments));
+		public void info(String format, Object... arguments) {
+			logWithOptionalThrowable("INFO", format, arguments);
```
chemicL, line 633: I am guessing the space was to align INFO under DEBUG and such in a single column so the space was a filler. With the spaces all of (TRACE, DEBUG, INFO, WARN, ERROR) take up the same amount of horizontal space. Please roll back.

chemicL (changes requested):
Thank you for the proposal. I've been occupied with other priority subjects and now found the time to review the PR. I think it is a useful utility. I left some comments, please address them if you'd like to integrate this change. Most importantly, I look forward to tests validating the change has some meaningful effect.

raccoonback:
@chemicL 
First of all, thanks for your review!
I have incorporated all of your suggestions.
Please let me know if there's anything else that needs adjustment!

reactor-core/src/main/java/reactor/util/Loggers.java
raccoonback: Thanks for detailed explanation!
I have reverted the code.

reactor-core/src/main/java/reactor/util/Loggers.java
raccoonback: I have merged the methods into one, and modified it to accept a `PrintStream` as a parameter.

reactor-core/src/main/java/reactor/util/Loggers.java
raccoonback: I have added test code!

reactor-core/src/main/java/reactor/util/Loggers.java
```diff
@@ -454,6 +455,24 @@ final String format(@Nullable String from, @Nullable Object... arguments){
...
 			return null;
 		}
+
+		private void logWithOptionalThrowable(Level level, String format, Object... arguments) {
+			if(isLastElementThrowable(arguments)) {
+				int lastIndex = arguments.length - 1;
+				Object[] args = Arrays.copyOfRange(arguments, 0, lastIndex);
+				Throwable t = (Throwable) arguments[lastIndex];
+
+				logger.log(level, format(format, args), t);
```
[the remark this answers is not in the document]
raccoonback: I added the `skipLast` parameter to the `format(...)` method. Since `Varargs` must be placed at the end, I added `skipLast` as a **middle** parameter. 
Additionally, I overloaded the existing method.

reactor-core/src/main/java/reactor/util/Loggers.java
```diff
@@ -494,19 +521,57 @@ public String getName() {
...
+					for (int index = 0; index < lastIndex; ++index) {
+						computed = computed.replaceFirst("\\{\\}", Matcher.quoteReplacement(String.valueOf(arguments[index])));
 					}
 				}
 				return computed;
 			}
 			return null;
 		}
 
+		private synchronized void logWithPotentialThrowable(PrintStream logger, String level, String format, Object... arguments) {
```
raccoonback, line 544: add a `PrintStream` parameter.

reactor-core/src/main/java/reactor/util/Loggers.java
```diff
@@ -442,18 +442,45 @@ public void error(String msg, Throwable t) {
 		}
 
 		@Nullable
-		final String format(@Nullable String from, @Nullable Object... arguments){
-			if(from != null) {
+		String format(@Nullable String from, @Nullable Object... arguments) {
```
chemicL, line 445: This one can also be private.

reactor-core/src/main/java/reactor/util/Loggers.java
```diff
@@ -442,18 +442,45 @@ public void error(String msg, Throwable t) {
...
+
+		@Nullable
+		private String format(@Nullable String from, boolean skipLast, @Nullable Object... arguments) {
+			if (from != null) {
 				String computed = from;
 				if (arguments != null && arguments.length != 0) {
-					for (Object argument : arguments) {
-						computed = computed.replaceFirst("\\{\\}", Matcher.quoteReplacement(String.valueOf(argument)));
+					int lastIndex = arguments.length;
+					if (skipLast) --lastIndex;
```
chemicL, line 455: Please make sure you use braces around the instructions following an if condition.

reactor-core/src/main/java/reactor/util/Loggers.java
```diff
@@ -442,18 +442,45 @@ public void error(String msg, Throwable t) {
...
 
 		@Nullable
-		final String format(@Nullable String from, @Nullable Object... arguments){
-			if(from != null) {
+		String format(@Nullable String from, @Nullable Object... arguments) {
+			return format(from, false, arguments);
+		}
+
+		@Nullable
+		private String format(@Nullable String from, boolean skipLast, @Nullable Object... arguments) {
```
chemicL, line 450: > I added the skipLast parameter to the format(...) method. Since Varargs must be placed at the end, I added skipLast as a middle parameter.
Additionally, I overloaded the existing method.

Would it work if instead of `Object...` you used `Object[]` ? Since this is an internal API it should be totally fine with the usages here. Then the order of arguments will be more appropriate IMO.

chemicL (changes requested):
Thanks for the follow up. I added a few more comments, please apply them also to similar areas where appropriate. Please check the build results, they are failing with NPE. Before the next round please make sure the build is passing on your machine.

reactor-core/src/main/java/reactor/util/Loggers.java
raccoonback: I have changed it to use `Object[]` instead.

reactor-core/src/main/java/reactor/util/Loggers.java
raccoonback: Reflect it!

reactor-core/src/main/java/reactor/util/Loggers.java
raccoonback: Changed to `private` accessor.

raccoonback:
@chemicL 
Hello.

Thank you for your detailed review!
I have addressed all the comments and also fixed the issue causing the NPE.

chemicL (changes requested):
Thank you for the follow up. The production code looks good. Regarding the tests, please refrain from using mocks when more end-to-end tests are possible. The existing tests allow validating the output. Check the use of `ConsoleLoggerTest#outContent` and `JdkLoggerTest#nullFormatIsAcceptedByUnderlyingLogger`. Also, please revert unrelated formatting changes so that the PR consists only of changes relevant to the problem that's being solved. Thanks.

raccoonback:
@chemicL 

I have rolled back the unrelated formatting changes.

Additionally, since ConsoleLoggerTest already covers all cases for ConsoleLogger, I found the newly added test to be redundant and have therefore removed it.

chemicL:
> I have rolled back the unrelated formatting changes.
> 
> Additionally, since ConsoleLoggerTest already covers all cases for ConsoleLogger, I found the newly added test to be redundant and have therefore removed it.

I'm not sure I understand. Since the functionality of `ConsoleLogger` has been enhanced, there should be a need for more tests. In `JdkLoggerTest` there is no validation of the stacktrace. Please thoroughly review the changes and expectations. Also, make sure the test methods are descriptive as the existing methods are and capture the circumstance and expectation.

raccoonback:
@chemicL 

I misunderstood;;
As you mentioned, I have added a test for the case where the last argument of varargs is a `Throwable`.
Additionally, I modified `JdkLoggerTest` to validate the stacktrace and extracted the duplicate code for creating JdkLogger into the `getJdkLogger()` method.

chemicL (approved):

chemicL:
@raccoonback thank you for the contribution! 🚀

PR #3966, opened 2025-01-14
Remove redundant return in subscribe catch block

Aleexender:
Remove redundant return statement in CorePublisher subscribe method

The explicit return statement in the catch block is unnecessary since the method would naturally terminate after reporting the error through reportThrowInSubscribe.

### User Impact
Cleaner code with no behavioral changes for end users. This is purely a maintenance improvement that removes redundant code.

### Technical Details
The catch block in the subscribe method contained an unnecessary explicit return statement after `Operators.reportThrowInSubscribe()`. Since the method would naturally terminate after error reporting, this return statement was redundant and has been removed to improve code clarity.

### Related Issues
See #2072

### DIFF
--- a/reactor-core/src/main/java/reactor/core/publisher/Mono.java
+++ b/reactor-core/src/main/java/reactor/core/publisher/Mono.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2016-2024 VMware Inc. or its affiliates, All Rights Reserved.
+ * Copyright (c) 2016-2025 VMware Inc. or its affiliates, All Rights Reserved.
  *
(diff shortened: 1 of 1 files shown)


chemicL (approved):

chemicL (changes requested):
Thank you for the PR! Please note you need to update the license header as per the CI run. Please run `./gradlew spotlessApply` locally and update the PR. Thanks!

Aleexender:
@chemicL  hello! I ran spotlessApply locally! Thank you for your help!

chemicL (approved):

chemicL:
Thanks, @Aleexender :)

PR #3982, opened 2025-02-06
Refactor mono wrap method

Aleexender:
Description:
This refactoring improves the maintainability and readability of the Mono.wrap method through the following changes:

1. Extract Mono source creation logic into a separate method (createMonoSource)
   - Reduces cognitive complexity of the main wrap method
   - Makes the source creation logic more focused and easier to maintain

2. Simplify conditional logic using ternary operators
   - Reduces nesting levels in the code
   - Makes the flow control more straightforward

3. Improve type handling consistency
   - Refined type casting structure following existing patterns
   - Maintained consistent use of SuppressWarnings annotations in line with the codebase's existing conventions

The changes maintain the exact same functionality while making the code structure clearer and more maintainable.

PR #3984, opened 2025-02-10
change to ternary return

Aleexender:
**Refactor point**

•	Replaced explicit if-else statement with a ternary return for conciseness.
•	This improves readability and maintains the same logic.

changed point - Mono.fromDirect()


chemicL:
Can you please make this change part of #3982 ?

PR #3985, opened 2025-02-10
refector when methods(when, whenDelayError)

Aleexender:
Refactored an if-else statement into a ternary return to improve readability and conciseness.

This change simplifies the return logic by replacing an explicit if-else with a ternary operator. However, I am aware that this syntax is available from Java 14 onwards. I checked git blame and noticed that this code has not been modified for quite some time, so I took the opportunity to refactor it.

If this change is acceptable, please feel free to leave a comment.

Effect points: when(Publisher<?>... sources) , whenDelayError(Publisher<?>... sources)


chemicL:
Hey, our codebase needs to stay on Java 8. I'm sorry but I have to reject this proposal.

Aleexender:
oh sorry i forgot closed to this PR.

thanks for help.

PR #4009, opened 2025-03-09
Add more description on tap operator

ttddyy:
Add an explanation of where the observation created by the `tap` operator is available.


chemicL (approved):

chemicL:
Thanks, @ttddyy 🚀

PR #4023, opened 2025-04-18
Refactor to ternary operator

Aleexender:
### Refactor: use ternary operators for conciseness

This PR refactors a few conditional return statements to use ternary operators in order to improve readability and conciseness.

### Changes:
- Replaced simple `if-else` statements in methods like `Mono#flux()`, `Mono#filter()` with equivalent ternary expressions.
- Ensured that the logic remains unchanged and safe (e.g., `instanceof` checks and casting).

### Rationale:
- Improves brevity and readability where the condition and branches are simple.
- Helps align the style with similar concise expressions already present in the codebase.

Let me know if there's any concern about the use of ternary in these contexts.


chemicL:
Hi, thanks for the contribution. I don't see much value added in swapping ifs with ternary operators in a codebase that is stable. Perhaps this can be done in the future with some other functional changes, but for now I prefer to discard the change.

PR #4027, opened 2025-05-09
Repeat specification for use with repeatWhen operator

kwondh5217:
### Description
This PR addresses #3545.

This introduces RepeatSpec into core, aligning with the structure of Retry, without modifying the existing repeatWhen or Mono#repeatWhen method signatures.

The implementation brings over the functionality of repeat from reactor-addons, allowing users to configure repeat behavior through static factory methods like times(...), create(...), and once().

### Result
Closes gh-3545


kwondh5217:
@chemicL, Appreciate the review! Glad to hear it’s a good starting point.
Let me know if you’d like help with the follow-up cleanup or adding missing features.
I’d be happy to contribute further.

PR #4090, opened 2025-09-16
Add Hooks#shouldPropagateContextToThreadLocals

sdeleuze:
This commit adds a public API to get the status of the context propagation.


reactor-core/src/main/java/reactor/core/publisher/Hooks.java
```diff
@@ -561,6 +561,16 @@ public static void disableAutomaticContextPropagation() {
 		}
 	}
 
+	/**
+	 * @return true if automatic context propagation to {@link ThreadLocal}s is enabled, false otherwise
+	 * @see #enableAutomaticContextPropagation
+	 * @see #disableAutomaticContextPropagation
+	 * @since 3.8.0
+	 */
+	public static boolean shouldPropagateContextToThreadLocals() {
```
chemicL, line 570:
```suggestion
	public static boolean isAutomaticContextPropagationEnabled() {
```

chemicL (changes requested):
Thanks! I think we could change the name, since the one in an internal class guides the internal behaviour and the intention is well known. But for a public API we don't know the intention, so a check whether the automatic propagation mode is in fact enabled is all we know we are exposing.

Can we target 3.7.x branch? It might be beneficial in projects depending on that line as well. We forward merge such changes.

chemicL (approved):
Thank you!

PR #4091, opened 2025-09-17
Introduce JSpecify for nullability annotations

chemicL:
The JSR-305-based reactor.util.annotation.* classes have been deprecated in favour of JSpecify annotations.
The build runs NullAway errorprone plugin. Several public APIs have been refined to better reflect the nullability characteristics and also some internal implementations have also been adjusted based on the new compilation warnings.
This PR replaces #4016.

### DIFF
--- a/reactor-core/src/main/java/reactor/core/publisher/DelegateProcessor.java
+++ b/reactor-core/src/main/java/reactor/core/publisher/DelegateProcessor.java
@@ -19,6 +19,7 @@ package reactor.core.publisher;
 import java.util.Objects;
 import java.util.stream.Stream;
 
+import org.jspecify.annotations.NonNull;
 import org.jspecify.annotations.Nullable;
 import org.reactivestreams.Publisher;
 import org.reactivestreams.Subscriber;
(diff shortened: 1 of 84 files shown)


reactor-core/src/main/java/reactor/core/publisher/ConnectableFluxOnAssembly.java
```diff
@@ -55,7 +55,7 @@ final class ConnectableFluxOnAssembly<T> extends InternalConnectableFluxOperator
 	}
 
 	@Override
-	public final CoreSubscriber<? super T> subscribeOrReturn(CoreSubscriber<? super T> actual) {
```
sdeleuze, line 58: Is it done on purpose?

reactor-core/src/main/java/reactor/core/publisher/DelegateProcessor.java
```diff
@@ -99,10 +99,11 @@ public int getBufferSize() {
 		                .scanOrDefault(Attr.CAPACITY, super.getBufferSize());
 	}
 
+	// super.getError() returns null by default and this method would throw NPE in that
+	// case, however the parent's contract is to return nullable Throwable
 	@Override
-	@Nullable
```
sdeleuze, line 103: Have you tried explicit `@NonNull` instead of `@SuppressWarnings`?

sdeleuze (commented):
LGTM, modulo 2 minor comments and the remark I shared previously on double checking arrays nullability manually as the JSpecifyMode is not yet enabled yet.

reactor-core/src/main/java/reactor/core/publisher/DelegateProcessor.java
chemicL: The problem is actually in the argument passed to `scanOrDefault` which is required to be non-null, otherwise the method throws. Somehow the contract of that method is that it should not accept null values, yet there is extra verification using `Object.requireNonNull` in place in the body. This is the only weird use of it that can pass on a null argument. Since the class is deprecated I don't want to spend much more time investigating.

PR #4112, opened 2025-10-08
Fix elapsed time computation in OptimisticEmitFailureHandler

bjmi:
This change addresses the elapsed time computation in `OptimisticEmitFailureHandler` due to the possibility of a numerical overflow. The described behavior isn't changed from user's perspective.

Javadoc of System.nanoTime() explicitly mentions how to measure elapsed time.
> To compare elapsed time against a timeout, use
> `if (System.nanoTime() - startTime >= timeoutNanos) ...`
> instead of
> `if (System.nanoTime() >= startTime + timeoutNanos) ...`
> because of the possibility of **numerical overflow**.

See [System.nanoTime()](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/lang/System.html#nanoTime())

### DIFF
--- a/reactor-core/src/main/java/reactor/core/publisher/Sinks.java
+++ b/reactor-core/src/main/java/reactor/core/publisher/Sinks.java
@@ -224,20 +224,22 @@ public final class Sinks {
 	}
 
 	/**
-	 *
 	 * @author Animesh Chaturvedi
 	 */
 	static class OptimisticEmitFailureHandler implements EmitFailureHandler {
 
+        private final long startTime;
 		private final long deadline;
 
 		OptimisticEmitFailureHandler(Duration duration){
-			this.deadline = System.nanoTime() + duration.toNanos();
+            this.startTime = System.nanoTime();
+			this.deadline = duration.toNanos();
 		}
 
 		@Override
 		public boolean onEmitFailure(SignalType signalType, EmitResult emitResult) {
-			return emitResult.equals(Sinks.EmitResult.FAIL_NON_SERIALIZED) && (System.nanoTime() < this.deadline);
+			return emitResult.equals(Sinks.EmitResult.FAIL_NON_SERIALIZED)
+                    && System.nanoTime() - startTime < deadline;
 		}
 	}
 
--- a/reactor-core/src/main/java/reactor/core/publisher/Sinks.java
+++ b/reactor-core/src/main/java/reactor/core/publisher/Sinks.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2020-2025 VMware Inc. or its affiliates, All Rights Reserved.
+ * Copyright (c) 2020-2026 VMware Inc. or its affiliates, All Rights Reserved.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


reactor-core/src/main/java/reactor/core/publisher/Sinks.java
```diff
@@ -224,20 +224,22 @@ public EmitResult getReason() {
 	}
 
 	/**
-	 *
 	 * @author Animesh Chaturvedi
 	 */
 	static class OptimisticEmitFailureHandler implements EmitFailureHandler {
 
+        private final long startTime;
```
chemicL: While the change sounds correct and protects against overflow, it introduces an additional `long` type field to a dynamically allocated object which in a high-load scenario can lead to unnecessary allocation.

How about skipping `startTime`, sticking with the constructor as-is and relying on the overflow but adjusting the check in `onEmitFailure`?

Moving from `System.nanoTime() < this.deadline` to `this.deadline - System.nanoTime() > 0` has the same meaning, but would tackle the overflow:

Suppose `System.nanoTime()` is `Long.MAX_VALUE - 5`.
`durationNanos` is 10.
`deadline = Long.MAX_VALUE - 5 + 10 = Long.MAX_VALUE + 5 = Long.MIN_VALUE + 4` (overflows to negative).

Let's check if we get the `durationNanos` back with `deadline - System.nanoTime()`:
```
Long.MIN_VALUE + 4 - (Long.MAX_VALUE - 5) = ?
// Long.MIN_VALUE - Long.MAX_VALUE is 1
So (Long.MIN_VALUE + 4) - (Long.MAX_VALUE - 5) = 1 + 4 - (-5) = 10.
```
 
A unit test for this:

```
          // Simulate System.nanoTime() near Long.MAX_VALUE
          long startTime = Long.MAX_VALUE - 5;
          long durationNanos = 10;

          // This is what the current code does:
          long deadline = startTime + durationNanos;

          // Verify it overflowed to a negative number
          assertThat(deadline).isLessThan(0);

          // 1. Simulate time BEFORE deadline (e.g., 3 nanos passed)
          long timeBefore = startTime + 3;

          // Current buggy check: System.nanoTime() < deadline
          // Fails because timeBefore is positive, deadline is negative
          assertThat(timeBefore < deadline).isFalse();

          // Proposed fix: deadline - System.nanoTime() > 0
          assertThat(deadline - timeBefore > 0).isTrue();

          // 2. Simulate time EXACTLY AT deadline (e.g., 10 nanos passed)
          // This will also overflow to a negative number
          long timeAt = startTime + 10;

          // Proposed fix: deadline - System.nanoTime() > 0
          // Should be false because it's exactly at the deadline (time is up)
          assertThat(deadline - timeAt > 0).isFalse();
          assertThat(deadline - timeAt).isEqualTo(0);

          // 3. Simulate time AFTER deadline (e.g., 15 nanos passed)
          long timeAfter = startTime + 15;

          // Proposed fix: deadline - System.nanoTime() > 0
          // Should be false because deadline has passed
          assertThat(deadline - timeAfter > 0).isFalse();
          assertThat(deadline - timeAfter).isLessThan(0);
```

Let me know your feedback.

chemicL (changes requested):

reactor-core/src/main/java/reactor/core/publisher/Sinks.java
bjmi: Agree.
That sounds like a feasible suggestion for cutting out another long at the expense of readability.

reactor-core/src/main/java/reactor/core/publisher/Sinks.java
bjmi, line 234: Should there be a check for non-negative values for the `duration`?

chemicL (approved):

reactor-core/src/main/java/reactor/core/publisher/Sinks.java
chemicL: I don't see a need TBH. A negative duration would simply mean the retry doesn't happen.

chemicL:
Thank you for the contribution @bjmi !

PR #4206, opened 2026-02-18
[doc] Fix an outdated link in the Mono class javadoc

RenanMarques:
The javadoc of the Mono class had an outdated link to the "Which operator do I need?" documentation page that was redirecting readers to the root page of the documentation (About the Documentation). Now the link is fixed and references the correct page.


chemicL (approved):

chemicL:
Thanks for the contribution!

PR #4231, opened 2026-03-31
[#4230] FluxTakeUntilOther: Cancel main when other errors

### LINKED ISSUE #4230: FluxTakeUntilOther does not cancel infinite Main on async companion error
When applying `source.takeUntilOther(other)` to an infinite source (`Flux`), that infinite source is not canceled if/when the companion `other` emits an asynchronous error.

## Expected Behavior
An infinite `Flux` source with `takeUntilOther` applied should be canceled if/when the `other` companion asynchronously emits an error.

## Actual Behavior
Source `Flux` is not canceled and left dangling.

## Steps to Reproduce
The [test method `FluxTakeUntilOtherTest.otherSignalsError()`](https://github.com/reactor/reactor-core/blob/2a22bcc1bb957f14df3d1cd0819d2b12b931efc2/reactor-core/src/test/java/reactor/core/publisher/FluxTakeUntilOtherTest.java#L196-L217) can be updated to the following to expose the bug:

```java
    @Test
    public void otherSignalsError() {
        AssertSubscriber<Integer> ts = AssertSubscriber.create();
        AtomicBoolean mainCancelled = new AtomicBoolean(false);
        AtomicBoolean otherCancelled = new AtomicBoolean(false);

        Flux<Object> other =
            Flux.error(new RuntimeException("forced " + "failure"))
                .delaySubscription(Duration.ofMillis(1)) // Make other async
                .doOnCancel(() -> otherCancelled.set(true));
        Flux.<Integer>never() // Changed to infinite source
            .doOnCancel(() -> mainCancelled.set(true))
            .takeUntilOther(other)
            .subscribe(ts);

        ts.await() // Await termination
            .assertNoValues()
            .assertNotComplete()
            .assertError(RuntimeException.class)
            .assertErrorMessage("forced failure");

        Assertions.assertThat(mainCancelled).isTrue();
        Assertions.assertThat(otherCancelled).isFalse();
    }
```

Note that this error does not occur if source is finite and synchronous, nor does it occur if the other companion is finite and synchronous. In the former case, the source can complete before the other emits asynchronously. In the latter case, the error subscription and immediate emission causes subscription to an empty main subscription, and then the main subscription does get cancelled, since the "already subscribed" guard is hit (already referencing a canceled subscription).

## Possible Solution
Main should be canceled if/when other emits an error asynchronously.

## Your Environment
* Reactor version(s) used: 3.8.4
* JVM version (`java -version`): 17.0.16
* OS and version (eg `uname -a`): Darwin Kernel Version 25.3.0

Sage-Pierce:
Update FluxTakeUntilOther to cancel `main` if/when `other` errors

Copy similar behavior of `other` completion to cancel `main` before emitting termination on `other` error. 

Fixes #4230

### DIFF
--- a/reactor-core/src/main/java/reactor/core/publisher/FluxTakeUntilOther.java
+++ b/reactor-core/src/main/java/reactor/core/publisher/FluxTakeUntilOther.java
@@ -114,7 +114,8 @@ final class FluxTakeUntilOther<T, U> extends InternalFluxOperator<T, T> {
 				return;
 			}
 			once = true;
-			main.onError(t);
+			//cancel main.main, and ensure that if this happens early an empty Subscription is passed down
+			main.cancelMainAndError(t);
 		}
 
 		@Override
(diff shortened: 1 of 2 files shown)


Sage-Pierce:
@chemicL @violetagg may I request a review on this? TIA!

reactor-core/src/main/java/reactor/core/publisher/FluxTakeUntilOther.java
```diff
@@ -183,6 +184,26 @@ public void request(long n) {
 			s.request(n);
 		}
 
+		void cancelMainAndError(Throwable t) {
```
Sage-Pierce, line 187: Note that this is virtually identical to the `cancelMainAndComplete` method below, except signaling error vs. complete

PR #4268, opened 2026-06-08
Fix gradle setup when handling hotfix versions

chemicL:


reactor-tools/build.gradle
chemicL, line 1: changes here address a potential stale cache problem I ran into locally that the test fails if the project version is changed locally

violetagg (approved):

PR #4347, opened 2026-08-04
Fix groupBy cancellation signal loss

### LINKED ISSUE #2352: FluxGroupBy silently drops onNext signals

noodlemind:
Cancelled GroupedFlux instances no longer silently lose values that were accepted or dequeued while cancellation raced with delivery. Buffered values are handed back to GroupByMain and routed into a replacement group, preserving signal counts without scheduler isolation.

A cancellation could previously win after GroupByMain selected a group, or after the inner drain polled a value, and then clear the queue or complete downstream. This change keeps completed groups attached long enough to hand off buffered values, makes group removal identity-aware so an old group cannot remove its replacement, serializes queue mutation with cancellation, and keeps dequeue-and-delivery atomic. Outer completion waits for subscribed groups to finish any pending handoff.

Checks:
- deterministic cancellation-after-poll regression test
- maintainer simplified range/groupBy/take(1) reproducer
- timed multi-threaded group cancellation reproducer
- high-cardinality multi-scheduler stress scenario (5 consecutive runs)
- ./gradlew :reactor-core:test spotlessCheck (10,005 tests; 9,595 passed, 410 skipped)

Fixes #2352.

### DIFF
--- a/reactor-core/src/main/java/reactor/core/publisher/FluxGroupBy.java
+++ b/reactor-core/src/main/java/reactor/core/publisher/FluxGroupBy.java
@@ -121,6 +121,8 @@ final class FluxGroupBy<T, K, V> extends InternalFluxOperator<T, GroupedFlux<K, 
 				AtomicLongFieldUpdater.newUpdater(GroupByMain.class, "requested");
 
 		volatile boolean done;
+		volatile boolean completing;
+		volatile boolean terminated;
 
 		volatile @Nullable Throwable error;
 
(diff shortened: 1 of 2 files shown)


noodlemind:
@chemicL @Sage-Pierce Can you please review this request. Thank you!

Sage-Pierce:
@noodlemind it's neat to see progress on this issue! Your changes look _reasonable_ to me, but as mainly a Reactor _user_ and _occasional_ contributor, I don't know that I'll be able to do much but cheer on 😅 I would much more trust reviewership from Reactor maintainers, or anyone with more expertise in how the GroupBy operator is intended to be implemented.

Just FYI (and as I understand it), projects that fall under the Spring ecosystem umbrella are a bit strained for maintenance resources these days. See [this blog](https://spring.io/blog/2026/06/01/spring_and_security_in_the_times_of_ai) for more details. I also think this time of year is a popular holiday/vacation window for maintainers, so that may further lengthen turnaround time.

It may be helpful to further set expectations by keeping the [Spring release calendar](https://spring.io/projects#release-calendar) in mind when trying to judge when any proposed changes might be published for availability.

skrcode:
I ran JAIPilot Cloud against this exact PR head. The only safe follow-up it retained was a behavior-neutral cleanup: remove the now-unused private queue parameter after the cancellation fix.

The cloud run exercised the repository Java 8, 11, and 21 verification, including the concurrency checks. No performance claim is being made.

Cloud-generated draft and evidence: https://github.com/skrcode/reactor-core/pull/2  
PR directly onto this source branch: https://github.com/noodlemind/reactor-core/pull/1

Feel free to merge or cherry-pick if the cleanup is useful.

PR #4356, opened 2026-08-14
Document that reactor-core-micrometer meters stay registered

### LINKED ISSUE #4271: Document reactor-core-micrometer policy on registered metrics
All metrics created via reactor-core-micrometer are long-lived and are not deregistered. Despite reactive pipelines being potentially short lived, the repeated creation of pipelines with the same name lends itself to holding the metrics in memory for the lifetime of the app. While this is a regular practice, it is worth documented so users are not surprised and can plan accordingly or deregister metrics by hand if they no longer wish to hold these registrations in memory.

AzazelSensei:
Meters from reactor-core-micrometer stay on the registry after a short-lived pipeline finishes. Same name/tags reuse them; delete the meter yourself if you don't want that.

Fixes #4271.

### DIFF
--- a/docs/modules/ROOT/pages/metrics.adoc
+++ b/docs/modules/ROOT/pages/metrics.adoc
@@ -85,6 +85,11 @@ listenToEvents()
 
 The detail of the exposed metrics is available in xref:metrics-details.adoc#micrometer-details-metrics[Micrometer.metrics()].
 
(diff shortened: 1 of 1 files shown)


AzazelSensei:
Closing this. Open PRs with no eyes on them are just wasted work.

PR #4388, opened 2026-09-08
Add support for reproducible builds

mbhave:
-noExtraHeaders in the bnd properties makes sure that non-deterministic headers like Created-By do not get added to the manifest.

### DIFF
--- a/gradle/setup.gradle
+++ b/gradle/setup.gradle
@@ -21,13 +21,18 @@ apply plugin: 'maven-publish'
 //we also conditionally apply artifactory and signing plugins below
 
 jar {
-	manifest.attributes["Created-By"] = "${System.getProperty("java.version")} (${System.getProperty("java.specification.vendor")})"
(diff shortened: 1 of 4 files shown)


gradle/setup.gradle
```diff
@@ -21,7 +21,7 @@ apply plugin: 'maven-publish'
 //we also conditionally apply artifactory and signing plugins below
 
 jar {
-	manifest.attributes["Created-By"] = "${System.getProperty("java.version")} (${System.getProperty("java.specification.vendor")})"
+	manifest.attributes["Jdk-Build-Spec"] = "${System.getProperty("java.specification.version")}"
```
chemicL, line 24: I believe this should be Build-Jdk-Spec (https://maven.apache.org/shared/maven-archiver/)

chemicL:
Thank you. I merged this together with additional changes proposed by @violetagg from reactor-netty: 8d2d42c353da32694bfd67d06779721b73b84bd7

